PR #1, opened 2016-04-06
Fix test-scoped deps

jglick:
Otherwise with https://github.com/jenkinsci/jenkins/pull/2172 plugins will (correctly) be disabled due to version mismatches.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,12 @@
             <artifactId>workflow-basic-steps</artifactId>
             <version>1.15</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jenkins-ci.plugins</groupId>
+                    <artifactId>mailer</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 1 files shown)


jglick:
Switching CI builder to Java 8.

pom.xml
```diff
@@ -172,6 +178,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency> <!-- optional dep of workflow-scm-step, have to override 1.54 version that gets pulled in somehow during tests -->
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>subversion</artifactId>
+            <version>2.5</version>
```
recena: 🐜  Please, `2.5.7`

recena:
🐝

pom.xml
jglick: 2.5 was intentional—the version declared as a minimum by `workflow-scm-step`.

recena: This version should not use at all. Next time :bug: instead of :ant: :smile:

jglick: `workflow-scm-step` does not _advocate_ use of 2.5. Whatever the current release is on the update center, that is what the user will get. It simply _requires_ 2.5 or higher.

PR #3, opened 2016-04-08
[JENKINS-26156] BodyInvoker.displayName was broken

jglick:
[JENKINS-26156](https://issues.jenkins-ci.org/browse/JENKINS-26156)

Before:
* if you passed null, or simply did not call it at all, there were body block nodes contrary to documentation
* if you passed non-null, you would get an `IllegalStateException` (!)

After:
* body block nodes are created unconditionally
* you may call it, with a non-null name, if you want a label on the body block

@kohsuke seems to have been trying to elide body block nodes when they contribute no interesting information, as in most steps accepting a closure (`node`, say, as opposed to `retry` which _might_ have something interesting to show like a retry count). But this was never actually implemented, and seems too dangerous to try anyway. What if a step asked to elide the body block nodes but then actually called its body twice? We would get a confused flow graph.

I _did_ clean up the log display a bit, from what @amuniz did in https://github.com/jenkinsci/pipeline-plugin/pull/215 (cf. [JENKINS-31595](https://issues.jenkins-ci.org/browse/JENKINS-31595)). Before:

```
[Pipeline] Some display name intended for GUI : Start
[Pipeline] theActualStepName {
body text…
[Pipeline] } //theActualStepName
[Pipeline] Some display name intended for GUI : End
```

After:

```
[Pipeline] theActualStepName {
[Pipeline] {
body text…
[Pipeline] }
[Pipeline] } // theActualStepName
```

With some more work in `WorkflowRun.logNodeMessage` this could probably be improved further, to check for `BodyInvocationAction` and `LabelAction` on `BlockStartNode`s, giving us a more streamlined

```
[Pipeline] theActualStepName {
body text…
[Pipeline] } // theActualStepName
```

plus some variants for bodies which are actually run >1 time and/or specify a label, but I will leave that for another day.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -6,7 +6,6 @@ import com.cloudbees.groovy.cps.Next;
 import com.cloudbees.groovy.cps.Outcome;
 import com.cloudbees.groovy.cps.impl.CpsCallableInvocation;
 import com.cloudbees.groovy.cps.impl.FunctionCallEnv;
-import com.cloudbees.groovy.cps.impl.SourceLocation;
 import com.cloudbees.groovy.cps.impl.TryBlockEnv;
 import com.cloudbees.groovy.cps.sandbox.SandboxInvoker;
 import com.google.common.util.concurrent.FutureCallback;
@@ -18,14 +17,12 @@ import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
-import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.steps.BodyExecution;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
-import javax.annotation.CheckForNull;
 import javax.annotation.concurrent.GuardedBy;
 import java.io.IOException;
 import java.util.Collection;
@@ -37,6 +34,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.*;
+import javax.annotation.Nonnull;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
 
 /**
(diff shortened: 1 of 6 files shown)


jglick:
Note that there was no need to touch `ParallelStepExecution`—it does call `withStartAction(Action)`, which is not currently exposed as an API in `BodyInvoker` for use from other plugins, but the action it attaches is not only `LabelAction` but a `ThreadNameAction`. That is why this step successfully attached labels to its body invocations from the start: it bypassed the buggy code by using a plugin-private method.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -86,15 +84,9 @@
     @GuardedBy("this")
     private Outcome outcome;
 
-    /**
-     * @see CpsBodyInvoker#createBodyBlockNode
-     */
-    private final boolean createBodyBlockNode;
-
-    public CpsBodyExecution(CpsStepContext context, List<BodyExecutionCallback> callbacks, boolean createBodyBlockNode) {
+    CpsBodyExecution(CpsStepContext context, List<BodyExecutionCallback> callbacks) {
```
amuniz: Backward incompatible?

amuniz:
That nested `{` and `}` confused me. Why not this:

```
[Pipeline] theActualStepName
[Pipeline] {
body text…
[Pipeline] }
[Pipeline] // theActualStepName
```

Other than that :bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: No. The class as a whole was not `public`, so no one could have been calling this constructor from outside anyway. (Not that we would want to support them if they did.)

amuniz: Oh, right, I missed that.

amuniz:
:bee: Thanks!

recena:
@amuniz Obviously, this is only personal likes, but I prefer:

```
[Pipeline] theActualStepName
[Pipeline] {
body text…
[Pipeline] } // theActualStepName
```

it is only a note.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -318,57 +310,37 @@ public Next receive(Object o) {
...
-                LOGGER.log(WARNING, "Failed to grow the flow graph", e);
-                throw new Error(e);
-            }
-        } else {
-            return null;
+    private @Nonnull StepEndNode addBodyEndFlowNode() {
+        try {
+            FlowHead head = CpsThread.current().head;
+
+            StepEndNode end = new StepEndNode(head.getExecution(),
```
recena: Why not one line? :smile:

recena:
LGTM :bee: though this part of Pipeline is completely new to me (no experience). I've tried the PR and does what is described in the PR.

jglick:
> `[Pipeline] } // theActualStepName`

Not possible without much bigger changes: these are two `FlowNode`s, not one.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: Ask whoever wrote the code originally, @kohsuke I suppose. I was merely deleting an `if`-block.

jglick:
> I've tried the PR and does what is described in the PR.

Well the change to the build log was really just a side change. The main point was a fixed API, which is used by downstream PRs.

svanoort:
:bee:  I think, though I'm not 100% sure I understand all of this

recena:
@jglick In fact, I have several of his PR in my local env

PR #4, opened 2016-04-10
[JENKINS-34139] - Document potential requirements to masters in CpsScmFlowDefinition

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-34139

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
@@ -4,4 +4,12 @@
     (If you wish to use other files from the same repository during your Pipeline,
     you will need to check them out separately on some slave;
     this checkout cannot be reused.)
+    
+    This Pipeline script provider uses the functionality provided in standalone SCM plugins.
+    In particular cases these plugins may require custom environment on the Jenkins master in order 
+    to get the data from the configured SCM.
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
```diff
@@ -4,4 +4,12 @@
     (If you wish to use other files from the same repository during your Pipeline,
     you will need to check them out separately on some slave;
     this checkout cannot be reused.)
+
```
jglick: :ant: I think you forgot a `<p>` here.

jglick:
Rather than guiding the user toward satisfying these snowflake conditions on the Jenkins master, it is better advice to remind them that `CpsScmFlowDefinition` as a whole is merely a convenience, and that [you can customize the behavior, including selecting a specific node, by doing the checkout explicitly](https://github.com/jenkinsci/pipeline-plugin/blob/72a4ac9c280013c8a68a7b8b24b9b1a8d6ea045a/TUTORIAL.md#triggering-manual-loading).

amuniz:
:bee:

jglick:
:bug: as above

oleg-nenashev:
@jglick I barely remember what it was about. Do you think https://issues.jenkins-ci.org/browse/JENKINS-34139 worth fixing in this or any other way?

PR #6, opened 2016-04-12
[JENKINS-25894] Fix reporting of errors from parallel step

jglick:
[JENKINS-25894](https://issues.jenkins-ci.org/browse/JENKINS-25894)

- [X] remove bogus wrapper exception type
- [X] log which branch(es) failed, to restore the information formerly kept in the wrapper exception
- [X] retain stack traces of subsequent branch failures, if not fail-fast
- [x] verify improvements in a test

@reviewbybees esp. @kohsuke


jglick:
Note that this also allows the main patch (to `SandboxContinuable`) in https://github.com/jenkinsci/pipeline-plugin/pull/43 to be reverted without causing a failure in `WorkflowJobNonRestartingTest.sandboxRejection`, since the `RejectedAccessException` is no longer wrapped in `ParallelStepException`. I am not positive this is dead code, though; there might be some other case where `RejectedAccessException` gets wrapped. So for now I am leaving that alone.

recena:
I assume that `ParallelStepException` was not being used in other places, right?

jglick:
[not that I can see](https://github.com/search?ref=simplesearch&type=Code&q=user%3Ajenkinsci+user%3Acloudbees+ParallelStepException)

jglick:
CI build seems to be dying, not sure why, will restart it once again.

jglick:
Build 19 [was successful](https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-cps-plugin/19/) so not sure why the PR builder did not pick that up.

recena:
LGTM :bee:

PR #8, opened 2016-04-27
Enable the sandbox by default

jglick:
@HRMPW suggested turning on sandbox mode by default in all cases, even for Jenkins administrators. (See https://github.com/jenkinsci/pipeline-plugin/pull/67 for the original conditional logic.) There are several arguments for doing this:

* At least pending [JENKINS-28178](https://issues.jenkins-ci.org/browse/JENKINS-28178), if you want to convert to `Jenkinsfile` (whether for multibranch support, or simply to keep single-project sources in SCM) you will need to have a sandboxable script anyway, so you might as well start learning Pipeline this way.
* The sandbox will catch certain common beginner mistakes, such as
```groovy
node('slave') {
  checkout …
  def text = new File('something-in-workspace').text
}
```
where `readFile` was in fact needed.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
@@ -28,7 +28,7 @@
     <wfe:workflow-editor />
   </f:entry>
   <f:entry field="sandbox">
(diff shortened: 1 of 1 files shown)


HRMPW:
I mentioned this to @kohsuke over a :beer: a few weeks back as well so I hope he still agrees. For the reasons cited above I like this change.  Users can learn to write Pipeline scripts that can be easily shared or put into their SCM but the exception to this rule would be for users who need a reason to break out of the sand box. I think NOT using the sandbox should be the exception.  :bee:

PR #11, opened 2016-05-12
[JENKINS-26107] Examples for blocky stage

jglick:
For after https://github.com/jenkinsci/pipeline-stage-step-plugin/pull/5 is released.

@reviewbybees esp. @kohsuke


src/main/js/samples.js
```diff
@@ -46,31 +46,25 @@ samples.push({
     name: 'hello',
     title: 'Hello World',
     script: "node {\n" +
-        "   stage 'Stage 1'\n" +
```
jglick: Just seemed gratuitous here.

src/main/webapp/snippets/workflow.js
```diff
@@ -62,35 +62,24 @@ snippet for++\n\
...
-# node\n\
-snippet node\n\
-	node {\n\
-	    ${1}\n\
-	}\n\
-# node-l ('label') \n\
-snippet node-l\n\
-	node ('${1}') {\n\
-	    \n\
-	}\n\
```
jglick: redundant

src/main/webapp/snippets/workflow.js
```diff
@@ -62,35 +62,24 @@ snippet for++\n\
...
+	}\n\
 # node\n\
 snippet node\n\
 	node {\n\
 	    ${1}\n\
 	}\n\
 # node-l ('label') \n\
 snippet node-l\n\
-	node ('${1}') {\n\
+	node('${1}') {\n\
```
jglick: Bringing snippet style in line with what `Snippetizer` and most docs would suggest.

kohsuke:
:bee:

src/main/js/samples.js
```diff
@@ -45,38 +45,37 @@ function getSample(sampleName) {
 samples.push({
     name: 'hello',
     title: 'Hello World',
-    script: "node {\n" +
-        "   stage 'Stage 1'\n" +
```
jglick: Just seemed gratuitous here.

src/main/js/samples.js
```diff
@@ -45,38 +45,37 @@ function getSample(sampleName) {
...
+        "   }\n" +
+        "   stage('Build') {\n" +
+        "      // Run the maven build\n" +
+        "      if (isUnix()) {\n" +
+        "         sh \"'${mvnHome}/bin/mvn' -Dmaven.test.failure.ignore clean package\"\n" +
+        "      } else {\n" +
+        "         bat(/\"${mvnHome}\\bin\\mvn\" -Dmaven.test.failure.ignore clean package/)\n" +
+        "      }\n" +
+        "   }\n" +
+        "   stage('Results') {\n" +
```
jglick: Adding a new stage here.

svanoort:
@jglick Block scoped stage is now live and hitting the update center, can we go ahead and merge?

jglick:
Yup, let me just do one last sanity check…

pom.xml
```diff
@@ -153,12 +153,6 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>pipeline-stage-step</artifactId>
```
svanoort: 🐛  you'll still need a dependency on pipeline-stage-step 2.2 for the example to actually run.

jglick: Actually running the example requires all sorts of plugins which `workflow-cps` does not depend on. If we could define this snippets in another plugin like `workflow-aggregator`, I would do that, but @tfennelly did not write such an extension point.

jglick: Anyway this was a `test`-scoped dependency to begin with, which was only needed by the previous version of `SnippetizerTest`.

svanoort: Ah, okay that makes it clearer.  Okay, retracting my bug and granting the 🐝 instead!

PR #12, opened 2016-05-20
[JENKINS-31831] Make Snippetizer UI be displayed in a top-level page

jglick:
[JENKINS-31831](https://issues.jenkins-ci.org/browse/JENKINS-31831)

Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/3.

- [X] basic split
- [X] make per-job context work
- [x] create introductory page with syntax
- [x] get the snippetizer link to appear, e.g., from multibranch Pipelines

From the sidebar of a simple Pipeline job, or (with the downstream PR) from a multibranch Pipeline or organization folder, you will see a new link:

config-screen

The page first presented is the _Snippet Generator_, except now full-screen. Note that its behavior will in general vary slightly depending on where you started it from; for example, _Credentials_ pulldowns may offer different values. (You _can_ go directly to `/jenkins/pipeline-groovy/` as a permalink, but there will be no contextual information in that case.)

snippetizer

The reference documentation is a separate page, with a little introductory text added.

reference

And there are links to open the online documentation (from jenkins.io) and the GDSL in separate windows.

@reviewbybees


HRMPW:
Should we call the link something other than "Pipeline Groovy?" The reason snippet generator and the reference exists is to help learn Pipeline DSL specific steps that aren't part of Groovy itself. We could just make link label to be "Pipeline DSL" and the url path as "pipeline-dsl"

jglick:
I am fine with “Pipeline DSL”, though I was not sure everyone would know what a DSL was, either.

amuniz:
LGTM :bee: (tested locally).

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -247,4 +235,40 @@ public int compare(StepDescriptor o1, StepDescriptor o2) {
...
 
+    @Restricted(DoNotUse.class)
+    @Extension public static class PerJobAdder extends TransientActionFactory<Job> {
+
+        @Override public Class<Job> type() {
+            return Job.class;
+        }
+
+        @Override public Collection<? extends Action> createFor(Job target) {
+            // TODO probably want an API for FlowExecutionContainer or something
```
amuniz: Or move this factory and `LocalAction` to `workflow-job`?

jglick:
“Pipeline Syntax” perhaps?

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: Would make things worse for two reasons:

* `LocalAction` has to extend `Snippetizer`, which would force `workflow-job` to get a dependency on `workflow-cps`.
* Downstream PRs show callers which currently have a `workflow-cps` dependency anyway which would need to add a `workflow-job` dependency.

HRMPW:
👍  on "Pipeline Syntax"  

🐝

kzantow:
... or "Pipeline Script" ? ... since it goes in the "Pipeline Script" text box...

kzantow:
... or "Pipeline Script Generator", if you're referring to the link name. anyway, :bee:

jglick:
@kzantow well the link takes you to a set of pages only the first of which is the snippet generator.

amuniz:
:bee:

kzantow:
Ah right. How about 'Pipeline Help' or 'Pipeline Generator & Help' ? 'Pipeline Syntax' sounds like an explanation of the Groovy language.

jglick:
> Pipeline Help

Possible, though `Snippetizer` is more of a “live” tool, so I would be concerned this would give the wrong impression.

> Pipeline Generator & Help

Probably too verbose for use from the sidebar.

jglick:
At any rate, I think people can tweak the UI labels in various ways in the future as desired. Probably `http://jenkins/job/whatever/pipeline-syntax/` is an adequate URL.

kzantow:
:bee:

PR #13, opened 2016-05-24
[JENKINS-31831] Amendment to display a link in the config form

jglick:
@HRMPW noticed that #12 is not great in Jenkins 2.

Multibranch pipelines and organization folders are unaffected, at least until [JENKINS-33622](https://issues.jenkins-ci.org/browse/JENKINS-33622) is fixed.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
@@ -30,4 +30,7 @@
   <f:entry field="sandbox">
     <f:checkbox title="${%Use Groovy Sandbox}" default="true"/>
   </f:entry>
+    <f:block>
+        <a href="pipeline-syntax" target="_blank">${%Pipeline Syntax}</a>
+    </f:block>
 </j:jelly>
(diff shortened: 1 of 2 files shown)


HRMPW:
:bee: looks simple enough

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
```diff
@@ -30,4 +30,7 @@
   <f:entry field="sandbox">
     <f:checkbox title="${%Use Groovy Sandbox}" default="true"/>
   </f:entry>
+    <f:block>
+        <a href="pipeline-syntax" target="_blank">${%Pipeline Syntax}</a>
+    </f:block>
```
amuniz: Format (indentation).

amuniz:
:bee:

amuniz:
Does not this deserve a release by itself?
We have no snippet generator on any fresh 2.x install.

jglick:
You do, but it is less discoverable. It is in the sidebar when _not_ configuring the job, or in the bread crumb when you are.

At any rate, I do not mind cutting a release for it.

PR #14, opened 2016-05-31
[JENKINS-34064] Making fix work in Jenkins 2

jglick:
Downstream of https://github.com/cloudbees/groovy-cps/pull/24.

@reviewbybees


jglick:
`deploy`ed the snapshot but it went to `oss.sonatype.org` which apparently the CI builds do not read from.

jglick:
Seems

    mvn -Dtest=org.jenkinsci.plugins.workflow.SerializationTest#eachClosure -Djenkins.version=2.8 surefire:test

sometimes passes, sometimes fails.

jglick:
Closing in favor of #124.

PR #16, opened 2016-06-06
Fix some test failures on Jenkins 2

jglick:
As with

    mvn -Djenkins.version=2.8 clean package

@reviewbybees esp. @kwhetstone

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
@@ -107,7 +107,9 @@ public class WorkflowTest extends SingleJobTestBase {
         assertNotNull(e);
         assertEquals(e, b.getExecutor());
         assertTrue(e.isActive());
+        /* TODO seems flaky:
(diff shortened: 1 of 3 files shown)


kwhetstone:
:bee:

src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
```diff
@@ -107,7 +107,9 @@ private void liveness() {
         assertNotNull(e);
         assertEquals(e, b.getExecutor());
         assertTrue(e.isActive());
+        /* TODO seems flaky:
```
amuniz: Is not the auto-retry flaky tests there to avoid this kind of changes?

jglick: It is to avoid having an entire plugin build (for example, prior to release, or on an unrelated PR) fail due to a _rare_ error. But I saw this flake several times in a row. It looks like a race condition in the test. Until the test is improved to pass more reliably, I would rather suppress the flakiness.

PR #17, opened 2016-06-07
[FIXED JENKINS-35395] Re-add Global Variables docs.

abayer:
https://issues.jenkins-ci.org/browse/JENKINS-35395

cc @reviewbybees, @jglick

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -90,6 +90,20 @@ THE SOFTWARE.
                         }
                     </script>
                 </f:block>
+                <f:section title="Global variables"/>
+                <f:dropdownList name="var" title="Variable">
+                    <j:forEach var="var" items="${it.globalVariables}">
+                        <f:dropdownListBlock title="${var.name}" lazy="var,it">
+                            <l:ajax>
+                                <f:block>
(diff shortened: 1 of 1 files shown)


HRMPW:
Thanks @abayer . Several people have commented on the absence of this link already.

jglick:
:-1:, this was a deliberate omission. Identical content is already available in the reference tab. It was only ever beneath the “active” part of _Snippet Generator_ for lack of a better place to put it.

jglick:
I am fine with reorganizing the content in various ways, including renaming _Snippet Generator_ to emphasize that it is about preparing examples of _step call_ syntax, and/or dividing the static reference into multiple screens, and/or changing the navigational system of the single reference screen to collapse everything by default while providing a separate “print mode” link, etc., just so long as there is not this duplication.

abayer:
In practice the Snippet Generator is the default reference spot for most people at this point, so a lot of people had a response to the Global Variables no longer being available on the same page as the Snippet Generator along the lines of "Uh, there's no docs on global variables any more?"

rtyler:
> So if we're not going to have the global variable docs here, I think we do need a clear link to them from here.

:+1:  

In my local Jenkins installation the `/pipeline-syntax/html` page which is linked in the sidebar as "Reference" is 55k pixels tall, and I genuinely didn't know that way way way way down at the bottom was the Global Variable documentation which previously existed directly under the Snippet Generator.

abayer:
Replaced this with #18

PR #18, opened 2016-06-08
[FIXED JENKINS-35395] Copy Global Variable docs into their own page.

abayer:
...and link to it from pipeline-syntax/index.jelly.

https://issues.jenkins-ci.org/browse/JENKINS-35395

Replaces #17 - I tried an anchor link to `pipeline-syntax/html#globalVariables` (after adding that anchor, of course) and it didn't work nicely - it always ended up focusing on a different location than it should have when rendering. So I split it out to a separate page instead.

cc @reviewbybees, esp. @jglick and @rtyler

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
@@ -1,22 +1,10 @@
 package org.jenkinsci.plugins.workflow.cps.Snippetizer
 
-import org.jenkinsci.plugins.structs.describable.ArrayType
-import org.jenkinsci.plugins.structs.describable.AtomicType
-import org.jenkinsci.plugins.structs.describable.DescribableModel
-import org.jenkinsci.plugins.structs.describable.DescribableParameter
-import org.jenkinsci.plugins.structs.describable.EnumType
-import org.jenkinsci.plugins.structs.describable.ErrorType
-import org.jenkinsci.plugins.structs.describable.HeterogeneousObjectType
-import org.jenkinsci.plugins.structs.describable.HomogeneousObjectType
-import org.jenkinsci.plugins.structs.describable.ParameterType
+
 import org.jenkinsci.plugins.workflow.cps.GlobalVariable
 import org.jenkinsci.plugins.workflow.cps.Snippetizer
-import org.jenkinsci.plugins.workflow.steps.StepDescriptor
 
 import javax.servlet.RequestDispatcher
-
-// keeps track of recursion inside generateHelp
-stack = new Stack();
 
 Snippetizer snippetizer = my;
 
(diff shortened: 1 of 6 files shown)


jglick:
:-1:, now this just duplicates content from the `html` view. I want there to be one place for each thing (excepting perhaps a “print mode” view with no chrome).

bitwiseman:
@jglick - information being available via multiple paths is not a bad thing, is it?

jglick:
> It sounds like you disagree with the whole intent of this PR.

Not really, just with the fact of duplication. It would be fine if there were one designated place you went to for step documentation and one designated place you went to for variable documentation.

abayer:
@jglick Ok, now it's all split up - no more duplication. Thoughts?

abayer:
Whoops, will fix test failure shortly.

abayer:
Test fixed.

bitwiseman:
Many key features are accessed through global variables not steps. The link for the Global Variables should go above the steps and in it's own section with a helpful description.

abayer:
So you want it above the steps link in the side panel? Not sure what you mean by "in its own section with a helpful description", though.

bitwiseman:
screen shot 2016-06-13 at 9 43 20 am

This is landing page, I see when I open the snippet generator (ignoring the side panel). This link at the bottom is what I'm talking about.  As I said earlier, the distinction between global variables and steps is one that users who come to this page may or may not understand.  If we're going to make the distinction, we need to explain it to them and present it clearly from the start.   

Trying it out, I think the location of the link maybe fine, but it can't just be sitting there bare. 

So, the page for "Snippet generator" should look something like: 


# Overview
This "Snippet Generator" will help you learn the Groovy code which can be used to define various steps. Pick a step you are interested in from the list, configure it, click Generate Groovy, and you will see a Groovy statement that would call the step with that configuration. You may copy and paste the whole statement into your script, or pick up just the options you care about. (Most parameters are optional and can be omitted in your script, leaving them at default values.)  

## Steps
(snippet generator here)


## Global Variables
There are many features of the Pipeline that are not steps.  These are often exposed via global variables, which are not supported by the snippet generator.  See the "Global Variable reference" for details.

bitwiseman:
Also, should the side panel say "Global variable reference" instead of "Variables reference"?
screen shot 2016-06-13 at 10 18 29 am

abayer:
Went with "Variables Reference" to avoid being too damned long and breaking rendering, but I'm good either way. =) Adding the text you suggested for the snippet generator page.

abayer:
@bitwiseman Updated. Let me know what you think - @jglick too.

abayer:
Build failure for no good reason - closing/reopening.

bitwiseman:
Looks better to me.  I'm still not loving the global variables not being part of snippet generator, but that's a separate issue.  This looks good to me.

abayer:
@bitwiseman then you should a `:bee:`. =)

abayer:
i.e., see https://github.com/reviewbybees/about#about-reviewbybees

bitwiseman:
@abayer - ah, I see.  :bee:

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
```diff
@@ -0,0 +1,169 @@
...
+        }
+      }
+    }
+  }
+}
+
+    }
+}
+
+def generateHelp(DescribableModel model, int headerLevel) throws Exception {
```
jglick: :bug: you seem to have just copied all this stuff, but it is not needed here.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
```diff
@@ -0,0 +1,169 @@
...
+import org.jenkinsci.plugins.structs.describable.HomogeneousObjectType
+import org.jenkinsci.plugins.structs.describable.ParameterType
+import org.jenkinsci.plugins.workflow.cps.GlobalVariable
+import org.jenkinsci.plugins.workflow.cps.Snippetizer
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor
+
+import javax.servlet.RequestDispatcher
+
+// keeps track of recursion inside generateHelp
+stack = new Stack();
```
jglick: ditto

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/html.groovy
```diff
@@ -24,59 +24,36 @@ def l = namespace(lib.LayoutTagLib)
 def st = namespace("jelly:stapler")
 
 l.layout(title:_("Pipeline Syntax: Reference"), norefresh: true) {
-    st.include(page: 'sidepanel')
```
jglick: Not clear what you are doing to this page, which IIUC should not have been affected.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
abayer: Argh, thought I'd removed the irrelevant stuff. Fixing.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/html.groovy
abayer: That'd be a typo, yeah. Fixing.

jglick: Ah, [whitespace-ignoring diff](https://github.com/jenkinsci/workflow-cps-plugin/pull/18/files?w=1) makes it clearer that you are just deleting a block from this page. So maybe just `git checkout master -- src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/html.groovy` and do that deletion so that the PR diff shows it naturally.

abayer: Hrm, actually, I'm not seeing that missing. Lemme poke.

abayer: Yeah, fixing that - not sure why it went so weird.

abayer:
There we go.

jglick:
:bee:

kwhetstone:
:bee:  I like avoiding the infinite scroll to get to the variables section.  Looks like a good solution to me.

PR #20, opened 2016-06-10
[JENKINS-26130] Allow pickle rehydration to be interrupted

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/5, https://github.com/jenkinsci/workflow-support-plugin/pull/4, and https://github.com/jenkinsci/workflow-job-plugin/pull/5.

@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -749,25 +753,42 @@ public void addListener(GraphListener listener) {
...
             @Override
             public void onFailure(Throwable t) {
+                LOGGER.log(Level.WARNING, "failed to interrupt steps in " + owner, t);
             }
         });
+
+        // If we are still rehydrating pickles, try to stop that now.
+        Collection<ListenableFuture<?>> futures = pickleFutures;
+        if (futures != null) {
+            LOGGER.log(Level.FINE, "we are still rehydrating pickles in {0}", owner);
```
oleg-nenashev: 🐜 lowercase again. In this method both lowercase and uppercase letters are being used, so it is a kind of mess

oleg-nenashev:
Mostly LGTM

oleg-nenashev:
re-:bee:

PR #22, opened 2016-06-17
Better Maven sample

jglick:
Otherwise you get builds marked as failed due to JUnit failures, which looks crappy. With the patch:

pipeline-with-tests

@reviewbybees


abayer:
:bee:

PR #23, opened 2016-06-17
Show Snippetizer link also from ReplayAction

jglick:
Like #13, displays a link to _Pipeline Syntax_ from with the _Replay_ screen, one of the places you are most likely to need it.

replay-syntax

@reviewbybees


abayer:
:bee:

PR #27, opened 2016-06-28
[JENKINS-36289] Added ReplayAction#run2

imeredith:
Added run2() to ReplayAction which returns Queue.Item rather than
QueueTaskFuture, so that Blue Ocean can reference the item in the
queue for its REST api.

Note: I'm not sure what the best way to do this was so I added run2 rather than modifying run's return signature.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
@@ -171,7 +171,7 @@ public class ReplayAction implements Action {
     }
 
     /**
-     * For use in projectis that want initiate a replay via the Java API.
+     * For use in projects that want initiate a replay via the Java API.
      *
      * @param replacementMainScript main script; replacement for {@link #getOriginalScript}
      * @param replacementLoadedScripts auxiliary scripts, keyed by class name; replacement for {@link #getOriginalLoadedScripts}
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -165,6 +166,18 @@ public void doRun(StaplerRequest req, StaplerResponse rsp) throws ServletExcepti
      * @return a way to wait for the replayed build to complete
      */
     public @CheckForNull QueueTaskFuture/*<Run>*/ run(@Nonnull String replacementMainScript, @Nonnull Map<String,String> replacementLoadedScripts) {
+        Queue.Item item = run2(replacementMainScript, replacementLoadedScripts);
+        return item == null ? null : item.getFuture();
+    }
+
+    /**
+     * For use in projectis that want initiate a replay via the Java API.
```
michaelneale: projectis/project?

michaelneale:
Seems like a needed feature - I wonder if "run2" can have a better name (really the return type gives the purpose away, which is something people will see when they use content-assist in an IDE and type "run.." so having the method start with run makes sense to me.

michaelneale:
🐝  pending 🐜  for javadoc typo

michaelneale:
@imeredith this needs a corresponding JIRA ticket

src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: :ant: `projects` I think was meant.

jglick:
The name is a little weird but good enough. :bee:

michaelneale:
@jglick I propose win32 api conventions of runEx runExA runEx2A etc ;)

PR #28, opened 2016-06-30
[JENKINS-29922] $class must die

kohsuke:
- [x] allow `Describable` to be run like a step via meta-step
- [x] handle nested `Describable` correctly
- [x] handle single argument construction like `f("foo")`
- [x] split arguments to `Describable` and metastep
- [x] write some test cases
- [x] update `Snippetizer`
- [x] add some more snippetizer tests


kohsuke:
See [JENKINS-29922](https://issues.jenkins-ci.org/browse/JENKINS-29922) for all the other relevant PRs that span across multiple plugins.

kohsuke:
This PR and its associated PR is all ready

@reviewbybees

abayer:
:bee:

pom.xml
```diff
@@ -158,6 +163,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>junit</artifactId>
+            <version>1.15-SNAPSHOT</version>
```
jglick: :bug: update

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -106,6 +117,24 @@
...
 
+    @Test public void coreStepWithSymbol() throws Exception {
+        JUnitResultArchiver aa = new JUnitResultArchiver("target/surefire/*.xml");
+        aa.setAllowEmptyResults(true);
+        assertRoundTrip(new CoreStep(aa), "junit allowEmptyResults: true, testResults: 'target/surefire/*.xml'");
+    }
+
+    @Test public void coreStepWithSymbolWithSoleArg() throws Exception {
+        JUnitResultArchiver aa = new JUnitResultArchiver("target/surefire/*.xml");
+        assertRoundTrip(new CoreStep(aa), "junit 'target/surefire/*.xml'");
```
jglick: Is this not duplicating stuff already tested (or which should be tested) in `CoreStepTest`?

pom.xml
```diff
@@ -112,7 +117,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-basic-steps</artifactId>
-            <version>1.15</version>
+            <version>2.1-SNAPSHOT</version>
```
jglick: :bug: this will be a cyclic dependency I think. That is why I purposefully kept tests specific to the `core` step in `CoreStepTest`.

pom.xml
jglick: (or just delete; see above)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
```diff
@@ -46,6 +47,9 @@
 
 /**
  * The script of a workflow.
+ *
+ * Every {@link Script} we load in Pipeline execution derives from this subtype.
```
jglick: What does this have to do with this PR? Revert.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -175,6 +195,97 @@ public Object invokeMethod(String name, Object args) {
...
+                DescribableModel<?> dm = new DescribableModel(d.clazz);
+                DescribableParameter p = mm.getFirstRequiredParameter();
+
+                // order of preference:
+                //      1. mandatory parameter in mm
+                //      2. mandatory parameter in dm
+                //      3. other parameters in mm
+                //      4. other parameters in dm
+                // mm is preferred over dm because that way at least the arguments that dm defines
+                // act consistently
```
jglick: You mean, the arguments that `dm` defines. Right?

For example, `polling` should always mean the same thing to any call to `checkout`, regardless of any `polling` parameter that the `SCM` might have.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -175,6 +195,97 @@ public Object invokeMethod(String name, Object args) {
...
+        }
+    }
+
+    /**
+     * Finds a meta step that can handle a Describable of the given type 'd'
+     */
+    private StepDescriptor findMetaStep(Descriptor d) {
+        OUTER:
+        for (StepDescriptor sd : StepDescriptor.allMeta()) {
+            if (sd.getMetaStepArgumentType().isAssignableFrom(d.clazz)) {
```
jglick: Seems like it would be simpler to delete `allMeta` and do the null check here instead.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -62,105 +65,235 @@
...
     static String object2Groovy(Object o) throws UnsupportedOperationException {
-        Class<? extends Object> clazz = o.getClass();
+        return object2Groovy(new StringBuilder(),o, false).toString();
+    }
+
+    /**
+     * Renders the invocation syntax to re-create a given object 'o' into 'b'
+     *
+     * @param nestedExp
+     *      true if this object is written as a nested expression (in which case we always produce parenthesis for readability
```
jglick: missing `)`

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -62,105 +65,235 @@
...
+                    DescribableParameter p = m.getFirstRequiredParameter();
+                    Object wrapped = uninst.getArguments().get(p.getName());
+                    if (wrapped instanceof UninstantiatedDescribable) {
+                        UninstantiatedDescribable nested = (UninstantiatedDescribable) wrapped;
+                        TreeMap<String, Object> copy = new TreeMap<>(nested.getArguments());
+                        for (Entry<String, ?> e : uninst.getArguments().entrySet()) {
+                            if (!e.getKey().equals(p.getName())) {
+                                if (copy.put(e.getKey(),e.getValue())!=null) {
+                                    // collision between a parameter in meta-step and wrapped-step,
+                                    // which cannot be reconcilled unless we explicitly write out
```
jglick: typo

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -62,105 +65,235 @@
...
-                        first = false;
-                        if (d.takesImplicitBlockArgument() || singleMap) {
-                            b.append('(');
-                        } else {
-                            b.append(' ');
+                UninstantiatedDescribable uninst = d.uninstantiate(step);
+
+                if (d.isMetaStep()) {
+                    // if we cannot represent this 'o' in a concise syntax that hides meta-step, set this to true
+                    boolean failSimplification = false;
```
jglick: I think there are some cases you are forgetting here. For example, there is a `@Symbol` but it clashes with a `StepDescriptor.functionName`. Or there are two different metasteps which can both accept the same `@Symbol`.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -62,105 +65,235 @@
...
+        //   a block argument normally requires a () around arguments, and if arguments are empty you need explicit (),
+        //   but not if both is the case!
+        final boolean needParenthesis = (blockArgument ^ args.isEmpty()) || isSingleMap(args) || nested;
+
+        b.append(ud.getSymbol());
+        b.append(needParenthesis ? '(': ' ');
+
+        if (ud.hasSoleRequiredArgument()) {
+            // lone argument optimization, which gets rid of named arguments and just write one value, like
+            // retry 5 { ... }
```
jglick: Does that actually work?

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -62,105 +65,235 @@
...
+        }
+
+        return b;
+    }
+
+    /**
+     * If the sole argument is a map, its [...] bracket cannot be present.
+     *
+     * Historically we've disambiguated this by adding (...) around the function call.
+     * TODO: I claim removing both () and [] would be better.
```
jglick: OK, well make that change and write a test running the script using that syntax to prove that `DSL` actually gets the write parameters.

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -125,34 +154,43 @@
...
-                    }
-                } else {
-                    return super.invokeMethod(name, args);
+        CompilerConfiguration cc = new CompilerConfiguration();
+        cc.setScriptBaseClass(DelegatingScript.class.getName());
+        GroovyShell shell = new GroovyShell(r.jenkins.getPluginManager().uberClassLoader,new Binding(),cc);
+
+        DelegatingScript s = (DelegatingScript) shell.parse(expected);
+        s.o = new DSL(new DummyOwner()) {
+            // for testing, instead of executing the step just return an instantiated Step
```
jglick: Would it not be easier (and more convincing) to just define actual `Step`s?

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -125,34 +154,43 @@
         BuildTriggerStep step = new BuildTriggerStep("downstream");
         assertRoundTrip(step, "build 'downstream'");
         step.setParameters(Arrays.asList(new StringParameterValue("branch", "default"), new BooleanParameterValue("correct", true)));
+        // when this test fails because of new core, see below "buildTriggerStepWith_2_2"
```
jglick: :bug: due to `plugin-compat-tester` we do not wish to write tests which we _know_ will fail with a newer core. Make it pass on any sufficiently new core version, for example by checking `StringParameterValue.class.getAnnotation(Symbol.class) != null` and selecting the expected assertion accordingly.

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
kohsuke: I don't think `CoreStepTest` runs snippetizer.

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
jglick: Or what was wrong with the original trick of evaluating `steps.something`?

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/StateMetaStep.java
```diff
@@ -0,0 +1,71 @@
...
+    public final State state;
+    public boolean moderate;
+
+    @DataBoundConstructor
+    public StateMetaStep(State state) {
+        this.state = state;
+    }
+
+    @DataBoundSetter
+    public void setModerate(boolean m) {
```
jglick: Might be useful to also add an (unrelated) `moderate` parameter to one of the `State` implementations, so as to verify that

```groovy
california ocean: 'pacific', mountain: 'sierra', moderate: true
```

behaves the same as

```groovy
state state: california(ocean: 'pacific', mountain: 'sierra'), moderate: true
```

but differently from

```groovy
state state: california(ocean: 'pacific', mountain: 'sierra', moderate: true)
```

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
kohsuke: Should be another PR

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
kohsuke: (1) it doesn't work with nested use of symbols, and (2) duplicate code that interprets method name.

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
kohsuke: As more things get symbols there will be a lot of them.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/StateMetaStep.java
kohsuke: Added

kohsuke:
Outstanding issues:
- [ ] `@Symbol` that clashes with a `StepDescriptor.functionName`
- [ ] Or there are two different metasteps which can both accept the same `@Symbol`

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
kohsuke: It's a harmless doc that came up during our hackathon. It's OK.

kohsuke:
I claim code complete

kohsuke:
Come on Jenkins, let's try this again

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
jglick: Yes, but we should fix the cases we know about now, and continue to fix others as they become known.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
```diff
@@ -56,4 +57,74 @@
...
+    @Issue("JENKINS-29922")
+    @Test
+    public void dollar_class_must_die_onearg() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "die4");
+        p.setDefinition(new CpsFlowDefinition("newYork 'Empire'"));
+        r.assertLogContains("The Empire State", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
+    }
+
+    @Issue("JENKINS-29922")
+    @Test public void runMetaBlockStep() throws Exception {
```
abayer: Would still like to see tests/how this would look for `properties`, but that's probably fairly special-casing and would likely be in `workflow-multibranch-plugin` anyway.

abayer:
Only question I have is probably a pretty simple one - this won't break existing `step([[$class: ... ]])` invocations, right?

abayer:
Answered by tests over at https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/10 so we're good - :bee:

jglick:
> this won't break existing `step([[$class: ... ]])` invocations, right?

Right, those will continue to work as before, and in fact _Snippet Generator_ will still offer that syntax when it cannot do better.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
jglick: IIRC @kohsuke did some temporary sanity checks before reverting, but you are right it would be helpful to have these tests codified. I will play with it.

kohsuke:
:bee:

PR #30, opened 2016-06-30
Extracted forRun helper method

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -114,13 +114,19 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
         owner = r;
     }
 
-    public static @Nonnull EnvActionImpl forRun(@Nonnull Run<?,?> run) {
-        EnvActionImpl action = run.getAction(EnvActionImpl.class);
-        if (action == null) {
-            action = new EnvActionImpl();
-            run.addAction(action);
+    /**
+     * Gets the singleton instance for a given build, creating it on demand.
+     */
+    public static @Nonnull EnvActionImpl forRun(@Nonnull Run<?,?> run) throws IOException {
+        synchronized (run) {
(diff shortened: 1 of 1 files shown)


amuniz:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -113,19 +114,23 @@ private EnvActionImpl() {
         owner = r;
     }
 
+    public static @Nonnull EnvActionImpl forRun(@Nonnull Run<?,?> run) {
```
oleg-nenashev: Show me the docs ;)

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -113,19 +114,23 @@ private EnvActionImpl() {
...
+        EnvActionImpl action = run.getAction(EnvActionImpl.class);
+        if (action == null) {
+            action = new EnvActionImpl();
+            run.addAction(action);
```
oleg-nenashev: Without save?

oleg-nenashev:
🐜 since the code seems to be at risk of concurrency issues.
Does not decrease the technical debt, hence 🐝 if you want to merge it ASAP

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
jglick: Suppose it could `run.save()`. The build is likely to get saved for some other reason soon afterwards anyway.

I would note that numerous calls to `Run.addAction` in Jenkins core decline to call `save()`, with no apparent ill effect. Probably this override could call `save()` on itself.

jglick:
> the code seems to be at risk of concurrency issues

Doubtful. The expected call of the new method would happen before the build starts, so not a risk. The _original_ call could only happen from the CPS VM thread, which uses coöperative multitasking, so also not a risk.

> Does not decrease the technical debt

You mean _increase_?

abayer:
:bee:

PR #32, opened 2016-07-11
[JENKINS-25736] Ability to pause a running flow

jglick:
[JENKINS-25736](https://issues.jenkins-ci.org/browse/JENKINS-25736)

Originally filed by @kohsuke as https://github.com/jenkinsci/pipeline-plugin/pull/195.

- [X] basic implementation
- [X] make it survive a restart
- [X] also solve [JENKINS-32015](https://issues.jenkins-ci.org/browse/JENKINS-32015)
- [x] provide UI

@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/cps/PauseUnpauseAction.java
```diff
@@ -0,0 +1,58 @@
...
+        return null; // special presentation
+    }
+
+    @Override public String getDisplayName() {
+        return null; // special presentation
+    }
+
+    @RequirePOST
+    public void doToggle() {
+        execution.pause(!execution.isPaused());
```
jtnord: No authentication check?  would expect at least build..

jtnord:
what appears to be a :bug: as it looks to me like anyone can pause/unpause a pipeline.

src/main/java/org/jenkinsci/plugins/workflow/cps/PauseUnpauseAction.java
jglick: Good point, I think this needs a permissions check, will add one.

jtnord:
:bee:

abayer:
:bee:

PR #33, opened 2016-07-26
[JENKINS-34650] Added a trusted classloader that runs CPS code outside sandbox

kohsuke:
See `doc/classloader.md` for the details.


kohsuke:
This change depends on https://github.com/cloudbees/groovy-cps/pull/36

kohsuke:
Pushed groovy-cps:1.8-SNAPSHOT. retrying.

kohsuke:
I guess there's no easy way to hook up this PR build into CB's snapshot repo.

kohsuke:
Deployed SNAPSHOT to repo.jenkins-ci.org now, so we should be able to see the status now.

doc/classloader.md
```diff
@@ -0,0 +1,44 @@
...
+
+Additionally, scripts loaded in RCL lives in the security sandbox
+(unless the user opts out of it, in which case it requires approval.)
+This classloader is meant to be used to load user-written Groovy scripts,
+which is not trusted.
+
+Scripts loaded in TCL, OTOH, does not live in the security sandbox. This
+classloader is meant to be used to load Groovy code packaged inside
+plugins and global libraries. Write access to these sources should be
+restricted to `RUN_SCRIPT` permission.
```
jglick: :ant: `RUN_SCRIPTS`

doc/classloader.md
```diff
@@ -0,0 +1,44 @@
...
+
+This behaviour is desirable for the situation where script is supplied
+from outside and not readily available at the point of restart.
+For example, `Jenkinsfile` and scripts loaded from the `load` step
+fits this description.
+
+It's also possible to augument `GroovyClassLoader` classpath via
+`addURL()` so that scripts are loaded on-demand whenever needed.
+This is more suitable for scripts that are considered a part of the
+system, such as global libraries or plugin code.
```
jglick: What if a build is started, it begins using some global library, then the library is drastically modified, Jenkins is restarted, and the resumed build tries to use code from the same library? Will the build possibly break?

src/main/java/org/jenkinsci/plugins/workflow/cps/GroovyClassLoaderWhitelist.java
```diff
@@ -31,17 +35,17 @@
      */
     private final Whitelist delegate;
 
-    GroovyClassLoaderWhitelist(GroovyClassLoader scriptLoader, Whitelist delegate) {
-        this.scriptLoader = scriptLoader;
+    GroovyClassLoaderWhitelist(Whitelist delegate, GroovyClassLoader... scriptLoaders) {
+        this.scriptLoaders = Arrays.<ClassLoader>asList(scriptLoaders);
```
jglick: :ant: Make this an `IdentityHashMap<ClassLoader,Void>` so that `permits` is not calling `Object.equals`, which makes me nervous.

src/main/java/org/jenkinsci/plugins/workflow/cps/GroovyShellDecorator.java
```diff
@@ -38,7 +38,26 @@ public void configureCompiler(@CheckForNull CpsFlowExecution context, CompilerCo
      */
     public void configureShell(@CheckForNull CpsFlowExecution context, GroovyShell shell) {}
 
+    /**
+     * Obtains a contextualized {@link GroovyShellDecorator} used to decorate the trusted shell.
+     *
+     * <p>
+     * By default, this method returns null decorator that doesn't do anything.
```
jglick: Not obvious to me that this is a compatible change. In particular, did you try running `docker-workflow` tests against this?

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -340,4 +352,60 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
...
+        }
+    }
+
+    private void trustedShell(final boolean pos) {
+        SECRET = false;
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition("new foo().attempt()", true));
+                WorkflowRun b = story.j.waitForCompletion(p.scheduleBuild2(0).get());
```
jglick: :ant: the `waitForCompletion` call is gratuitous.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -340,4 +352,60 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
...
+                if (pos) {
+                    story.j.assertBuildStatusSuccess(b);
+                    assertTrue(SECRET);
+                } else {
+                    // should have failed with RejectedAccessException trying to touch 'SECRET'
+                    story.j.assertBuildStatus(Result.FAILURE, b);
+                    story.j.assertLogContains("RejectedAccessException: Scripts not permitted to use staticField org.jenkinsci.plugins.workflow.cps.CpsFlowExecutionTest SECRET",b);
```
jglick: :ant: Can do various things to avoid hardcoding this much text here: use `CpsFlowExecutionTest.class.getName()`; create an actual `RejectedAccessException` and `toString` it (IIRC)

jglick:
:bee: AFAIUI

src/main/java/org/jenkinsci/plugins/workflow/cps/GroovyClassLoaderWhitelist.java
kohsuke: `IdentityHashMap<ClassLoader,Void>` obscures the code and I have never seen any funny `ClassLoader.equals()` implementation, so I don't think it's a problem.

doc/classloader.md
kohsuke: Added some more text after this to cover that topic.

src/main/java/org/jenkinsci/plugins/workflow/cps/GroovyShellDecorator.java
kohsuke: Previously there was no trusted classloader, so old plugins like docker-workflow that implements `GroovyShellDecorator` will continue to insert classpath into the untrusted classloader. So I think this is compatible.

I'll run the docker workflow test to verify.

doc/classloader.md
```diff
@@ -41,4 +41,9 @@ It's also possible to augument `GroovyClassLoader` classpath via
 This is more suitable for scripts that are considered a part of the
 system, such as global libraries or plugin code.
 
-
+Those scripts are not snapshotted & persisted with the running program,
+so if they change while the program is running, with or without Jenkins restarts
+in the middle, then the program could fail. (For example, if a program
+call stack includes a class from a global library and that class goes away,
+then the program fails to survive the restart because the call stack cannot
+be deserialized.
```
jglick: missing rparen

doc/classloader.md
```diff
@@ -41,4 +41,9 @@ It's also possible to augument `GroovyClassLoader` classpath via
 This is more suitable for scripts that are considered a part of the
 system, such as global libraries or plugin code.
 
-
+Those scripts are not snapshotted & persisted with the running program,
+so if they change while the program is running, with or without Jenkins restarts
+in the middle, then the program could fail. (For example, if a program
+call stack includes a class from a global library and that class goes away,
+then the program fails to survive the restart because the call stack cannot
```
jglick: …which is a serious flaw IIUC, since the global lib feature makes no attempt to ensure that libraries are not refactored or deleted while builds are running. So its replacement will need to do better than this.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -390,14 +391,16 @@ private void trustedShell(final boolean pos) {
...
+                WorkflowRun b = p.scheduleBuild2(0).get();
                 if (pos) {
                     story.j.assertBuildStatusSuccess(b);
                     assertTrue(SECRET);
                 } else {
                     // should have failed with RejectedAccessException trying to touch 'SECRET'
                     story.j.assertBuildStatus(Result.FAILURE, b);
-                    story.j.assertLogContains("RejectedAccessException: Scripts not permitted to use staticField org.jenkinsci.plugins.workflow.cps.CpsFlowExecutionTest SECRET",b);
+                    story.j.assertLogContains(
+                            new RejectedAccessException("staticField",CpsFlowExecutionTest.class.getName()+" SECRET").getMessage(),
```
jglick: :ant: use `StaticWhitelist.rejectStaticField` to avoid hard-coding the format.

jglick:
re:bee:

doc/classloader.md
kohsuke: I consider that a part of another ticket where we support locking in on a specific commit of a global library. In the documentation I'm merely describing the current behaviour.

I don't consider it a "serious flaw" - the same problem happens if you upgrade plugins that incompatibly change the behaviours, etc.

pom.xml
```diff
@@ -160,7 +160,7 @@
         <dependency>
             <groupId>com.cloudbees</groupId>
             <artifactId>groovy-cps</artifactId>
-            <version>1.6</version>
+            <version>1.8-SNAPSHOT</version>
```
jglick: Unfortunately this pulls in the broken changes in 1.7 reported in JENKINS-34064 which https://github.com/cloudbees/groovy-cps/pull/24/commits/ae43475fe6a814e2562978fcdd03e30a2d8d91b1 tries to deal with.

PR #34, opened 2016-07-27
[JENKINS-37011] step implemented in Groovy

kohsuke:
[JENKINS-37011](https://issues.jenkins-ci.org/browse/JENKINS-37011)

This change depends on https://github.com/jenkinsci/workflow-step-api-plugin/pull/7

### DIFF
--- a/doc/step-in-groovy.md
+++ b/doc/step-in-groovy.md
@@ -1,0 +1,55 @@
+# Implementing Step in Groovy
+The standard way of implementing a step is by writing Java code that extends from `Step` and `StepDescriptor`,
+but you can also define a step in Jenkins Pipeline Script.
+
+This approach is quite convenient when your step composes other
+existing steps. For example, the following `hello.groovy` defines
+a step that prints a greeting message to log:
+
+```
+String getDisplayName() {
+    return 'Greets someone'
+}
+
+void call(String person) {
+    echo "Hello ${person}";
+}
+```
+
+From Jenkins Pipeline you can use this like you'd use any other steps:
+
+```
+hello 'Duke'    // this will print 'Hello Duke'
+```
+
+`src/main/webapp/WEB-INF/steps` is the content root for Groovy files
+to be run by the CPS interpreter (aka Pipeline classpath). You can create regular package
+structures in this content root directory. For example, the above
+helloWorld.groovy could be in `src/main/webapp/WEB-INF/steps/org/kohsuke/jenkins/helloWorld.groovy`
+
+It is also possible to put other supporting classes in this content root directory.
+
+These files will be packaged into the plugin as source files, then
+interpreted at runtime together with user's `Jenkinsfile`.
+
+## Details
+To define a step in Groovy, the following requirements have to be met:
+
+* There must be a public default constructor.
+* There must be a `call()` method that takes parameters that the step needs.
+  This can include a `Closure` parameter, but it has to be the last parameter.
+* There must be a `getDisplayName` method that returns a short human readable name
+  of your step. See `Descriptor#getDisplayName()` for more about it.
+* There must be `src/main/webapp/WEB-INF/steps/META-INF/index` that lists fully-qualified class names
+  of all the steps implemented in Groovy. This is how the runtime discovers all the available steps.
+
+## Groovy steps and sandbox
+Groovy code in Pipeline classpath are trusted,
+so it is not subject to the sandbox execution environment nor the script approval process.
+This means code in Pipeline classpath can invoke any Java code.
+
+User-written `Jenkinsfile` can in turn access any code in Pipeline classpath,
+subject to the normal Java access restriction.
+
+This means a care must be taken to ensure that your code
+does not expose any methods and properties that can be exploited by malicious `Jenkinsfile`.
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -12,6 +12,7 @@ import com.google.common.util.concurrent.FutureCallback;
 import hudson.model.Action;
 import hudson.model.Result;
 import jenkins.model.CauseOfInterruption;
+import org.codehaus.groovy.runtime.InvokerInvocationException;
 import org.jenkinsci.plugins.workflow.actions.BodyInvocationAction;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
@@ -25,6 +26,7 @@ import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
 import javax.annotation.concurrent.GuardedBy;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -116,7 +118,16 @@ class CpsBodyExecution extends BodyExecution {
 
         try {
             // TODO: handle arguments to closure
-            Object x = params.body.getBody(currentThread).call();
+            Object x;
+            try {
+                x = params.body.getBody(currentThread).call();
+            } catch (InvokerInvocationException e) {
+                // depending on how the 'call' method is routed Groovy can wrap this exception.
+                // this happens in CachedMethod.invoke() because it doesn't pass through Error
+                if (e.getCause() instanceof CpsCallableInvocation)
+                    throw e.getCause();
+                throw e;
+            }
 
             // body has completed synchronously. mark this done after the fact
             // pointless synchronization to make findbugs happy. This is already done, so there's no cancelling this anyway.
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -37,6 +37,7 @@ import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.cps.steps.LoadStep;
 import org.jenkinsci.plugins.workflow.cps.steps.ParallelStep;
+import org.jenkinsci.plugins.workflow.cps.steps.ingroovy.StepInGroovy.StepDescriptorInGroovy;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
@@ -116,7 +117,8 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         FlowNode an;
 
         // TODO: generalize the notion of Step taking over the FlowNode creation.
-        boolean hack = d instanceof ParallelStep.DescriptorImpl || d instanceof LoadStep.DescriptorImpl;
+        boolean hack = d instanceof ParallelStep.DescriptorImpl || d instanceof LoadStep.DescriptorImpl
+                    || d instanceof StepDescriptorInGroovy;
 
         if (ps.body == null && !hack) {
             an = new StepAtomNode(exec, d, thread.head.get());
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -60,7 +60,6 @@ import java.util.TreeMap;
 
 import static org.jenkinsci.plugins.workflow.cps.ThreadTaskResult.*;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
-import org.kohsuke.stapler.ClassDescriptor;
 
 /**
  * Scaffolding to experiment with the call into {@link Step}.
@@ -283,9 +282,9 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         return new NamedArgsAndClosure(singleParam(d, arg), null);
     }
     private static Map<String,Object> singleParam(StepDescriptor d, Object arg) {
-        String[] names = new ClassDescriptor(d.clazz).loadConstructorParamNames();
-        if (names.length == 1) {
-            return Collections.singletonMap(names[0], arg);
+        Map<String, Object> m = d.singleArgument(arg);
+        if (m!=null) {
+            return m;
         } else {
             throw new IllegalArgumentException("Expected named arguments but got " + arg);
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -64,10 +64,17 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import jenkins.model.Jenkins;
+import org.jenkinsci.Symbol;
 
 import static org.jenkinsci.plugins.workflow.cps.ThreadTaskResult.*;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
+import org.jvnet.hudson.annotation_indexer.Index;
 
 import org.kohsuke.stapler.ClassDescriptor;
 import org.kohsuke.stapler.NoStaplerConstructorException;
@@ -124,8 +131,22 @@ public class DSL extends GroovyObjectSupport implements Serializable {
             return invokeDescribable(name,args);
         }
 
+        Set<String> symbols = new TreeSet<>();
+        // TODO SymbolLookup only lets us find a particular symbol, not enumerate them
+        try {
+            for (Class<?> e : Index.list(Symbol.class, Jenkins.getActiveInstance().pluginManager.uberClassLoader, Class.class)) {
+                if (Descriptor.class.isAssignableFrom(e)) {
+                    Symbol s = e.getAnnotation(Symbol.class);
+                    if (s != null) {
+                        symbols.addAll(Arrays.asList(s.value()));
+                    }
+                }
+            }
+        } catch (IOException x) {
+            Logger.getLogger(DSL.class.getName()).log(Level.WARNING, null, x);
+        }
         // TODO probably this should be throwing a subtype of groovy.lang.MissingMethodException
-        throw new NoSuchMethodError("No such DSL method '" + name + "' found among " + functions.keySet());
+        throw new NoSuchMethodError("No such DSL method '" + name + "' found among steps " + functions.keySet() + " or symbols " + symbols);
     }
 
     /**
@@ -283,19 +304,6 @@ public class DSL extends GroovyObjectSupport implements Serializable {
                 throw new IllegalArgumentException("Failed to prepare "+symbol+" step",e);
             }
         }
-    }
-
-    /**
-     * Finds a meta step that can handle a Describable of the given type 'd'
-     */
-    private static StepDescriptor findMetaStep(Descriptor d) {
-        OUTER:
-        for (StepDescriptor sd : StepDescriptor.allMeta()) {
-            if (sd.getMetaStepArgumentType().isAssignableFrom(d.clazz)) {
-                return sd;
-            }
-        }
-        return null;
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -26,11 +26,17 @@ package org.jenkinsci.plugins.workflow.cps;
 
 import com.cloudbees.groovy.cps.Continuable;
 import com.cloudbees.groovy.cps.Outcome;
+import com.google.common.base.Function;
 import groovy.lang.Closure;
 import groovy.lang.GString;
 import groovy.lang.GroovyObject;
 import groovy.lang.GroovyObjectSupport;
+import hudson.model.Describable;
+import hudson.model.Descriptor;
 import hudson.model.TaskListener;
+import org.jenkinsci.plugins.structs.describable.DescribableModel;
+import org.jenkinsci.plugins.structs.describable.DescribableParameter;
+import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepAtomNode;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode;
@@ -46,6 +52,7 @@ import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
+import org.jenkinsci.plugins.structs.SymbolLookup;
 
 import java.io.IOException;
 import java.io.PrintStream;
@@ -56,10 +63,16 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.TreeMap;
 
 import static org.jenkinsci.plugins.workflow.cps.ThreadTaskResult.*;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
+
+import org.kohsuke.stapler.ClassDescriptor;
+import org.kohsuke.stapler.NoStaplerConstructorException;
+
+import javax.annotation.Nullable;
 
 /**
  * Scaffolding to experiment with the call into {@link Step}.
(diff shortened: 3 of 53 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -116,7 +118,16 @@
 
         try {
             // TODO: handle arguments to closure
-            Object x = params.body.getBody(currentThread).call();
+            Object x;
+            try {
+                x = params.body.getBody(currentThread).call();
+            } catch (InvokerInvocationException e) {
```
abayer: For my curiosity/learning - what's throwing the exception now that wasn't previously?

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -116,7 +116,8 @@ public Object invokeMethod(String name, Object args) {
         FlowNode an;
 
         // TODO: generalize the notion of Step taking over the FlowNode creation.
```
abayer: FYI, we actually have a JIRA for this TODO now - [JENKINS-38638](https://issues.jenkins-ci.org/browse/JENKINS-36838). Which I probably should assign away from me.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/ExtensionFinderImpl.java
```diff
@@ -0,0 +1,96 @@
+package org.jenkinsci.plugins.workflow.cps.steps.ingroovy;
+
+import com.trilead.ssh2.util.IOUtils;
```
abayer: Is that the `IOUtils` you meant to import?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyCompiler.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.logging.Logger;
+
+/**
+ * Used to compile Groovy step implementation to be able to reflect on its properties.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+@Extension
+@Restricted(NoExternalUse.class) // SezPoz demands that this is public, but this is an implementation detail in this plugin
+public class GroovyCompiler {
```
abayer: I was gonna throw a note on this because I was sure there was already a class in Groovy named this, but no! There isn't! So moving on... =)

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyCompiler.java
```diff
@@ -0,0 +1,106 @@
...
+    /*package*/ StepInGroovyScript parse(StepDescriptorInGroovy d) throws IOException {
+        try {
+            return (StepInGroovyScript)compile(d).newInstance();
+        } catch (ReflectiveOperationException e) {
+            throw new IOException("Failed to instantiate "+d.getClassName(),e);
+        }
+    }
+
+    /**
+     * Obtains the content root folders for CPS classpath
```
abayer: Could we have an explanation of what the content root folders are for?

kohsuke:
This change is ready.
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovy.java
```diff
@@ -0,0 +1,167 @@
...
+            Class<?>[] p = callMethod.getParameterTypes();
+            /*
+                TODO: ClassDescriptor.loadParameterNames requires .class file to be present, which isn't
+                We need AST tranformation to capture CapturedParameterNames
+             */
+//            String[] n = ClassDescriptor.loadParameterNames(callMethod);
+//            assert p.length==n.length;
+//            for (int i=0; i<n.length; i++) {
+//                params.add(new Parameter(p[i],n[i]));
+//            }
```
abayer: Is this still here despite being commented out for a reason?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovy.java
```diff
@@ -0,0 +1,167 @@
...
+            return super.getId()+":"+functionName;
+        }
+
+        /**
+         * If the groovy code takes the closure parameter in the end, that's a sign that we take
+         * body.
+         */
+        @Override
+        public boolean takesImplicitBlockArgument() {
+            return params.isEmpty() && Closure.class.isAssignableFrom(params.get(params.size()-1).type);
```
abayer: :bug: I may be wrong about this, but wouldn't this never actually be true? i.e., if `params` isn't empty, we'd never actually be checking its last element...

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovy.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        @Override
+        public Set<Class<?>> getRequiredContext() {
```
abayer: Would be interesting to see if we can figure out required context from the step itself, but yeah, that's not viable for now.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyExecution.java
```diff
@@ -0,0 +1,81 @@
...
+
+        List<Parameter> pd = descriptor.getParameters();
+        Object[] args = new Object[pd.size()];
+        for (int i=0; i<args.length; i++) {
+            args[i] = arguments.get(pd.get(i).name);
+        }
+
+        // see
+        Script impl = (Script)t.getExecution().getTrustedShell().getClassLoader()
+                .loadClass(descriptor.getClassName()).newInstance();
```
abayer: So this means we'll be reloading the script on every execution, right? Might it make sense to cache the scripts somehow?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyScript.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.jenkinsci.plugins.workflow.cps.CpsScript;
+import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
+
+import java.io.IOException;
+
+import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
+
+/**
+ * Base class that defines contract for steps written in Groovy.
```
abayer: So if we wanted to add more metadata that steps-in-groovy would expose (required context, etc), we'd add more methods here?

src/main/webapp/WEB-INF/steps/META-INF/index
```diff
@@ -0,0 +1,2 @@
+# file that lists FQCN of all the Groovy steps
+helloWorldGroovy
```
abayer: Shouldn't this be in `src/test`?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovy.java
kohsuke: Great catch!

jglick:
Needs merge conflict resolution.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
kohsuke: There is no new exception thrown here. It's just that I discovered `CacheMethod` interferes with the exception thrown from the closure when it's a subtype of `Error`

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyExecution.java
kohsuke: No, `loadClass` will not result in recompilation of the source files. It is already cached by `ClassLoader`.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyScript.java
kohsuke: Yes. Will capture that in Javadoc.

src/main/webapp/WEB-INF/steps/META-INF/index
kohsuke: Oops

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovy.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A {@link Step} written entirely in Groovy.
+ *
+ * Unlike regular extensions, there will be one instance of {@link StepDescriptorInGroovy}
+ * for every step written in Groovy.
+ *
+ * @see 'doc/step-in-groovy.md'
```
jtnord: invalid javadoc comment. causes build failure.

To use a non class reference you must used a " double quote  - e.g. @see "doc/step-in-groovy.md"

jglick:
This is a superset of #33 I guess?

kohsuke:
Yes, it is a super set of #33

kohsuke:
After discussion with Jesse, remaining items:
- [ ] prove that snippet generator works with this
- [ ] test case that covers body closure
- [ ] tests that demonstrate multiple and optional parameters

abayer:
Yeah, getting Snippet Generator support should also mean that introspection in general would work, so +1!

jglick:
> generate `Step` definition class with `@DataBoundConstructor`

No bytecode generation hacks please. We should be able to patch `DescribableModel` to allow you to generate a model directly rather than using reflection.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+From Jenkins Pipeline you can use this like you'd use any other steps:
+
+```
+hello 'Duke'    // this will print 'Hello Duke'
+```
+
+`src/main/webapp/WEB-INF/steps` is the content root for Groovy files
+to be run by the CPS interpreter (aka Pipeline classpath). You can create regular package
+structures in this content root directory. For example, the above
+helloWorld.groovy could be in `src/main/webapp/WEB-INF/steps/org/kohsuke/jenkins/helloWorld.groovy`
```
jglick: Ulch, this is an unpleasant directory structure from the perspective of Maven usage. But apparently we cannot use the normal

    src/main/resources/org/kohsuke/jenkins/helloWorld.groovy

because

> Paths designated here get loaded into trusted CPS classloader, which is exposed to untrusted user CPS classloader. So we need them to be isolated from the rest of the classpath of Jenkins

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+}
+
+void call(String person) {
+    echo "Hello ${person}";
+}
+```
+
+From Jenkins Pipeline you can use this like you'd use any other steps:
+
+```
```
jglick:
```markdown
```groovy
```

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
+# Implementing Step in Groovy
+The standard way of implementing a step is by writing Java code that extends from `Step` and `StepDescriptor`,
+but you can also define a step in Jenkins Pipeline Script.
+
+This approach is quite convenient when your step composes other
+existing steps. For example, the following `hello.groovy` defines
+a step that prints a greeting message to log:
+
+```
```
jglick:
```markdown
```groovy
```

kohsuke:
I wasn't going to generate code via ASM but I was going to create Groovy source file via StringBuilder
... as opposed to continue the current path of one Java class multiple Descriptors, which tends to cause troubles, as in templates. I wanted to see how you weigh both options.

As a case in point Snippetizer as it is today doesn't work with one-class-many-Descriptor. Now I can fix that one, but IMHO it's just the first of many to come.

I can live with either way. Just trying to choose the best option.

doc/step-in-groovy.md
jglick: `helloWorld.groovy`? I thought it was `hello.groovy`?

jglick:
> I was going to create Groovy source file via StringBuilder

Not quite as bad but still sounds like a mess. Would rather avoid tricks. If there is a model which does not correspond to a particular Java class, then fine, represent that explicitly in the `structs` API. Deprecate any API methods which imply that the model is derived from introspecting a `Class`.

> one Java class multiple Descriptors, which tends to cause troubles, as in templates

To the contrary. In `cloudbees-template` you created an elaborate system of generating bytecode so there would be one `Descriptor` per `Class`. It was impossible to understand and full of bugs. I deleted it and the code became much more straightforward.

> Snippetizer as it is today doesn't work with one-class-many-Descriptor

Then we need to fix that.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+hello 'Duke'    // this will print 'Hello Duke'
+```
+
+`src/main/webapp/WEB-INF/steps` is the content root for Groovy files
+to be run by the CPS interpreter (aka Pipeline classpath). You can create regular package
+structures in this content root directory. For example, the above
+helloWorld.groovy could be in `src/main/webapp/WEB-INF/steps/org/kohsuke/jenkins/helloWorld.groovy`
+
+It is also possible to put other supporting classes in this content root directory.
```
jglick: But not resources, I suppose. At worst these could always be loaded from the plugin’s regular `src/main/resources/` by

```java
Jenkins.instance.pluginManager.uberClassLoader.getResource('org/jenkins-ci/plugins/whatever/something.xml')
```

or similar.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+interpreted at runtime together with user's `Jenkinsfile`.
+
+## Details
+To define a step in Groovy, the following requirements have to be met:
+
+* There must be a public default constructor.
+* There must be a `call()` method that takes parameters that the step needs.
+  This can include a `Closure` parameter, but it has to be the last parameter.
+* There must be a `getDisplayName` method that returns a short human readable name
+  of your step. See `Descriptor#getDisplayName()` for more about it.
```
jglick: What about `help.html` and `help-paramName.html`?

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+* There must be `src/main/webapp/WEB-INF/steps/META-INF/index` that lists fully-qualified class names
```
jglick: Why not simply

    src/main/resources/META-INF/pipeline-groovy-steps

?

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+  of all the steps implemented in Groovy. This is how the runtime discovers all the available steps.
+
+## Groovy steps and sandbox
+Groovy code in Pipeline classpath are trusted,
+so it is not subject to the sandbox execution environment nor the script approval process.
+This means code in Pipeline classpath can invoke any Java code.
+
+User-written `Jenkinsfile` can in turn access any code in Pipeline classpath,
+subject to the normal Java access restriction.
```
jglick: Well, Groovy, so it could access even `private` methods, right?

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+
+This means a care must be taken to ensure that your code
+does not expose any methods and properties that can be exploited by malicious `Jenkinsfile`.
```
jglick: Hmm, this seems both dangerous, and too onerous for the step author (who might legitimately want to have some methods for internal use which are not safe to call with arbitrary parameters).

`call` of the step itself is already exposed implicitly as a step function. I suppose there is a use case for allowing the same packaging mechanism to be used to deliver class libraries for non-step-oriented usage. (BTW what about global variables?) But perhaps we should exclude the step package tree from `GroovyClassLoaderWhitelist`, meaning that any members you intend to expose to untrusted code must be marked `@Whitelisted`.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+  of your step. See `Descriptor#getDisplayName()` for more about it.
+* There must be `src/main/webapp/WEB-INF/steps/META-INF/index` that lists fully-qualified class names
+  of all the steps implemented in Groovy. This is how the runtime discovers all the available steps.
+
+## Groovy steps and sandbox
+Groovy code in Pipeline classpath are trusted,
+so it is not subject to the sandbox execution environment nor the script approval process.
+This means code in Pipeline classpath can invoke any Java code.
+
+User-written `Jenkinsfile` can in turn access any code in Pipeline classpath,
```
jglick: This needs test coverage!

doc/step-in-groovy.md
```diff
@@ -0,0 +1,55 @@
...
+* There must be a `call()` method that takes parameters that the step needs.
+  This can include a `Closure` parameter, but it has to be the last parameter.
+* There must be a `getDisplayName` method that returns a short human readable name
+  of your step. See `Descriptor#getDisplayName()` for more about it.
+* There must be `src/main/webapp/WEB-INF/steps/META-INF/index` that lists fully-qualified class names
+  of all the steps implemented in Groovy. This is how the runtime discovers all the available steps.
+
+## Groovy steps and sandbox
+Groovy code in Pipeline classpath are trusted,
+so it is not subject to the sandbox execution environment nor the script approval process.
```
jglick: For any such claims, we need test coverage: either a test demonstrating that the code is able to call non-whitelisted Java methods, or a test demonstrating that attempts by the code to call non-whitelisted Java methods fail with the expected `RejectedAccessException`.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: @abayer when filing an issue corresponding to a `TODO` comment, please either delete the comment, or update it to refer to the issue ID.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -211,7 +216,8 @@ protected Object invokeDescribable(String symbol, Object _args) {
         StepDescriptor metaStep = metaSteps.size()==1 ? metaSteps.get(0) : null;
 
         // The only time a closure is valid is when the resulting Describable is immediately executed via a meta-step
-        NamedArgsAndClosure args = parseArgs(_args, metaStep!=null && metaStep.takesImplicitBlockArgument(), UninstantiatedDescribable.ANONYMOUS_KEY);
+        NamedArgsAndClosure args = parseArgs(_args, metaStep != null && metaStep.takesImplicitBlockArgument(),
+                ANONYMOUS_KEY);
```
jglick: Why? Seems like a gratuitous diff hunk.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -344,8 +350,14 @@ private NamedArgsAndClosure(Map<?,?> namedArgs, Closure body) {
...
     }
 
-    static NamedArgsAndClosure parseArgs(Object arg, StepDescriptor d) {
-        return parseArgs(arg,d.takesImplicitBlockArgument(), loadSoleArgumentKey(d));
+    static NamedArgsAndClosure parseArgs(Object arg, final StepDescriptor d) {
+        return parseArgs(arg, d.takesImplicitBlockArgument(),
+                new Function<Object, Map<String,Object>>() {
+                    @Override
+                    public Map<String, Object> apply(Object arg) {
+                        return d.singleArgument(arg);
```
jglick: Wherefore art thou my lambda?

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: never mind, saw below that this is a new field with the same name

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/ExtensionFinderImpl.java
```diff
@@ -0,0 +1,94 @@
...
+            } catch (MalformedURLException e) {
+                LOGGER.log(WARNING, "Failed to resolve step index from " + root);
+                continue;
+            }
+            InputStream fin;
+
+            try {
+                fin = u.openStream();
+                if (fin==null)      continue;
+            } catch (IOException e) {
```
jglick: Probably better to catch `FileNotFoundException`.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/ExtensionFinderImpl.java
```diff
@@ -0,0 +1,94 @@
...
+                u = new URL(root, "META-INF/index");
+            } catch (MalformedURLException e) {
+                LOGGER.log(WARNING, "Failed to resolve step index from " + root);
+                continue;
+            }
+            InputStream fin;
+
+            try {
+                fin = u.openStream();
+                if (fin==null)      continue;
```
jglick: I do not think it is permitted for `openStream` to return null.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/ExtensionFinderImpl.java
jglick: …which would allow you to move `fin` into the same `try`-with-resources block.

doc/step-in-groovy.md
jglick: At any rate, should mention that this is in the syntax used by `ServiceLoader`.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyCompiler.java
```diff
@@ -0,0 +1,128 @@
...
+            CompilerConfiguration cc = new CompilerConfiguration();
+            cc.setScriptBaseClass(StepInGroovyScript.class.getName());
+
+            // needed to introspect the compiled step
+            cc.addCompilationCustomizers(new ASTTransformationCustomizer(new ParameterNameCaptureTransformation()));
+
+            sh = new GroovyShell(Jenkins.getActiveInstance().getPluginManager().uberClassLoader, new Binding(), cc);
+
+            // register all the content roots
+            addContentRootsTo(sh);
```
jglick: Not going to work if plugins are dynamically installed.

doc/step-in-groovy.md
jglick: And I think making that change would allow us to keep the Groovy class(es) comprising the step in the regular plugin resource tree, so we could load directly from `UberClassLoader`, which in fact already works without any new code. For example, [this whitelisted Java code](https://github.com/jenkinsci/docker-workflow-plugin/blob/dede7791a8fbc489aba6a77a9aa85df563d64ab5/src/main/java/org/jenkinsci/plugins/docker/workflow/ImageNameTokens.java#L45-L51) is accessed from [this untrusted Groovy code](https://github.com/jenkinsci/docker-workflow-plugin/blob/dede7791a8fbc489aba6a77a9aa85df563d64ab5/src/main/resources/org/jenkinsci/plugins/docker/workflow/Docker.groovy#L104-L166). Would make things significantly simpler.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovy.java
```diff
@@ -0,0 +1,160 @@
...
+         * body.
+         */
+        @Override
+        public boolean takesImplicitBlockArgument() {
+            return !params.isEmpty() && Closure.class.isAssignableFrom(params.get(params.size()-1).type);
+        }
+
+        @Override
+        public Set<Class<?>> getRequiredContext() {
+            return Collections.emptySet();
```
jglick: I guess any context has to be acquired by a binary step it calls, or via [JENKINS-28385](https://issues.jenkins-ci.org/browse/JENKINS-28385).

jglick:
Another thing that has no test here: a demonstration that a Groovy step really runs asynchronously, for example with `semaphore`. Would also be wise to demonstrate that it can call both `load` and `parallel`, since these behave specially.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyExecution.java
```diff
@@ -0,0 +1,81 @@
...
+//                .withStartAction(/*... maybe a marker for the future?*/)
+                .withCallback(BodyExecutionCallback.wrap(cps))
+                .start();
+
+        return false;
+    }
+
+    @Override
+    public void stop(@Nonnull Throwable cause) throws Exception {
+        execution.cancel(cause);
```
jglick: And we need a test demonstrating that interrupts are passed through to the innermost binary step.

jglick:
> really runs asynchronously

…including across Jenkins restarts.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyTest.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class StepInGroovyTest {
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    static {
+        try {
+            // fake content root at src/test/resources/stepsInGroovy
+            URL u = StepInGroovyTest.class.getResource("/stepsInGroovy/META-INF/index");
+            GroovyCompiler.additionalContentRoots.add(new URL(u,"../"));
```
jglick: Or you can use `@WithPlugin` for a more realistic test (at the expense of having test sources inside a ZIP).

If my idea about using the plain resource path is feasible, then this would not be necessary.

doc/step-in-groovy.md
kohsuke: You can put resources there too. The whole directory will be on classpath.

doc/step-in-groovy.md
kohsuke: That and `config.jelly` would be necessary if the step wants to support snippetizer.

doc/step-in-groovy.md
kohsuke: I think it's better that the index is on the same content root as the source files.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/ExtensionFinderImpl.java
kohsuke: Better error on the side of caution IMO.

jglick:
> Snippetizer as it is today doesn't work with one-class-many-Descriptor

BTW best not to write new code without merging #38.

jglick:
I am wondering whether we need to do this at all, given the difficulty of
making a quality configuration form fragment by convention. If the main
goal is to be able to implement the `StepExecution` in Pipeline script, it
would suffice to define an API for parsing and running a Groovy resource from your `start` method, right? Passing in some `Serializable` `Step` parameters as closure arguments I suppose, and supplying a path to the script so that `onLoad` can reparse it without needing to save a copy in `build.xml` (with some caveats about upgrading the plugin while a build is running).

Ideally you could reimplement [this](https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/86af7fdf73cdb1bca795c34a9e954a081670619e/src/main/java/org/jenkinsci/plugins/workflow/steps/WaitForConditionStep.java#L47-L137) with a single line of Java plus a resource like

```groovy
def call(body) {
  long delay = 250
  while (true) {
    if (body()) break
    sleep time: delay, unit: 'MILLISECONDS'
    delay = Math.min(delay * 1.2), 15000)
  }
}
```

with little loss of functionality.

Groovy resources should of course be in the trusted loader, so complex logic can stay in `src/main/java/` with unit tests, and the Groovy code can call it directly as a native method without `@Whitelisted`.

I would imagine that, like a `load`ed or global library, the Groovy step would create a block-scoped node in the flow graph, while `Step`s called from the CPS-transformed code would show up as nested nodes; in the case of `takesImplicitBlockArgument`, each body execution could create its own flow graph block, or not (you can always wrap the body execution in `stage(…) {…}` if you want a label for it).

jglick:
Obsolete I think.

PR #35, opened 2016-07-28
[JENKINS-29922] Describe step call formats in “Pipeline Syntax: Reference”

jglick:
Follow-up to #28.

@reviewbybees esp. @kohsuke, @abayer

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
@@ -54,6 +54,78 @@ THE SOFTWARE.
     </p>
     <pre>stage name: 'Build', concurrency: 1</pre>
     <p>
-        There are also some predefined variables, shown at the bottom.
+        Many steps require complex nested configuration for a parameter.
+        (And some nested configuration objects in turn have object-typed parameters.)
+        There are three ways these nested objects may be specified:
     </p>
+    <ol>
+        <li>
+            <p>
+                As maps of parameters. Default values may be omitted.
+                (Note that Groovy map literal syntax resembles list syntax in using square brackets, but keys and values are separated by colons.)
+            </p>
+            <p>
+                The special map key <code>$class</code> is used to represent the simple or (where necessary) fully-qualified class name
+                of the object being requested.
+                <code>$class</code> may be omitted where the containing parameter allows only a single kind of nested object (or list of them):
+            </p>
+            <pre>checkout([$class: 'GitSCM', userRemoteConfigs: [[url: 'git://…'], extensions: [[$class: 'CleanBeforeCheckout']]])</pre>
+            <p>
+                In this example, <code>GitSCM</code> must be specified to distinguish the kind of <code>SCM</code> used by the <code>delegate</code> of <code>checkout</code>
+                (the single mandatory parameter name <code>delegate</code> can be omitted),
+                and <code>CleanBeforeCheckout</code> must be specified to distinguish the kind of <code>GitSCMExtension</code> used by the <code>extensions</code> of <code>GitSCM</code> (a “heterogeneous” list);
+                but <code>$class: 'UserRemoteConfig'</code> may be omitted since the <code>userRemoteConfigs</code> of <code>GitSCM</code> are defined to contain only <code>UserRemoteConfig</code>s (it is a “homogeneous” list).
+            </p>
+            <p>
+                Note that in cases where a single parameter is given, with the name omitted, and that parameter is a map,
+                it must be enclosed in parentheses to avoid a syntactic ambiguity.
+                <!-- TODO see comment in Snippetizer about whether a shorter form is possible -->
+            </p>
+        </li>
+        <li>
+            <p>
+                Jenkins core 2.2+ and some newer plugins define custom <em>symbols</em> for many configuration objects.
+                These are used with a notation looking like a step call, or other function call taking a map of named parameters:
+            </p>
+            <pre>step archiveArtifacts('**.txt')</pre>
+            <p>
+                or to spell out the mandatory parameter names:
+            </p>
+            <pre>step delegate: archiveArtifacts(artifacts: '**.txt')</pre>
+            <p>
+                Some Pipeline steps (like <code>step</code>) delegating to a single object allow the actual step name to be omitted where unambiguous, so you can write simply:
+            </p>
+            <pre>archiveArtifacts '**.txt'</pre>
+            <p>
+                rather than the older form
+            </p>
+            <pre>step([$class: 'ArtifactArchiver', artifacts: '**.txt'])</pre>
+            <p>
+                The nested configuration is an actual value which can be saved into variables etc.:
+            </p>
+            <pre>def archiverConfig(fingerprinting) {
+    fingerprinting ? archiveArtifacts(artifacts: '**.txt', fingerprint: true) : archiveArtifacts('**.txt')
+}
+step archiverConfig(true)</pre>
+            <p>
+                Note that unlike with <code>$class</code> syntax, the symbol name must be specified even for elements of a homogeneous list.
+            </p>
+            <p>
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,78 @@ THE SOFTWARE.
...
-        There are also some predefined variables, shown at the bottom.
+        Many steps require complex nested configuration for a parameter.
+        (And some nested configuration objects in turn have object-typed parameters.)
+        There are three ways these nested objects may be specified:
     </p>
+    <ol>
+        <li>
+            <p>
+                As maps of parameters. Default values may be omitted.
+                (Note that Groovy map literal syntax resembles list syntax in using square brackets, but keys and values are separated by colons.)
```
kohsuke: Syntax example would be useful for the completeness sake

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,78 @@ THE SOFTWARE.
...
+            </p>
+            <p>
+                The special map key <code>$class</code> is used to represent the simple or (where necessary) fully-qualified class name
```
kohsuke: I think we should explain the better syntax first and position this as a fallback.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,78 @@ THE SOFTWARE.
...
+            </p>
+            <p>
+                Note that in cases where a single parameter is given, with the name omitted, and that parameter is a map,
+                it must be enclosed in parentheses to avoid a syntactic ambiguity.
+                <!-- TODO see comment in Snippetizer about whether a shorter form is possible -->
+            </p>
+        </li>
+        <li>
+            <p>
+                Jenkins core 2.2+ and some newer plugins define custom <em>symbols</em> for many configuration objects.
```
kohsuke: Version information is a minor detail that should be at the end, not at the top of the pitch.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,78 @@ THE SOFTWARE.
...
+            <pre>archiveArtifacts '**.txt'</pre>
+            <p>
+                rather than the older form
+            </p>
+            <pre>step([$class: 'ArtifactArchiver', artifacts: '**.txt'])</pre>
+            <p>
+                The nested configuration is an actual value which can be saved into variables etc.:
+            </p>
+            <pre>def archiverConfig(fingerprinting) {
+    fingerprinting ? archiveArtifacts(artifacts: '**.txt', fingerprint: true) : archiveArtifacts('**.txt')
```
kohsuke: This is actually incorrect as `archiveArtifacts` gets immediately executed via the meta step, and there's no way to avoid that.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,78 @@ THE SOFTWARE.
...
+}
+step archiverConfig(true)</pre>
+            <p>
+                Note that unlike with <code>$class</code> syntax, the symbol name must be specified even for elements of a homogeneous list.
+            </p>
+            <p>
+                The <b>Snippet Generator</b> tool will always produce this newer syntax when possible,
```
kohsuke: Perhaps I misunderstand the context in which this help is used, but I'd think this message should be at close to the top, before we start going into encyclopedic details of this syntax.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
jglick: But the example _does_ show both map and list literals.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
jglick: That was my first thought, but it did not work out well, since the newer syntax is of more limited applicability, and it is necessary to refer to concepts which are better explained with the earlier syntax.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
jglick: Mm OK, so I need a different example.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
jglick: Which message? That _Snippet Generator_ prefers this syntax? Not sure how you would phrase that without actually explaining what the syntax is first.

abayer: Yeah, I'd prefer the deep dark details like that be lower in the doc. I'm good with this.

abayer:
:bee:

abayer:
re-:bee:

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,77 @@ THE SOFTWARE.
...
+        (And some nested configuration objects in turn have object-typed parameters.)
+        There are three ways these nested objects may be specified, in decreasing order of preference:
+    </p>
+    <ol>
+        <li>
+            <p>
+                Some configuration objects define custom <em>symbols</em>.
+                These are used with a notation looking like a step call, or other function call taking a map of named parameters:
+            </p>
+            <pre>step archiveArtifacts('**.txt')</pre>
```
kohsuke: This doesn't actually work like you wrote. at the end of `archiveArtifacts` evaluation, it always gets executed with a meta-step. The evaluation of that method doesn't know that there's a surrounding `step` function.

Is this indicative of a problem in the current code? I understand why people would expect the behaviour you wrote.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
```diff
@@ -54,6 +54,77 @@ THE SOFTWARE.
...
+                rather than the older form described below:
+            </p>
+            <pre>step([$class: 'ArtifactArchiver', artifacts: '**.txt'])</pre>
+            <p>
+                The nested configuration is an actual value which can be saved into variables etc.:
+            </p>
+            <pre>def parallelism(deterministic) {
+    deterministic ? count(3) : time(15)
+}
+splitTests parallelism(true)</pre>
```
kohsuke: :+1:

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/help.jelly
kohsuke: Here is one way to fix this. Normal execution environment doesn't allow `archiveArtifacts` call being invoked here to tell apart whether there's a function call around it or not (`step` in this case), but since we run this in our own interpreter that is groovy-cps, we can actually tell it apart.

By using the call site tagging mechanism in https://github.com/cloudbees/groovy-cps/pull/36, we can mark those call sites that are at the statement level (aka top-level call), and change the behaviour of meta-step handling accordingly.

jglick: Too magical. I will just change the documentation.

kohsuke:
:bee:

abayer:
re-re-:bee: =)

PR #36, opened 2016-08-04
Include the running node count in Pipeline thread dumps in support bundles

jglick:
Useful to be able to see if very large graphs are being encountered, which can have performance implications.

@reviewbybees esp. @svanoort & @fbelzunc

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -117,6 +117,7 @@ import org.jboss.marshalling.reflect.SerializableClassRegistry;
 
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionList;
+import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 
 /**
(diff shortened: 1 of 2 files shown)


abayer:
:bee: assuming tests pass.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -408,6 +408,10 @@ public int iota() {
         return iota.incrementAndGet();
     }
 
+    int currentNodeCount() {
+        return iota.get();
```
svanoort: :bug:  Gotcha here: the iota != the actual node count.  FlowStartNode is iota 2, parallel blocks skip an iota value too.  

If you call it "approximateNodeCount" with a javadoc that the real count may be slightly less I'll be happy.

svanoort:
I love that this is included, make no mistake.  Can't remember if the thread dump goes into the support bundle, but would love it to be. 

This would be highly useful to some of the scalability stuff I tend to do.

jglick:
> Can't remember if the thread dump goes into the support bundle

By default, yes.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -408,7 +409,13 @@ public int iota() {
         return iota.incrementAndGet();
     }
 
-    int currentNodeCount() {
+    /**
+     * Returns an approximate size of the flow graph, based on the heuristic that the iota is incremented once per new node.
+     * The exact count may be a little different due to special cases.
+     * ({@link FlowNodeStorage} does not currently offer a size, or a set of all nodes.
+     * An exact count could be obtained with {@link FlowGraphWalker}, but this could be more overhead.)
```
svanoort: Much, much more overhead, since you're creating a HashSet of all existent flownodes. ;)

Multiple seconds even is possible.

svanoort:
:bee: with the addendum.   🚢  it!

PR #38, opened 2016-08-05
[JENKINS-29922] Display single-argument metasteps inline in Snippetizer

jglick:
[JENKINS-29922](https://issues.jenkins-ci.org/browse/JENKINS-29922)

Means that plugins which offer `SimpleBuildStep`s or `SimpleBuildWrapper`s with `@Symbol`s will have those extensions be displayed like native steps in _Snippet Generator_:

snippetizer

Also solves one component of [JENKINS-37215](https://issues.jenkins-ci.org/browse/JENKINS-37215), mainly because I had to do _something_ just to make the existing content generators compile again: the static HTML reference (but not GDSL/DSLD) now shows such extensions like top-level steps:

html

@reviewbybees esp. @kohsuke

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
@@ -399,6 +399,14 @@ import org.kohsuke.stapler.StaplerRequest;
             return getSymbol().compareTo(o.getSymbol());
         }
 
+        @Override public boolean equals(Object obj) {
+            return obj instanceof QuasiDescriptor && real == ((QuasiDescriptor) obj).real;
+        }
+
+        @Override public int hashCode() {
+            return real.hashCode();
+        }
+
         @Override public String toString() {
             return getSymbol() + "=" + real.clazz.getSimpleName();
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
@@ -27,15 +27,14 @@ package org.jenkinsci.plugins.workflow.cps;
 import hudson.Extension;
 import hudson.Functions;
 import hudson.model.Action;
+import hudson.model.Describable;
 import hudson.model.Descriptor;
 import hudson.model.DescriptorByNameOwner;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.RootAction;
-import java.io.Serializable;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -48,8 +47,10 @@ import javax.lang.model.SourceVersion;
 import jenkins.model.Jenkins;
 import jenkins.model.TransientActionFactory;
 import net.sf.json.JSONObject;
+import org.jenkinsci.Symbol;
 import org.jenkinsci.plugins.structs.describable.DescribableModel;
 import org.jenkinsci.plugins.structs.describable.DescribableParameter;
+import org.jenkinsci.plugins.structs.describable.HeterogeneousObjectType;
 import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
 import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
@@ -69,8 +70,8 @@ import org.kohsuke.stapler.StaplerRequest;
     /**
      * Short-hand for the top-level invocation.
      */
-    static String object2Groovy(Object o) throws UnsupportedOperationException {
-        return object2Groovy(new StringBuilder(),o, false).toString();
+    static String step2Groovy(Step s) throws UnsupportedOperationException {
+        return object2Groovy(new StringBuilder(), s, false).toString();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -349,17 +350,69 @@ private static boolean isSingleMap(Map<String, ?> args) {
...
+                            }
+                        }
+                    } // TODO currently not handling metasteps with other parameters, either required or (like GenericSCMStep) not
+                }
             }
         }
         return t;
     }
 
+    /** Similar to {@link StepDescriptor} but could also represent a metastep’s delegate. */
```
kohsuke: IIUC, I suggest this javadoc to be expanded to something like this:

---

{@code QuasiDescriptor} represents a step and other step-like things that show up in Snippetizer, which can generate some Pipeline Script fragment.

{@link #real} can be a {@code StepDescriptor}, in which case Snippetizer generates an invocation of that step. Or it can be a {@link Descriptor} that can be executed via {@linkplain StepDescriptor#isMetaStep() meta step}, because from users' point of view a regular Describable (say {@code SimpleBuildStep}) executed via a meta step is indistinguishable from a {@code Step}.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -349,17 +350,69 @@ private static boolean isSingleMap(Map<String, ?> args) {
...
+                                    Class<?> delegateOptionType = delegateOptionSchema.getType();
+                                    Descriptor<?> delegateDescriptor = Jenkins.getActiveInstance().getDescriptorOrDie(delegateOptionType.asSubclass(Describable.class));
+                                    Symbol symbol = delegateDescriptor.getClass().getAnnotation(Symbol.class);
+                                    if (symbol != null && symbol.value().length > 0) {
+                                        t.add(new QuasiDescriptor(delegateDescriptor));
+                                    }
+                                }
+                            }
+                        }
+                    } // TODO currently not handling metasteps with other parameters, either required or (like GenericSCMStep) not
```
kohsuke: Why? I understand that there's no way to offer configuration for those extra parameters without further change, but isn't it still useful to list them in Snippetizer?

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -349,17 +350,69 @@ private static boolean isSingleMap(Map<String, ?> args) {
...
         for (StepDescriptor d : StepDescriptor.all()) {
             if (d.isAdvanced() == advanced) {
-                t.add(d);
+                t.add(new QuasiDescriptor(d));
+                if (d.isMetaStep()) {
+                    DescribableModel<?> m = new DescribableModel<>(d.clazz);
+                    Collection<DescribableParameter> parameters = m.getParameters();
+                    if (parameters.size() == 1) {
+                        DescribableParameter delegate = parameters.iterator().next();
+                        if (delegate.isRequired()) {
```
kohsuke: I think `DescribableModel.getSoleRequiredParameter()` captures the intent and the constraints better. A meta step, it should have one and the only one required parameter.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
kohsuke: And this scheme will be expanded further with `GlobalVariable` I assume.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dsld.groovy
```diff
@@ -17,16 +19,20 @@ Map<StepDescriptor,DescribableModel> steps = new LinkedHashMap<StepDescriptor, D
...
-    snippetizer.getStepDescriptors(isAdvanced).each { d ->
+    snippetizer.getQuasiDescriptors(isAdvanced).each { d ->
+        if (!Step.isAssignableFrom(d.real.clazz)) {
+            return // TODO JENKINS-37215
+        }
+        StepDescriptor step = Jenkins.instance.getDescriptor(d.real.clazz)
         DescribableModel model
         try {
-            model = new DescribableModel(d.clazz)
+            model = new DescribableModel(d.real.clazz)
```
kohsuke: Makes one class one Descriptor assumption that will be lifted soon.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -349,17 +350,69 @@ private static boolean isSingleMap(Map<String, ?> args) {
...
+                            if (delegate.getType() instanceof HeterogeneousObjectType) {
```
kohsuke: If we look through `ExtensionList.lookup(Descriptor.class)` and find all of them that satisfies `delegate.getErasedType().isAssignableFrom(Descriptor.clazz)` then, single case can cover all `ParameterType`s, heterogeneous or not.

kohsuke:
I'm done reviewing this. :bee: to unblock since I'll be away next week

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: No `getSoleRequiredParameter` is inappropriate here, because it would return non-null for `checkout` if made a metastep, which is not a case we are able to handle yet. See the follow-up issue.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: Perhaps, though a metastep delegate parameter really should be a `HeterogeneousObjectType`, we are just neglecting to enforce this currently.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: See [JENKINS-37227](https://issues.jenkins-ci.org/browse/JENKINS-37227).

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: > And this scheme will be expanded further with `GlobalVariable` I assume.

I do not think so; why do you say that?

> I suggest this javadoc to be expanded

Yes, will do.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dsld.groovy
jglick: Yes, see note in #34, the replacement remains TBD.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -349,17 +352,77 @@ private static boolean isSingleMap(Map<String, ?> args) {
...
+    public static final class QuasiDescriptor implements Comparable<QuasiDescriptor> {
+
+        public final Descriptor<?> real;
+
+        QuasiDescriptor(Descriptor<?> real) {
+            this.real = real;
+        }
+
+        public String getSymbol() {
+            return real instanceof StepDescriptor ? ((StepDescriptor) real).getFunctionName() : real.getClass().getAnnotation(Symbol.class).value()[0];
```
oleg-nenashev: 🐜 Missing null check in <code> getAnnotation</code>

jglick: Because we already did the null check on line 373.

oleg-nenashev:
🐝

PR #39, opened 2016-08-08
[JENKINS-34064] Pick up newer groovy-cps that reverts the JENKINS-26481 fix which only works in Jenkins 1

jglick:
Downstream of cloudbees/groovy-cps#38.

@reviewbybees esp. @kohsuke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,7 +171,7 @@
         <dependency>
             <groupId>com.cloudbees</groupId>
             <artifactId>groovy-cps</artifactId>
-            <version>1.9-SNAPSHOT</version>
+            <version>1.9</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.ui</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.18</version>
+            <version>1.21</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -89,7 +89,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.18</version>
+            <version>1.21</version>
```
jglick: Picks up https://github.com/jenkinsci/script-security-plugin/pull/63.

pom.xml
```diff
@@ -171,7 +171,7 @@
         <dependency>
             <groupId>com.cloudbees</groupId>
             <artifactId>groovy-cps</artifactId>
-            <version>1.8-SNAPSHOT</version>
+            <version>1.9-SNAPSHOT</version>
```
jglick: Downstream of https://github.com/cloudbees/groovy-cps/pull/38.

src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
```diff
@@ -290,6 +290,7 @@ public void stop(Throwable cause) throws Exception {
         story.addStep(new Statement() {
             @Override public void evaluate() throws Throwable {
                 ScriptApproval.get().approveSignature("staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods plus java.util.Collection java.lang.Object");
+                ScriptApproval.get().approveSignature("staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods plus java.util.List java.lang.Object"); // Groovy 2
```
jglick: Miscellaneous Groovy 2 overload not yet handled in `generic-whitelist-groovy2`.

jglick:
Fails due to snapshot issues; last I checked this CI job will not pick up snapshots of the dependency, due to repository configuration. Anyway the interesting thing here still needs to be done manually, which is to rerun all tests with `-Djenkins.version=2.7.2`.

oleg-nenashev:
LGTM 🐝

PR #40, opened 2016-08-10
Print a message whenever we are ready to resume running the program

jglick:
Complements https://github.com/jenkinsci/workflow-job-plugin/pull/12. Makes problems like the test failure described in https://github.com/jenkinsci/pipeline-input-step-plugin/pull/2 more apparent. Also complements [JENKINS-26130](https://issues.jenkins-ci.org/browse/JENKINS-26130): in addition to messages explaining that resumption is blocked and why, print a message when it is unblocked.

@reviewbybees


svanoort:
🐝

PR #29, opened 2016-06-30
[JENKINS-36372] Root bindings not accessible to scripts loaded after restart

jglick:
[JENKINS-36372](https://issues.jenkins-ci.org/browse/JENKINS-36372)

@reviewbybees esp. @kohsuke

PR #41, opened 2016-08-12
Memory leak introduced with the fix of JENKINS-36372

jglick:
Had disabled a test due to still-undiagnosed failures on Jenkins 2, which concealed the fact that a recent bug fix (#29) regressed the leak on Jenkins 1:

```
{groovy.lang.GroovyClassLoader$InnerLoader@571df0f9=private static final org.jboss.marshalling.reflect.SerializableClassRegistry org.jboss.marshalling.reflect.SerializableClassRegistry.INSTANCE->
org.jboss.marshalling.reflect.SerializableClassRegistry@67289b2f-registry->
org.jboss.marshalling.reflect.UnlockedHashMap@46269dbf-table->
org.jboss.marshalling.reflect.UnlockedHashMap$Table@4a13d1ef-array->
[Ljava.lang.Object;@2b2b3f3-[121]->
[Lorg.jboss.marshalling.reflect.UnlockedHashMap$Item;@182e8565-[0]->
org.jboss.marshalling.reflect.UnlockedHashMap$Item@13d66b69-key->
groovy.lang.GroovyClassLoader$InnerLoader@571df0f9}
```

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <jenkins-test-harness.version>2.15-SNAPSHOT</jenkins-test-harness.version>
         <npm.loglevel>--silent</npm.loglevel>
     </properties>
     <dependencies>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -901,9 +901,11 @@ public class CpsFlowExecution extends FlowExecution {
     synchronized void cleanUpHeap() {
         shell = null;
         trusted = null;
-        SerializableClassRegistry.getInstance().release(scriptClass.getClassLoader());
-        Introspector.flushFromCaches(scriptClass); // does not handle other derived script classes, but this is only SoftReference anyway
-        scriptClass = null;
+        if (scriptClass != null) {
+            SerializableClassRegistry.getInstance().release(scriptClass.getClassLoader());
+            Introspector.flushFromCaches(scriptClass); // does not handle other derived script classes, but this is only SoftReference anyway
+            scriptClass = null;
+        }
     }
 
     synchronized FlowHead getFirstHead() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -587,6 +587,7 @@ public class CpsFlowExecution extends FlowExecution {
             @Override public void onFailure(Throwable t) {
                 LOGGER.log(Level.WARNING, "Failed to set program failure on " + owner, t);
                 onProgramEnd(new Outcome(null, t));
+                cleanUpHeap();
             }
         });
     }
@@ -895,8 +896,9 @@ public class CpsFlowExecution extends FlowExecution {
         first.setNewHead(head);
         heads.clear();
         heads.put(first.getId(),first);
+    }
 
-        // clean up heap
+    synchronized void cleanUpHeap() {
         shell = null;
         trusted = null;
         SerializableClassRegistry.getInstance().release(scriptClass.getClassLoader());
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -898,7 +898,7 @@ public class CpsFlowExecution extends FlowExecution {
         heads.put(first.getId(),first);
     }
 
-    synchronized void cleanUpHeap() {
+    void cleanUpHeap() {
         shell = null;
         trusted = null;
         if (scriptClass != null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -104,10 +104,13 @@ import hudson.model.User;
 import hudson.security.ACL;
 import hudson.security.AccessControlled;
 import java.beans.Introspector;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.annotation.concurrent.GuardedBy;
 import jenkins.security.NotReallyRoleSensitiveCallable;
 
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -63,6 +63,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <jenkins-test-harness.version>2.15-SNAPSHOT</jenkins-test-harness.version>
```
jglick: Downstream of https://github.com/jenkinsci/jenkins-test-harness/pull/32. Actually passes without this dependency, but takes three times as long to collect the reference, due to lack of exponential escalation in the original version.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -391,6 +391,10 @@ private void run() throws IOException {
         if (doneSomeWork) {
             saveProgram();
         }
+        if (threads.isEmpty()) {
```
jglick: cf. lines 379–380 above

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -587,6 +587,7 @@ private void loadProgramFailed(final Throwable problem, SettableFuture<CpsThread
             @Override public void onFailure(Throwable t) {
                 LOGGER.log(Level.WARNING, "Failed to set program failure on " + owner, t);
                 onProgramEnd(new Outcome(null, t));
+                cleanUpHeap();
```
jglick: there is no `CpsThreadGroup` here to clear

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -895,8 +896,9 @@ public boolean isComplete() {
         first.setNewHead(head);
         heads.clear();
         heads.put(first.getId(),first);
+    }
 
-        // clean up heap
+    synchronized void cleanUpHeap() {
```
jglick: `register` call in original bug fix added some objects to a record in this class. `SerializableClassRegistry.release` was getting called on the right loader, but too soon—_before_ the final call to `saveProgram`, which would up adding the loader back to the registry.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -391,6 +391,10 @@ private void run() throws IOException {
...
+            execution.cleanUpHeap();
+            scripts.clear();
```
jglick: Clearing `scripts` was also necessary, or even with the fix to the timing of `SerializableClassRegistry`, the test would still fail on Jenkins 1. The failure showed no root references, so I had to guess at the cause.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -115,42 +109,6 @@ public static void register(Object o) {
         });
     }
 
-    /* Failed attempt to make the test print soft references it has trouble clearing. The test ultimately passes, but cannot find the soft references via any root path.
```
jglick: The useful pieces incorporated into the upstream PR.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -92,18 +87,17 @@
     public static void register(Object o) {
         LOADER = new WeakReference<>(o.getClass().getClassLoader());
     }
-    @Ignore("TODO fails in Jenkins 2 for reasons TBD (no root references detected)")
     @Test public void loaderReleased() {
         story.addStep(new Statement() {
             @Override public void evaluate() throws Throwable {
+                Assume.assumeThat("TODO fails in Jenkins 2 for reasons TBD (no root references detected)", Jenkins.VERSION, Matchers.startsWith("1."));
```
jglick: Additional calls to `fromRoots` in the upstream PR were an attempt to diagnose this, so far without success.

jglick:
Seem to be test failures, trying to figure out what they mean.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -391,6 +393,10 @@ private void run() throws IOException {
         if (doneSomeWork) {
             saveProgram();
         }
+        if (ending) {
+            execution.cleanUpHeap();
+            scripts.clear();
```
jglick: Clearing `scripts` (introduced in #29) was also necessary, or even with the fix to the timing of `SerializableClassRegistry.release`, the test would still fail on Jenkins 1. The failure showed no root references, so I had to guess at the cause.

jglick:
FindBugs failure, will deal with it.

In the meantime I found the cause of the Jenkins 2 failure at last:

```
groovy.lang.GroovyClassLoader$InnerLoader [Stack Local]
<loader> of WorkflowScript
klazz of org.codehaus.groovy.reflection.ClassInfo
value of org.codehaus.groovy.reflection.GroovyClassValuePreJava7$EntryWithValue
[0] of java.lang.Object[3]
[12] of java.lang.Object[32]
table of org.codehaus.groovy.reflection.GroovyClassValuePreJava7$GroovyClassValuePreJava7Segment
[14] of org.codehaus.groovy.util.AbstractConcurrentMapBase$Segment[16]
segments of org.codehaus.groovy.reflection.GroovyClassValuePreJava7$GroovyClassValuePreJava7Map
map of org.codehaus.groovy.reflection.GroovyClassValuePreJava7
globalClassValue of org.codehaus.groovy.reflection.ClassInfo
[6128] of java.lang.Object[20480]
elementData of java.util.Vector size = 13,066
classes of sun.misc.Launcher$AppClassLoader [Stack Local]
contextClassLoader of java.util.TimerThread [Thread, Stack Local]  "Timer-2" daemon tid=78 [TIMED_WAITING]
```

Unless you are running on IBM J9 VM, Groovy 2 switches to a broken `GroovyClassValuePreJava7` which leaks `ClassLoader`s. (They claim to be affected by JDK-8136353 but I think the mistake is in Groovy, not Java.) The switch is controlled by a system property, but you would have to set it before Jenkins starts, so Pipeline cannot do so. So I am looking for a way to either clear all values from `ClassInfo.globalClassValue` (unfortunately Groovy’s nonstandard map implementations lack a `clear` method), or reset the whole collection to a version that works better (which is scary since it will affect every use of Groovy in Jenkins, pending JENKINS-29068), or somehow look up individual `Class` entries and clear them.

jglick:
With a patched INSANE, I can now diagnose the problem properly in the test:

```
{groovy.lang.GroovyClassLoader$InnerLoader@368430db=private static final org.codehaus.groovy.reflection.GroovyClassValue org.codehaus.groovy.reflection.ClassInfo.globalClassValue->
org.codehaus.groovy.reflection.GroovyClassValuePreJava7@1a70d5df-map->
org.codehaus.groovy.reflection.GroovyClassValuePreJava7$GroovyClassValuePreJava7Map@54d79178-segments->
[Lorg.codehaus.groovy.util.AbstractConcurrentMapBase$Segment;@4a3f3c19-[12]->
org.codehaus.groovy.reflection.GroovyClassValuePreJava7$GroovyClassValuePreJava7Segment@3aa898c1-table->
[Ljava.lang.Object;@4bb9c2ea-[11]->
org.codehaus.groovy.reflection.GroovyClassValuePreJava7$EntryWithValue@274aa2f8-value->
org.codehaus.groovy.reflection.ClassInfo@3eacd88f-klazz->
java.lang.Class@4f65c3ca-<changed>->
groovy.lang.GroovyClassLoader$InnerLoader@368430db}
```

jglick:
Fixed that problem what I can tell, but now the test fails again on Jenkins 2 with no apparent GC root path. Heap analysis tools just tell me that the `InnerLoader` is being held as a “JNI global”, which does not say much. So I guess I leave it as is and hope this leak does not manifest itself in production.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -895,13 +899,63 @@ public boolean isComplete() {
         first.setNewHead(head);
         heads.clear();
         heads.put(first.getId(),first);
+    }
 
-        // clean up heap
+    void cleanUpHeap() {
```
jglick: `SerializableClassRegistry.release` was getting called on the right loader, but too soon—_before_ the final call to `saveProgram`, which would up adding the loader back to the registry. Not sure when this regressed—presumably at some point after the test was disabled.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -895,13 +899,63 @@ public boolean isComplete() {
...
+            try {
+                cleanUpGlobalClassValue(loader);
+            } catch (Exception x) {
+                LOGGER.log(Level.WARNING, "failed to clean up memory from " + owner, x);
+            }
+            scriptClass = null;
+        }
+    }
+
+    private void cleanUpGlobalClassValue(@Nonnull ClassLoader loader) throws Exception {
```
jglick: This is to deal with Jenkins 2 issues.

abayer:
:bee:

PR #42, opened 2016-08-12
Set the Thread.contextClassLoader to the GroovyClassLoader

jglick:
Intended to augment https://github.com/jenkinsci/script-security-plugin/pull/80 a bit: some class loading calls from deep in the bowels of Groovy use the CCL, and so we can get a performance benefit by making sure that is set to the `GroovyClassLoader`, since that will delegate to the `SandboxResolvingClassLoader` which will then cache the expensive core results.

Only partially effective, since plenty of other calls either using `Class.forName(String)`—meaning, look it up in the Jenkins core class loader, typically provided by the application server—or use the `ClassLoader` associated with some reference class, which might be in the JRE, Jenkins core (incl. Groovy itself), a plugin, or the Groovy script. Any such calls which use an initiating loader below `GroovyClassLoader`¹ will still go through the full stack and will introduce lock contention and I/O overhead.

At any rate, it seems generally appropriate for the CCL used during program execution to be that of the program.

@reviewbybees

¹`SandboxResolvingClassLoader` is never an initiating loader, since it defines no classes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
-        <jenkins-test-harness.version>2.15-SNAPSHOT</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2.15</jenkins-test-harness.version>
         <npm.loglevel>--silent</npm.loglevel>
     </properties>
     <dependencies>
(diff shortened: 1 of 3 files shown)


jglick:
Would perhaps be useful merged with #41.

abayer:
:bee: lgtm

svanoort:
🐝  AFAICT

PR #43, opened 2016-08-15
JENKINS-37403 - downstream change to use SymbolLookup.getSymbolValue(…)

abayer:
[JENKINS-37403](https://issues.jenkins-ci.org/browse/JENKINS-37403)

Downstream of https://github.com/jenkinsci/structs-plugin/pull/10

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -132,10 +132,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         try {
             for (Class<?> e : Index.list(Symbol.class, Jenkins.getActiveInstance().pluginManager.uberClassLoader, Class.class)) {
                 if (Descriptor.class.isAssignableFrom(e)) {
-                    Set<String> s = SymbolLookup.getSymbolValue(e);
-                    if (!s.isEmpty()) {
-                        symbols.addAll(s);
-                    }
+                    symbols.addAll(SymbolLookup.getSymbolValue(e));
                 }
             }
         } catch (IOException x) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -132,9 +132,9 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         try {
             for (Class<?> e : Index.list(Symbol.class, Jenkins.getActiveInstance().pluginManager.uberClassLoader, Class.class)) {
                 if (Descriptor.class.isAssignableFrom(e)) {
-                    Symbol s = e.getAnnotation(Symbol.class);
+                    String s = SymbolLookup.getSymbolValue(e);
                     if (s != null) {
-                        symbols.addAll(Arrays.asList(s.value()));
+                        symbols.addAll(Arrays.asList(s));
                     }
                 }
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -132,9 +132,9 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         try {
             for (Class<?> e : Index.list(Symbol.class, Jenkins.getActiveInstance().pluginManager.uberClassLoader, Class.class)) {
                 if (Descriptor.class.isAssignableFrom(e)) {
-                    String s = SymbolLookup.getSymbolValue(e);
-                    if (s != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -132,9 +132,9 @@ public Object invokeMethod(String name, Object args) {
         try {
             for (Class<?> e : Index.list(Symbol.class, Jenkins.getActiveInstance().pluginManager.uberClassLoader, Class.class)) {
                 if (Descriptor.class.isAssignableFrom(e)) {
-                    Symbol s = e.getAnnotation(Symbol.class);
+                    String s = SymbolLookup.getSymbolValue(e);
                     if (s != null) {
-                        symbols.addAll(Arrays.asList(s.value()));
+                        symbols.addAll(Arrays.asList(s));
```
jglick: :bug: this removes support for multiple symbols.

jglick: (Which is something you need to think about in the other downstream PRs. What is your behavior when there are multiple allowed symbols? Probably rather than `@CheckForNull String` you need to return `@Nonnull Set<String>`.)

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/StateMetaStep.java
```diff
@@ -45,7 +46,7 @@ public void setModerate(boolean m) {
 
         @Override protected Void run() throws Exception {
             if (step.moderate) {
-                listener.getLogger().println("Introducing "+step.state.getDescriptor().getClass().getAnnotation(Symbol.class).value()[0]);
+                listener.getLogger().println("Introducing "+ SymbolLookup.getSymbolValue(step.state.getDescriptor()));
```
jglick: I suppose `SymbolLookup.getSymbolValue(step.state)` would suffice now.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: Ah, good point. Never thought about that! Reworking to come.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/StateMetaStep.java
abayer: Yup, timing, man. =)

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: Actually, hmm...might we actually want both? Not sure...

abayer: Nah, I take that back. I'm overthinking. =)

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -132,9 +132,9 @@ public Object invokeMethod(String name, Object args) {
         try {
             for (Class<?> e : Index.list(Symbol.class, Jenkins.getActiveInstance().pluginManager.uberClassLoader, Class.class)) {
                 if (Descriptor.class.isAssignableFrom(e)) {
-                    Symbol s = e.getAnnotation(Symbol.class);
-                    if (s != null) {
-                        symbols.addAll(Arrays.asList(s.value()));
+                    Set<String> s = SymbolLookup.getSymbolValue(e);
+                    if (!s.isEmpty()) {
```
jglick: Not really necessary; could just write

```java
symbols.addAll(SymbolLookup.getSymbolValue(e));
```

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -402,7 +404,16 @@ private static boolean isSingleMap(Map<String, ?> args) {
...
 
         public String getSymbol() {
-            return real instanceof StepDescriptor ? ((StepDescriptor) real).getFunctionName() : real.getClass().getAnnotation(Symbol.class).value()[0];
+            if (real instanceof StepDescriptor) {
+                return ((StepDescriptor) real).getFunctionName();
+            } else {
+                Set<String> symbolValues = SymbolLookup.getSymbolValue(real);
+                if (!symbolValues.isEmpty()) {
+                    return symbolValues.iterator().next();
+                } else {
```
jglick: Should not happen, since we have already checked for this above; could just as well throw an `AssertionError`.

jglick:
:bee:

jglick:
re:bee:

abayer:
@reviewbybees done

PR #44, opened 2016-08-16
 [JENKINS-37400] Allow to use constant variable names for parameters.

ikedam:
[JENKINS-37400](https://issues.jenkins-ci.org/browse/JENKINS-37400)

I want to use constant values in workflow-cps like
```
def runWrapper = selectRun job: 'upstream-project-name', 
  selector: status(SUCCESSFUL)
```

This change requires jenkinsci/structs-plugin#11. Build will fail.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.3</version>
+            <version>1.4-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsWhitelist.java
@@ -38,14 +38,13 @@ class CpsWhitelist extends AbstractWhitelist {
                 return true;
             }
             if (name.equals("getProperty") && args.length == 1 && args[0] instanceof String) {
-                for (GlobalVariable v : GlobalVariable.ALL) {
-                    if (v.getName().equals(args[0])) {
-                        return true;
-                    }
-                }
+                return true;
             }
         }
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.3</version>
+            <version>1.4-SNAPSHOT</version>
```
ikedam: As this requires jenkinsci/structs-plugin#11.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsWhitelist.java
```diff
@@ -43,11 +44,23 @@ public boolean permitsMethod(Method method, Object receiver, Object[] args) {
...
                 }
+                // Safe if it's considered a constant variable name.
+                String property = (String)args[0];
+                try {
+                    if (((CpsScript)receiver).getProperty(property) instanceof UninstantiatedConstant) {
+                        return true;
+                    }
+                } catch (Exception e) {
+                    // ignore any exceptions
+                }
```
ikedam: Though I don't think this is a good way, I have no good idea.

jglick:
> Build will fail.

`mvn -DskipTests deploy` dependencies.

ikedam:
> `mvn -DskipTests deploy` dependencies.

That means deploying a snapshot release with jenkinsci/structs-plugin#11 to maven.jenkins-ci.org, doesn't it?
I've run a build and a test in my environment by running `mvn install`.
I don't think it's a good idea to deploy a plugin with local changes even if it's a snapshot release, and I'll do that at least after jenkinsci/structs-plugin#11 is merged.

jglick:
> That means deploying a snapshot release with jenkinsci/structs-plugin#11 to maven.jenkins-ci.org, doesn't it?

Yes, we do that all the time for chained PRs.

ikedam:
Abondon the review.
I'll try another way.

PR #46, opened 2016-08-20
[JENKINS-25623] time out & kill infinite recursion

kohsuke:
[JENKINS-25623](https://issues.jenkins-ci.org/browse/JENKINS-25623)


kohsuke:
Depends on https://github.com/cloudbees/groovy-cps/pull/39
@reviewbybees

jglick:
Does this extend pause (#32) to non-`Step` code? If so, there is no test coverage for it.

src/main/java/org/jenkinsci/plugins/workflow/cps/Safepoint.java
```diff
@@ -0,0 +1,30 @@
...
+import com.cloudbees.groovy.cps.Continuable;
+import com.cloudbees.groovy.cps.Outcome;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
+
+/**
+ * Safepoint called from the CPS interpreter.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+// classes compiled from Pipeline Script refers to this class, so it needs to be public, but
```
jglick: …but what?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -151,62 +152,59 @@ public StepExecution getStep() {
     @Nonnull Outcome runNextChunk() throws IOException {
         assert program!=null;
 
-        while (true) {
```
jglick: Reviewers need to look [here](https://github.com/jenkinsci/workflow-cps-plugin/pull/46/files?w=1#diff-1a70e438dd28d7a6307fc2bfb55904e8L154). Basically almost all of this was just reindentation.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -221,96 +219,52 @@ public void unexport(BodyReference ref) {
      *      {@link Future} object that represents when the CPS VM is executed.
      */
     public Future<?> scheduleRun() {
-        final Future<Future<?>> f;
+        final SettableFuture<Void> f = SettableFuture.create();
         try {
-        f = runner.submit(new Callable<Future<?>>() {
```
jglick: [reindentation](https://github.com/jenkinsci/workflow-cps-plugin/pull/46/files?w=1#diff-6452cd1bea46404d75cc1ba4559cc3adL226)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -347,56 +301,66 @@ public boolean isPaused() {
...
+     *      is akin to a Unix thread waiting for I/O completion.
      */
     @CpsVmThreadOnly("root")
-    private void run() throws IOException {
-        boolean doneSomeWork = false;
-        boolean changed;    // used to see if we need to loop over
+    private boolean run() throws IOException {
+        boolean changed = false;
         boolean ending = false;
-        do {
```
jglick: [reindentation](https://github.com/jenkinsci/workflow-cps-plugin/pull/46/files?w=1#diff-6452cd1bea46404d75cc1ba4559cc3adL357)

src/main/java/org/jenkinsci/plugins/workflow/cps/Safepoint.java
jglick: `@Restricted(NoExternalUse.class)`?

src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
```diff
@@ -236,4 +238,25 @@ public void addAction() throws Exception {
         p.addAction(a);
         assertNotNull(p.getAction(WhoAmI.class));
     }
+
+    @Test
```
jglick: `@Issue`

src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
```diff
@@ -236,4 +238,25 @@ public void addAction() throws Exception {
...
+        p.setDefinition(new CpsFlowDefinition(WorkflowJobNonRestartingTest.class.getName()+".going(); while (true) {}", true));
+
+        QueueTaskFuture<WorkflowRun> f = p.scheduleBuild2(0);
+        WorkflowRun b = f.getStartCondition().get(3, TimeUnit.SECONDS);
+
+        going.block(3000);    // get the buld going, which will loop infinitely
+        b.doStop(); // abort, abort!
+
+        f.get(3, TimeUnit.SECONDS); // wait for the result to come in
+        assertEquals(Result.ABORTED, b.getResult());
```
jglick: Suffices to write

```java
jenkins.assertBuildStatus(Result.ABORTED, jenkins.waitForCompletion(b));
```

src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
```diff
@@ -236,4 +238,25 @@ public void addAction() throws Exception {
...
     }
+
+    @Test
+    public void killInfiniteLoop() throws Exception {
+        p.setDefinition(new CpsFlowDefinition(WorkflowJobNonRestartingTest.class.getName()+".going(); while (true) {}", true));
+
+        QueueTaskFuture<WorkflowRun> f = p.scheduleBuild2(0);
+        WorkflowRun b = f.getStartCondition().get(3, TimeUnit.SECONDS);
+
+        going.block(3000);    // get the buld going, which will loop infinitely
```
jglick: There is a race condition here, that the test thread might continue after `going` is called but before the loop is entered. Whether that matters or not, I am not sure.

jglick:
:ant: the JIRA summary claims that this change makes `timeout` work on an infinite loop (the original motivating use case), yet there is no test proving that this actually happens—only that an _external_ abort works.

Other than that, :bee: though I scarcely understand the implementation.

src/main/java/org/jenkinsci/plugins/workflow/cps/Safepoint.java
kohsuke: yes

kohsuke:
Addressed the reviewer comment. Assuming the :bee: still stands, I'll leave it up to @jglick to decide the right moment to merge this, so that I won't rock any boat.

jglick:
re:bee:

src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
kohsuke: FTR, this has since been addressed by change in the definition of the script.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -347,56 +301,66 @@ public boolean isPaused() {
...
 
-            doneSomeWork |= changed;
-        } while (changed);
+                changed = true;
+            }
+        }
 
-        if (doneSomeWork) {
+        if (changed) {
             saveProgram();
```
jglick: Caused some regressions in programs which were legitimately broken. I am more concerned that the constant saving may have caused some performance regressions. Perhaps we need to throttle saves.

PR #47, opened 2016-08-22
Adding a helper method for testing doc generation.

abayer:
This would be handy for testing - e.g., it'd simplify https://github.com/jenkinsci/workflow-multibranch-plugin/issues/28

cc @reviewbybees esp @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
@@ -120,7 +120,7 @@ public class SnippetizerTester {
      * @param describableClass
      *     A {@link Class} implementing {@link Describable}
      * @throws Exception
-     *     If any errors are encountered other than {@link NoStaplerConstructorException}, which is ignored for now.
+     *     If any errors are encountered.
      */
     @SuppressWarnings("unchecked")
     public static void assertDocGeneration(Class<? extends Describable> describableClass) throws Exception {
@@ -133,8 +133,7 @@ public class SnippetizerTester {
     }
 
     private static void recurseOnTypes(ParameterType type) throws Exception {
-        // For the moment, only care about types with @DataBoundConstructors.
-        if (type instanceof ErrorType && !(((ErrorType)type).getError() instanceof NoStaplerConstructorException)) {
+        if (type instanceof ErrorType) {
             throw ((ErrorType)type).getError();
         }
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
```diff
@@ -103,6 +114,48 @@ protected Object invokeStep(StepDescriptor d, Object args) {
...
+
+        assertNotNull(model);
+
+        recurseOnModel(model);
+
+    }
+
+    private static void recurseOnTypes(ParameterType type) throws Exception {
+        // For the moment, only care about types with @DataBoundConstructors.
+        if (type instanceof ErrorType && !(((ErrorType)type).getError() instanceof NoStaplerConstructorException)) {
```
jglick: :ant: this seems fragile, because you will not know about `NoStaplerConstructorException`s which are genuine problems.

I would suggest adding a `Class<? extends Describable>...` parameter listing nested structs in the test classpath which you _know_ are messed up and you are OK with it for the moment.

abayer: Blech. Ok.

jglick:
:bee:

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
```diff
@@ -103,6 +114,47 @@ protected Object invokeStep(StepDescriptor d, Object args) {
...
+
+        if (type instanceof ArrayType) {
+            recurseOnTypes(((ArrayType)type).getElementType());
+        } else if (type instanceof HomogeneousObjectType) {
+            recurseOnModel(((HomogeneousObjectType) type).getSchemaType());
+        } else if (type instanceof HeterogeneousObjectType) {
+            for (Map.Entry<String, DescribableModel<?>> entry : ((HeterogeneousObjectType) type).getTypes().entrySet()) {
+                recurseOnModel(entry.getValue());
+            }
+        }
```
oleg-nenashev: 🐜 does nothing on unexpected type

oleg-nenashev:
🐜 for the missing check of unexpected classes, other code LGTM

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
abayer: By design - any other type is a no-op in the docs generation from our perspective, in that they can't contain other types, and any error would have been encountered on entering the type itself.

jglick:
re:bee:

PR #49, opened 2016-08-24
[FIXED JENKINS-29711] Handle monomorphic single-parameters to steps.

abayer:
[JENKINS-29711](https://issues.jenkins-ci.org/browse/JENKINS-29711)

- [x] Guts work.
- [x] Tests for non-meta-steps taking monomorphic single parameter.
- [x] Tests for meta-step taking monomorphic single parameter.

cc @reviewbybees esp @jglick @kohsuke @oleg-nenashev @svanoort

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -231,8 +231,15 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         List<StepDescriptor> metaSteps = StepDescriptor.metaStepsOf(symbol);
         StepDescriptor metaStep = metaSteps.size()==1 ? metaSteps.get(0) : null;
 
+        boolean singleArgumentOnly = false;
+        if (metaStep != null) {
+            DescribableModel<?> metaModel = new DescribableModel(metaStep.clazz);
+            singleArgumentOnly = metaModel.hasSingleRequiredParameter() && metaModel.getParameters().size() == 1;
+        }
+
         // The only time a closure is valid is when the resulting Describable is immediately executed via a meta-step
-        NamedArgsAndClosure args = parseArgs(_args, metaStep!=null && metaStep.takesImplicitBlockArgument(), UninstantiatedDescribable.ANONYMOUS_KEY);
+        NamedArgsAndClosure args = parseArgs(_args, metaStep!=null && metaStep.takesImplicitBlockArgument(),
+                UninstantiatedDescribable.ANONYMOUS_KEY, singleArgumentOnly);
         UninstantiatedDescribable ud = new UninstantiatedDescribable(symbol, null, args.namedArgs);
 
         if (metaStep==null) {
@@ -353,7 +360,14 @@ public class DSL extends GroovyObjectSupport implements Serializable {
     }
 
     static NamedArgsAndClosure parseArgs(Object arg, StepDescriptor d) {
-        return parseArgs(arg,d.takesImplicitBlockArgument(), loadSoleArgumentKey(d));
+        boolean singleArgumentOnly = false;
+        try {
+            DescribableModel<?> stepModel = new DescribableModel<>(d.clazz);
+            singleArgumentOnly = stepModel.hasSingleRequiredParameter() && stepModel.getParameters().size() == 1;
+        } catch (NoStaplerConstructorException e) {
+            // Ignore steps without databound constructors and treat them as normal.
+        }
+        return parseArgs(arg,d.takesImplicitBlockArgument(), loadSoleArgumentKey(d), singleArgumentOnly);
     }
 
     /**
@@ -379,7 +393,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
      *      If the context in which this method call happens allow implicit sole default argument, specify its name.
      *      If null, the call must be with names arguments.
      */
-    static NamedArgsAndClosure parseArgs(Object arg, boolean expectsBlock, String soleArgumentKey) {
+    static NamedArgsAndClosure parseArgs(Object arg, boolean expectsBlock, String soleArgumentKey, boolean singleRequiredArg) {
         if (arg instanceof NamedArgsAndClosure)
             return (NamedArgsAndClosure) arg;
         if (arg instanceof Map) // TODO is this clause actually used?
@@ -400,9 +414,9 @@ public class DSL extends GroovyObjectSupport implements Serializable {
                 a = a.subList(0,a.size()-1);
             }
 
-            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class")) {
+            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class") && !singleRequiredArg) {
                 // this is how Groovy passes in Map
-                return new NamedArgsAndClosure((Map)a.get(0),c);
+                return new NamedArgsAndClosure((Map) a.get(0), c);
             }
 
             switch (a.size()) {
--- a/src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
@@ -163,7 +163,7 @@ public class DSLTest {
     @Test
     public void monomorphicSymbol() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "monSymbol");
-        p.setDefinition(new CpsFlowDefinition("monomorphWithSymbolStep([firstArg:'one', secondArg:'two'])", true));
+        p.setDefinition(new CpsFlowDefinition("monomorphWithSymbolStep monomorphSymbol(firstArg: 'one', secondArg: 'two')", true));
         r.assertLogContains("First arg: one, second arg: two", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
@@ -185,7 +185,7 @@ public class DSLTest {
     @Test
     public void monomorphicListWithSymbol() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "monListSymbol");
-        p.setDefinition(new CpsFlowDefinition("monomorphListSymbolStep([[firstArg:'one', secondArg:'two'], [firstArg:'three', secondArg:'four']])", true));
+        p.setDefinition(new CpsFlowDefinition("monomorphListSymbolStep([monomorphSymbol(firstArg: 'one', secondArg: 'two'), monomorphSymbol(firstArg: 'three', secondArg: 'four')])", true));
         WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
         r.assertLogContains("First arg: one, second arg: two", b);
         r.assertLogContains("First arg: three, second arg: four", b);
--- a/src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
@@ -155,7 +155,7 @@ public class DSLTest {
     @Test
     public void monomorphic() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "mon");
-        p.setDefinition(new CpsFlowDefinition("monomorphStep firstArg:'one', secondArg:'two'"));
+        p.setDefinition(new CpsFlowDefinition("monomorphStep([firstArg:'one', secondArg:'two'])", true));
         r.assertLogContains("First arg: one, second arg: two", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
@@ -166,7 +166,7 @@ public class DSLTest {
     @Test
     public void monomorphicList() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "mon");
-        p.setDefinition(new CpsFlowDefinition("monomorphListStep([[firstArg:'one', secondArg:'two'], [firstArg:'three', secondArg:'four']])"));
+        p.setDefinition(new CpsFlowDefinition("monomorphListStep([[firstArg:'one', secondArg:'two'], [firstArg:'three', secondArg:'four']])", true));
         WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
         r.assertLogContains("First arg: one, second arg: two", b);
         r.assertLogContains("First arg: three, second arg: four", b);
--- a/src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
@@ -150,4 +150,16 @@ public class DSLTest {
         r.assertLogContains("First arg: one, second arg: two", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
+    /**
+     * Tests the ability to execute a step with an unnamed monomorphic list argument.
+     */
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphicList() throws Exception {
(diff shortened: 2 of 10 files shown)


abayer:
And it works! Still need to add a test here for the list case, though.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -379,7 +393,7 @@ static NamedArgsAndClosure parseArgs(Object arg, StepDescriptor d) {
      *      If the context in which this method call happens allow implicit sole default argument, specify its name.
      *      If null, the call must be with names arguments.
      */
-    static NamedArgsAndClosure parseArgs(Object arg, boolean expectsBlock, String soleArgumentKey) {
+    static NamedArgsAndClosure parseArgs(Object arg, boolean expectsBlock, String soleArgumentKey, boolean singleRequiredArg) {
```
oleg-nenashev: Are you sure there is no usages in other core Pipeline plugins. IIRC there were overlaps

abayer: For `parseArgs`? Good question.

abayer: Doesn't appear to be any, at least pre-split into multiple repos. Checking in more detail now.

abayer: Yup, can't find anything.

oleg-nenashev:
Looks good 🐝

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -262,4 +266,22 @@
         GroovyShell shell = new GroovyShell(r.jenkins.getPluginManager().uberClassLoader);
         shell.parse(dsld);
     }
+
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphic() throws Exception {
+        MonomorphicStep monomorphicStep = new MonomorphicStep(new MonomorphicData("one", "two"));
+        st.assertRoundTrip(monomorphicStep, "monomorphStep([firstArg: 'one', secondArg: 'two'])");
```
jglick: Good, `SnippetizerTester.assertRoundTrip` is how you know the fix is for real.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/MonomorphicListStep.java
```diff
@@ -0,0 +1,61 @@
...
+    public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+        public DescriptorImpl() {
+            super(Execution.class);
+        }
+
+        @Override public String getFunctionName() {
+            return "monomorphListStep";
+        }
+
+        @Override public String getDisplayName() {
```
jglick: Could omit this.

jglick:
:ant: I see that you are testing the originally reported cases, but there could still be an untested problem for the case that the monomorphic describable has a `@Symbol`, which would result in different surface syntax and potentially tickle different code paths.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
```diff
@@ -148,6 +148,30 @@ public void nonexistentFunctions() throws Exception {
...
     }
 
+    /**
+     * Tests the ability to execute a step with an unnamed monomorphic describable argument.
+     */
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphic() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "mon");
+        p.setDefinition(new CpsFlowDefinition("monomorphStep firstArg:'one', secondArg:'two'"));
```
jglick: :ant: pass `true` for `sandbox`. Probably makes no difference here, but best to test things in the sandbox by default, since occasionally some Groovy idiom will find up being compiled to something bizarre that the default whitelist rejects.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
```diff
@@ -148,6 +148,30 @@ public void nonexistentFunctions() throws Exception {
...
+    }
+
+    /**
+     * Tests the ability to execute a step with an unnamed monomorphic list argument.
+     */
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphicList() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "mon");
+        p.setDefinition(new CpsFlowDefinition("monomorphListStep([[firstArg:'one', secondArg:'two'], [firstArg:'three', secondArg:'four']])"));
```
jglick: :ant: ditto

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: > there were namespace overlaps

Not in this package.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
jglick: :bug: hold on, this does not make sense to me.

In `SnippetizerTest` you are asserting that `Snippetizer` will produce

```groovy
monomorphStep([firstArg: 'one', secondArg: 'two'])
```

which is a syntax that has always worked in `DSL` (AFAIK), so we are just making sure that users are offered that instead of

```groovy
monomorphStep [firstArg: 'one', secondArg: 'two']
```

which did not work—fine.

But now here you seem to be patching `DSL` so that it will also accept a new and simplified syntax

```groovy
monomorphStep firstArg: 'one', secondArg: 'two'
```

That may be fine—I do not completely follow the details of that part of the patch, and probably you should read and delete [this comment from @kohsuke](https://github.com/jenkinsci/workflow-cps-plugin/blob/f5e4f4eafee42f1096a33baf18bfe69726d6d1e5/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java#L296-L324) if you want to go that way—but then surely `Snippetizer` should also be patched to _generate_ the simplified syntax?

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
abayer: Exactly. =)

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
abayer: Huh, good point. Let me think about that.

abayer: Oh, wait - so in practice, 
```groovy
monomorphStep([firstArg:'one', secondArg: 'two'])
```
and
```groovy
monomorphStep firstArg:'one', secondArg:'two'
```
Are effectively identical - they're both passing a `Map` to `monomorphStep`. I don't *like* the no-parens-at-all syntax, but it works (as should parens-but-not-brackets, which I'll test for). But if the general consensus is no-parens-at-all is the way to go, I'll tweak `Snippetizer` accordingly.

jglick: Anyway I do not think such a patch to `DSL` is even necessary for usability, since you can already add a `@Symbol` to the nested struct and then mark the step as a metastep and get similar syntax:

```groovy
monomorphStep monomorph(firstArg: 'one', secondArg: 'two')
```

can become simply

```groovy
monomorph firstArg: 'one', secondArg: 'two'
```

as of #28. Which reinforces my request to see analogous tests for `@Symbol`ic structs.

@oleg-nenashev’s original example with `PublishHTMLStep` is a perfect example of this. We would want to make this a metastep, add `@Symbol("publishHTML")` to `HtmlPublisherTarget.DescriptorImpl`, make `HtmlPublisherTarget` use `@DataBoundSetter` properly (leaving only `reportDir` in the `@DataBoundConstructor`), and make sure that `DSL` and `Snippetizer` understand that there is no real ambiguity in

```groovy
publishHTML 'target/javadoc'
```

abayer:
@jglick Can you give me an example of the `@Symbol` case?

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
abayer: So...we only care about the `@Symbol` case at all? I'm confused.

abayer: Also, turns out to be a lot harder to get it to exclude parentheses and brackets for cases like `metamorphStep` but not ones featuring `$class` as a key. So...I say let's just leave it as is, with `metamorphStep([firstArg:'one', secondArg:'two'])` as the output. It ain't the most aesthetically pleasing, but it *works*, so...

abayer:
Joy. The `@Symbol` test comes out with `monomorphWithSymbolStep monomorphSymbol(firstArg: 'one', secondArg: 'two')`...

abayer:
@jglick Updating with what I have now, without removing the parens and brackets from non-`@Symbol` monomorphic cases, because it's a pain in the arse that I don't think is worth the hassle, and with the `@Symbol` monomorphic case generating `monomorphWithSymbolStep monomorphSymbol(firstArg: 'one', secondArg: 'two')` because, well, that's what it generates.

jglick:
> The `@Symbol` test comes out with `monomorphWithSymbolStep monomorphSymbol(firstArg: 'one', secondArg: 'two')`

That is the expected current behavior, at least for a non-metastep.

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -284,4 +286,11 @@ public void monomorphicList() throws Exception {
         st.assertRoundTrip(monomorphicStep, "monomorphListStep([[firstArg: 'one', secondArg: 'two'], [firstArg: 'three', secondArg: 'four']])");
     }
 
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphicSymbol() throws Exception {
+        MonomorphicWithSymbolStep monomorphicStep = new MonomorphicWithSymbolStep(new MonomorphicDataWithSymbol("one", "two"));
+        st.assertRoundTrip(monomorphicStep, "monomorphWithSymbolStep monomorphSymbol(firstArg: 'one', secondArg: 'two')");
```
jglick: Please also test symbols with lists. I believe you need

```groovy
somestep([monomorphSymbol(…)])
```

jglick: (which, last I checked, is what was broken for `properties`)

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
```diff
@@ -148,6 +148,30 @@ public void nonexistentFunctions() throws Exception {
...
     }
 
+    /**
+     * Tests the ability to execute a step with an unnamed monomorphic describable argument.
+     */
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphic() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "mon");
+        p.setDefinition(new CpsFlowDefinition("monomorphStep([firstArg:'one', secondArg:'two'])", true));
```
jglick: So, since IIUC this is what was working before, why is the patch to `DSL` still here? In other words, if you revert that part, does any current test fail?

jglick:
still a :bug: pending some more explanation and testing

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
abayer: Let's find out...

abayer: Yup, back to the `rhs is null while lhs=org.jenkinsci.plugins.workflow.testMetaStep.MonomorphicData@3c88191b` error without the `DSL` changes.

abayer: So I'm leaving the `DSL` changes in. =)

jglick: What is this error? In which test? What user-visible problem does it correspond to? My impression was that the only actual problem was what `Snippetizer` suggested, but if you added the parentheses, your script would work.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
```diff
@@ -159,19 +159,39 @@ public void monomorphic() throws Exception {
...
         WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
         r.assertLogContains("First arg: one, second arg: two", b);
         r.assertLogContains("First arg: three, second arg: four", b);
     }
 
+    @Issue("JENKINS-29711")
+    @Test
+    public void monomorphicListWithSymbol() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "monListSymbol");
+        p.setDefinition(new CpsFlowDefinition("monomorphListSymbolStep([[firstArg:'one', secondArg:'two'], [firstArg:'three', secondArg:'four']])", true));
```
jglick: Why are you testing the deprecated syntax? If there is a symbol, you should be using it.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
abayer: Round-tripping *anything* with a monomorphic single argument will result in that error, even with the parentheses. So.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
abayer: Good question. Switching.

src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
abayer: Oh, and the failing test in this specific case was `SnippetizerTest.monomorphic`. The changes in `DSL` are literally the main point of this PR (well, that plus the `List` handling).

abayer: And no, no pre-existing tests from before this PR failed due to that change not being there...because we weren't running any tests against this use case! Now we are in the tests I've added in this PR. Ergo.

jglick: So how did, say, `publishHTML` work until now?

abayer: I don't believe the snippet generator-generated code worked.

jglick:
OK, I think I understand what was baffling me here.

In both cases `Snippetizer` was suggesting something that did not work, though for distinct reasons.

In one case (list, like `properties`) you fixed `Snippetizer` to reflect what did work, without needing to do anything to `DSL`; and in the other case (singleton, like `publishHTML`) you fixed `DSL` so that the desired syntax worked, without changing what `Snippetizer` produces.

So in some sense two related bugs, both fixed here.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/MonomorphicListStep.java
```diff
@@ -0,0 +1,61 @@
+package org.jenkinsci.plugins.workflow.testMetaStep;
```
jglick: :ant: package is a bit misleading for these tests, but OK

jglick:
:bee: with that cleared up

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/MonomorphicListStep.java
abayer: Yeah, I could move 'em or rename the package, but...eh.

PR #45, opened 2016-08-18
[JENKINS-31155] Infrastructure for shared libraries

jglick:
Prerequisites of https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/10.

@reviewbybees

PR #55, opened 2016-09-13
[JENKINS-38167] NPE from scripts using @Field with instance initializers loading properties

jglick:
Fixes a regression from #45.

@reviewbybees


armfergom:
🐝

recena:
:bee:

PR #56, opened 2016-09-13
[FIXED JENKINS-38169] Treat single required+named param differently.

abayer:
[JENKINS-38169](https://issues.jenkins-ci.org/browse/JENKINS-38169)

If you call `echo message:'foo'` previous to this change, the Map that
gets passed as the first element of args ends up being used as the
sole parameter itself, so you end up with `name:[name:'foo']` being
passed as the args. That's obviously wrong. So let's add a check to
see if there is a `soleArgumentKey`, the arg map only has one
key/value, and that key is equal to `soleArgumentKey` - if that's
true, just pass the arg map directly.

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -424,9 +424,13 @@ public class DSL extends GroovyObjectSupport implements Serializable {
                 a = a.subList(0,a.size()-1);
             }
 
-            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class") && !singleRequiredArg) {
-                // this is how Groovy passes in Map
-                return new NamedArgsAndClosure((Map) a.get(0), c);
+            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class")) {
+                Map mapArg = (Map) a.get(0);
+                if (!singleRequiredArg ||
+                        (soleArgumentKey != null && mapArg.size() == 1 && mapArg.containsKey(soleArgumentKey))) {
+                    // this is how Groovy passes in Map
+                    return new NamedArgsAndClosure(mapArg, c);
+                }
             }
 
             switch (a.size()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -424,9 +424,13 @@ static NamedArgsAndClosure parseArgs(Object arg, boolean expectsBlock, String so
                 a = a.subList(0,a.size()-1);
             }
 
-            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class") && !singleRequiredArg) {
-                // this is how Groovy passes in Map
-                return new NamedArgsAndClosure((Map) a.get(0), c);
+            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class")) {
+                Map mapArg = (Map) a.get(0);
+                if (!singleRequiredArg ||
+                        (soleArgumentKey != null && mapArg.size() == 1 && mapArg.containsKey(soleArgumentKey))) {
```
jglick: You have totally lost me. Thank goodness for tests!

jglick:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -424,9 +424,13 @@ static NamedArgsAndClosure parseArgs(Object arg, boolean expectsBlock, String so
                 a = a.subList(0,a.size()-1);
             }
 
-            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class") && !singleRequiredArg) {
-                // this is how Groovy passes in Map
-                return new NamedArgsAndClosure((Map) a.get(0), c);
+            if (a.size()==1 && a.get(0) instanceof Map && !((Map) a.get(0)).containsKey("$class")) {
+                Map mapArg = (Map) a.get(0);
```
jglick: You could have moved the `$class` check above.

daniel-beck:
Tentative :bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: More accurately, @kohsuke has totally lost all of us. =)

abayer:
FWIW, opened [JENKINS-38172](https://issues.jenkins-ci.org/browse/JENKINS-38172) at me for actually documenting what this all does and why.

PR #58, opened 2016-09-16
Add anchors for variables

daniel-beck:
This allows links to specific variables like this: `/pipeline-syntax/globals#env`

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
@@ -26,8 +26,10 @@ l.layout(title:_("Pipeline Syntax: Global Variable Reference"), norefresh: true)
           h2(_("Variables"))
           dl(class:'steps variables root'){
             for (GlobalVariable v : snippetizer.getGlobalVariables()) {
-              dt {
-                code(v.getName())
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/globals.groovy
```diff
@@ -26,8 +26,10 @@ l.layout(title:_("Pipeline Syntax: Global Variable Reference"), norefresh: true)
           h2(_("Variables"))
           dl(class:'steps variables root'){
             for (GlobalVariable v : snippetizer.getGlobalVariables()) {
-              dt {
-                code(v.getName())
+              dt(id: v.getName()) {
+                a(href: '#' + v.getName()) {
```
jglick: Huh? This will create a link to a nonexistent anchor.

I think what you meant was

```groovy
dt {
  a(name: v.name) {
    code(v.name)
  }
}
```

Did you test this in a browser?

daniel-beck: Yes, and `id` creates anchors as well.

daniel-beck:
The idea with the link is that you can also click the name to navigate to the anchor and copy & paste from the address bar.

jglick:
OK, :bee: if it works as described.

PR #59, opened 2016-09-17
[JENKINS-37011] step implemented in Groovy (take 2)

kohsuke:
[JENKINS-37011](https://issues.jenkins-ci.org/browse/JENKINS-37011)

Based on the discussion in https://github.com/jenkinsci/workflow-cps-plugin/pull/34, setting a less magical goal. See the test case and `doc/step-in-groovy.md` for the plugin developer experience.

@reviewbybees in particular @jglick and @abayer


abayer:
wooooo! =)

doc/step-in-groovy.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Groovy steps and sandbox
+`GroovyStepExecution` subtypes are trusted,
+so it is not subject to the sandbox execution environment nor the script approval process.
+IOW, it can invoke any Java code.
+
+User-written untrusted `Jenkinsfile` can in turn access `GroovyStepExecution`,
+and this can even ignore Java access control rules like private fields.
+
+If you have sensitive fields that you need to protect from malicious `Jenkinsfile`,
```
abayer: Tiny nit - maybe "`Jenkinsfile`s"?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStep.java
```diff
@@ -0,0 +1,29 @@
...
+import org.jenkinsci.plugins.workflow.cps.CpsThread;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
+
+/**
+ * @author Kohsuke Kawaguchi
+ * @see "doc/step-in-groovy.md"
+ */
+public abstract class GroovyStep extends Step {
```
abayer: Another little nit - I don't know if `GroovyStep` is the best class name?

kohsuke:
This PR is feature complete

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStep.java
kohsuke: Name suggestion welcome

doc/step-in-groovy.md
kohsuke: Updated

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStep.java
jglick: Seems intuitive enough to me.

jglick:
From what I can understand of it, this looks right, except for the questionable serialization of `step`. Missing test coverage (which I could work on if you prefer):

* non-block-scoped step
* `checkpoint` integration (downstream)
* flow graph structure
* script security (a bit open-ended, but to match your claims and warnings in the documentation)
* multiple (or zero) body invocations
* accessing structured (`Describable`/`Collection`) fields in step
* fields defined in execution (or do we simply want to forbid them in favor of local variables?)
* using contextual objects from another step (e.g., running a Groovy step calling `sh` from inside a `node` block)
* returning a value from the step
* throwing an exception from the step (esp. `error` ~ `AbortException` should elide stack trace in log)

jglick:
```
Possible null pointer dereference in new org.jenkinsci.plugins.workflow.cps.steps.ingroovy.GroovyStepExecutionLoader(GroovyResourceLoader) due to return value of called method [org.jenkinsci.plugins.workflow.cps.steps.ingroovy.GroovyStepExecutionLoader, org.jenkinsci.plugins.workflow.cps.steps.ingroovy.GroovyStepExecutionLoader] Dereferenced at GroovyStepExecutionLoader.java:[line 23]Known null at GroovyStepExecutionLoader.java:[line 23] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE
```

kohsuke:
I'll add tests, no worries.

kohsuke:
Any good test case that checks `FlowNode` structure?

kohsuke:
Boy lots of test cases to cover...

- [ ] non-block-scoped step
- [ ] checkpoint integration (downstream)
- [ ] flow graph structure
- [ ] script security (a bit open-ended, but to match your claims and warnings in the documentation)
- [ ] multiple (or zero) body invocations
- [ ] accessing structured (Describable/Collection) fields in step
- [ ] fields defined in execution (or do we simply want to forbid them in favor of local variables?)
- [ ] using contextual objects from another step (e.g., running a Groovy step calling sh from inside a node block)
- [ ] returning a value from the step
- [ ] throwing an exception from the step (esp. error ~ AbortException should elide stack trace in log)

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
```diff
@@ -0,0 +1,149 @@
...
+    public void stop(@Nonnull Throwable cause) throws Exception {
+        execution.cancel(cause);
+    }
+
+    /**
+     * Persist {@link #step} in a form that's serializable.
+     */
+    private void writeObject(ObjectOutputStream oos) throws IOException {
+        oos.defaultWriteObject();
+        oos.writeObject(UninstantiatedDescribable.from(step));
```
jglick: Fixed now I think…not sure why GitHub is not marking it as a comment on an outdated diff.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,96 @@
...
+From Jenkins Pipeline, you can use this like you'd use any other steps:
+
+```
+myRetryClone(3) {
+  sh './flakyTest.sh'
+}
+```
+
+Note the directory in which the files are defined. `GroovyStep` and `GroovyStepDescriptor`
+are defined in `src/main/java` and is written just like any other steps.
```
kohsuke: This is a wrong place to discuss how a `Step` should be written. There should be another document that discusses that, and this should just link to it.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,96 @@
...
+IOW, any additional fields you define in `GroovyStep` will not be persisted. Only the parameters
+users have given in Pipeline Script will.
+
+## Groovy steps and sandbox
+`GroovyStepExecution` subtypes are trusted,
+so it is not subject to the sandbox execution environment nor the script approval process.
+IOW, it can invoke any Java code.
+
+User-written untrusted `Jenkinsfile` can in turn access `GroovyStepExecution`,
+and this can even ignore Java access control rules like private fields.
```
kohsuke: By instantiating it explicitly, for example, outside what the pipeline runtime system does. The test case covers this. I assume you are not saying the documentation should discuss how this attack vector works.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -160,7 +161,8 @@ protected Object invokeStep(StepDescriptor d, Object args) {
         FlowNode an;
 
         // TODO: generalize the notion of Step taking over the FlowNode creation.
-        boolean hack = d instanceof ParallelStep.DescriptorImpl || d instanceof LoadStep.DescriptorImpl;
+        boolean hack = d instanceof ParallelStep.DescriptorImpl || d instanceof LoadStep.DescriptorImpl
+                || d instanceof GroovyStepDescriptor;
```
kohsuke: Yes, as captured in the TODO comment above. I consider this a separate task.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
```diff
@@ -0,0 +1,149 @@
...
+    public GroovyStep getStep() {
+        return step;
+    }
+
+    @Override
+    public boolean start() throws Exception {
+        CpsStepContext cps = getContext();
+        CpsThread t = CpsThread.current();
+
+        // TODO: make sure this class is actually CPS transformed
```
kohsuke: This is a TODO for catching common mistake that a plugin developer will make that's hard to diagnose.

This requires groovy-cps improvement. I'll come back to this later.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
```diff
@@ -0,0 +1,149 @@
...
+//                .withStartAction(/*... maybe a marker for the future?*/)
+                .withCallback(BodyExecutionCallback.wrap(cps))
+                .start();
+
+        return false;
+    }
+
+    @Override
+    public void stop(@Nonnull Throwable cause) throws Exception {
+        execution.cancel(cause);
```
kohsuke: I don't know if the pipeline foundation calls this or not, but it is a part of the API, so I think it's good to implement it.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
kohsuke: Fixed

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
```diff
@@ -0,0 +1,149 @@
...
+            step = (GroovyStep)ud.instantiate();
+        } catch (Exception e) {
+            throw new IOException("Failed to instantiate "+ud,e);
+        }
+    }
+
+    /**
+     * Attempt to resolve a method call as a step invocation like Pipeline Script.
+     */
+    public Object methodMissing(String name, Object args) throws IOException {
```
kohsuke: This method is invoked via convention. I added a link to the Groovy documentation about this.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
```diff
@@ -0,0 +1,149 @@
...
+                throw new AssertionError();
+            }
+        };
+    }
+
+    /**
+     * Passed to the user-written 'call' method in Groovy as 'body'.
+     * When invoked, executes the body block passed to the step.
+     */
+    private final class Body extends Closure implements Serializable {
```
kohsuke: Replaced with `@PersistIn(PROGRAM)` to call an attention to the fact that this is serialized.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecutionLoader.java
```diff
@@ -0,0 +1,39 @@
...
+
+    GroovyStepExecutionLoader(GroovyResourceLoader base) {
+        this.base = base;
+        this.uberClassLoader = Jenkins.getInstance().getPluginManager().uberClassLoader;
+    }
+
+    @Override
+    public URL loadGroovySource(String className) throws MalformedURLException {
+        for (StepDescriptor d : StepDescriptor.all()) {
+            if (className.equals(d.clazz.getName()+"Execution")) {
```
kohsuke: Fixed

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyTest.java
```diff
@@ -0,0 +1,58 @@
...
+                        "helloWorldGroovy('Duke') {\n" +
+                            "echo 'Hello body'\n"+
+                            "semaphore 'restart'\n"+
+                            "echo 'Good bye body'\n"+
+                        "}"));
+                WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
+                SemaphoreStep.waitForStart("restart/1", b);
+                story.j.waitForMessage("Hello body",b);
+                story.j.assertLogContains("Hello Duke",b);
+                story.j.assertLogContains("Hello body",b);
```
kohsuke: Fixed

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyTest.java
```diff
@@ -0,0 +1,58 @@
...
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("demo", WorkflowJob.class);
+                WorkflowRun b = p.getLastBuild();
+                SemaphoreStep.success("restart/1", null);
+                story.j.waitForCompletion(b);
+
+                story.j.assertLogContains("Hello Duke",b);
+                story.j.assertLogContains("Hello body",b);
```
kohsuke: Fixed

doc/step-in-groovy.md
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+}
+```
+
+```groovy
+// src/main/resources/acme/RetryStepExecution.groovy
+package acme;
+
+public class RetryStepExecution extends GroovyStepExecution {
```
kohsuke: Okay.
I'm from the camp that the point of an example is to highlight important bits, and if compiler accuracy gets in the way it should be omitted.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStep.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public GroovyStepDescriptor getDescriptor() {
+        return (GroovyStepDescriptor) super.getDescriptor();
+    }
+
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
```
kohsuke: I don't feel too strongly. I think it's OK.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/StepInGroovyTest.java
```diff
@@ -0,0 +1,58 @@
...
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition(
+                        "helloWorldGroovy('Duke') {\n" +
+                            "echo 'Hello body'\n"+
+                            "semaphore 'restart'\n"+
+                            "echo 'Good bye body'\n"+
+                        "}"));
```
kohsuke: Fixed

kohsuke:
Recent GitHub review made it really hard to see outstanding comments, but I believe I addressed all of them that should be addressed.

I think I delivered all the test cases except the downstream one, and the way it decorates `GroovyClassLoader` makes it fairly safe there.

Let's merge this! I want this to be my key accomplishment this weekend.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
jglick: Ah, I misinterpreted the comment entirely. You mean: make sure the _subclass_ of this class that the user supplied is actually CPS-transformed.

doc/step-in-groovy.md
```diff
@@ -95,3 +102,19 @@ define a subtype of `GroovyStepExecution` in Java and put those fields there,
...
+`StepStartNode` and `StepEndNode`, regardless of whether the step
+takes a body closure. All the steps invoked from the Groovy code
+are recorded as flow nodes inside this block.
+
+In contrast, a step written in Java only gets a block-scoped flow
+nodes when it has a body.
+
+When a Groovy step takes a body closure and invokes it, it will not
+automatically create a block-scoped step node. We intend to provide
+a step definition you can use if such block-scoped flow node is desirable.
```
jglick: I think you could simply write

```groovy
stage('whatever you like here') {body()}
```

right/

kohsuke: My understanding is that this use of the `stage` step is still contested, so I refrained from recommending that to plugin developers.

Since we explain `stage` primarily for users to organize a pipeline into stages, using the same thing for mere grouping of `FlowNode` is confusing for plugin developers. Even if it is the same code internally, I'd think it makes sense to define an alias.

More likely, I'd think we want to leave a different `FlowNode` for this kind of grouping.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepTest.java
```diff
@@ -0,0 +1,286 @@
...
+                // Groovy step should have access to the money in the vault
+                p.setDefinition(new CpsFlowDefinition(
+                        "assert bankTeller()==2000",
+                        true));
+                story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+                // but no direct access allowed
+                String className = BankTellerSecureStepExecution.class.getName().replace("Secure","");
+                p.setDefinition(new CpsFlowDefinition(
+                        "def o = new "+ className +"()\n"+
```
jglick: I think it would be better to simply blacklist the constructors of `GroovyStepExecution` subtypes, so you could not get even this far. After all there is no legitimate reason for untrusted code to be calling this. Then there would be no need for script authors to do the `BankTellerSecureStepExecution` trick, which I fear is too easily neglected.

jglick: For example, `GroovyClassLoaderWhitelist.permits` could start

```java
if (GroovyStepExecution.class.isAssignableFrom(declaringClass)) {
    return false;
}
```

kohsuke: From defense in depth perspective, I think we still need to advise plugin developers about this. All it takes is one reference to an object, for example with some return values, etc.

Sure, we can reject the instantiation of `GroovyStepExecution`, and it only adds to the depth of defense, but arbitrary rules like that is an indication to a deeper problem. Perhaps a better rejection rule is to prohibit code in untrusted classloader to access private fields/methods of trusted classloader.

jglick:
> we can reject the instantiation of `GroovyStepExecution`

Well the change I proposed would reject any access to any member defined in a `GroovyStepExecution` subtype. Granted it would not cover nested classes without a bit more work.

> Perhaps a better rejection rule is to prohibit code in untrusted classloader to access private fields/methods of trusted classloader.

That would be helpful, and is desirable for global libraries.

I think the crux of the issue is that while in a global library we _expect_ untrusted code to access members defined in the trusted code, in this case we do not want _any_ accesses to the trusted code—the step invocation itself is routed indirectly through `DSL`, and that is all there should be. The only case I can think of where you would want untrusted code accessing trusted code in a Groovy step is

```groovy
// pkg/MyExecution.groovy
interface Callback {
    void notifyMeOfSomething()
}
def call(Closure body) {
  int notifications
  body(new Callback() {
    void notifyMeOfSomething() {notifications++}
  })
  echo "notified me ${notifications} times"
}
// Jenkinsfile
myExecution {callback ->
  callback.notifyMeOfSomething()
  callback.notifyMeOfSomething()
}
```

In this case where the access is _expected_ you could add `@Whitelisted` to the definition of `notifyMeOfSomething`.

I thought of loading `GroovyStepExecution` implementations in a separate trusted `GroovyClassLoader` which is not a `parent` of that loading `WorkflowScript`. That would address the constructor issue, but not dynamic method lookup. I suspect it would suffice to make `GroovyClassLoaderWhitelist.permits` call `getEnclosingClass()` recursively. (If we even load nested types at all—from looking at the `ResourceLoader` code I think we might not, and there is no test demonstrating it working.)

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
jglick: There is a JIRA about it. tl;dr: AFAICT it will not be called. If it is, I suppose the current implementation is fine.

jglick: …because the exception will be delivered to the innermost running `StepExecution` or Groovy code. Right?

doc/step-in-groovy.md
jglick: Well I was going to call it `label` but you insisted on calling it `stage`. :-) At any rate, I think this can be solved with a step if anyone cares, so this is fine for now.

doc/step-in-groovy.md
jglick: If we have no choice but to leave this so wide open, then yes I think the user documentation _should_ explain how the attack vector would work, so you can be sure you are defending against it properly.

jglick:
> all the test cases except the downstream one, and the way it decorates `GroovyClassLoader` makes it fairly safe there.

In fact it does not work at all, but I will leave the details to a downstream PR.

kohsuke:
Syncing up offline with Jesse, remaining work:

- [ ] access to `GroovyStepExecution` and its associated nested types are blacklisted, except where expclitily whitelisted
- [ ] fix downstream PR

jglick:
> blacklisted, except where expclitily whitelisted

Looks like that is now fixed, right?

kohsuke:
I need to now sync up with @abayer to make sure his original use case is well supported; that is, pipeline-model plugin has non-trivial object model that should all run as CPS code.

To make that work well I think I really need content root for code that runs in trusted classloader.

abayer:
@kohsuke I think this'll suffice - I'll still have a little pain with `ClosureModelTranslator` and `PropertiesToMapTranslator` in `pipeline-model-definition`, since those don't translate to `Step`s in any useful manner, but I can live with that (and/or find an alternate approach). My main work is in `ModelInterpreter`, which translates to a `Step` perfectly.

abayer:
@kohsuke @jglick FYI, first effort at a proof of concept using this is up at https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/20

jglick:
:bee: although it would be wise to wait for confirmation that this works for the more complex cases in `pipeline-model-definition`.

abayer:
@kohsuke Is it possible to get a return value from a `GroovyStepExecution`, like `LoadStep` or `ReadFileStep`?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecutionCompiler.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Used to compile CPS-transformed code that does not run in the sandbox,
+ * but unlike {@linkplain CpsFlowExecution#getTrustedShell() trusted shell}, code
+ * loaded here will not automatically become visible to untrusted code.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+@Extension
+public class GroovyStepExecutionCompiler {
+//    private GroovyShell sh;
```
abayer: :bug: I really hate stumbling across commented out code like this where I don't know if it's actually valid, why it's there, why it's commented out, etc...so I'd strongly appreciate if it was removed. =)

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecutionCompiler.java
```diff
@@ -0,0 +1,49 @@
...
+     * cannot call methods on these classes unless explicitly whitelisted.
+     */
+    /*package*/ GroovyShell createShell(ClassLoader parent) {
+        GroovyShell sh = new CpsGroovyShellFactory(null).withParent(parent).build();
+        GroovyClassLoader cl = sh.getClassLoader();
+        cl.setResourceLoader(new GroovyStepExecutionLoader(cl.getResourceLoader()));
+        return sh;
+    }
+
+//    /*package*/ GroovyClassLoader getClassLoader() {
```
abayer: Same as above.

doc/step-in-groovy.md
```diff
@@ -0,0 +1,116 @@
...
+  sh './flakyTest.sh'
+}
+```
+
+Note the directory in which the files are defined. `GroovyStep` and `GroovyStepDescriptor`
+are defined in `src/main/java` and is written just like any other steps.
+`GroovyStepExecution` is defined in `src/main/resources` to be packaged into the plugin
+in the source form, so that it can run in the CPS groovy interpreter at the runtime.
+
+The `call` method on `GroovyStepExecution` subtype defines the actual logic of step.
```
abayer: I just asked whether it's possible to return a value from a `GroovyStepExecution` along the lines of, say, `ReadFileExecution` - it'd be nice to have some explanation of what the return type of `call` can be.

abayer:
@kohsuke Ok, I figured out how to return stuff, but now I need access to the `CpsScript` within my `GroovyStepExecution`...

abayer:
Not sure yet, but there seems to be a problem when you've got nested closures in the body - I just get hanging forever. Gotta do some more experimentation to get a test case.

abayer:
:bug: Found a definite problem - try putting `echo docker.image('ubuntu').imageName` inside a `GroovyStepExecution` and watch it fail hard. Specifically, in a kinda hacky test I set up:

```
java.lang.ClassNotFoundException: org.jenkinsci.plugins.docker.workflow.Docker
	at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
	at org.jenkinsci.plugins.docker.workflow.DockerDSL.getValue(DockerDSL.java:47)
	at org.jenkinsci.plugins.workflow.cps.CpsScript.getProperty(CpsScript.java:128)
	at org.jenkinsci.plugins.workflow.cps.steps.ingroovy.GroovyStepExecution.propertyMissing(GroovyStepExecution.java:107)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
	at groovy.lang.MetaClassImpl.invokeMissingProperty(MetaClassImpl.java:880)
	at groovy.lang.MetaClassImpl.getProperty(MetaClassImpl.java:1861)
	at groovy.lang.MetaClassImpl.getProperty(MetaClassImpl.java:3735)
	at org.jenkinsci.plugins.pipeline.modeldefinition.steps.ModelInterpreterStepExecution.getProperty(ModelInterpreterStepExecution.groovy)
	at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:172)
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.getProperty(ScriptBytecodeAdapter.java:456)
	at com.cloudbees.groovy.cps.sandbox.DefaultInvoker.getProperty(DefaultInvoker.java:26)
	at com.cloudbees.groovy.cps.impl.PropertyAccessBlock.rawGet(PropertyAccessBlock.java:20)
	at org.jenkinsci.plugins.pipeline.modeldefinition.steps.ModelInterpreterStepExecution.call(/Users/abayer/IdeaProjects/pipeline-config-plugin/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ModelInterpreterStepExecution.groovy:53)
	at ___cps.transform___(Native Method)
	at com.cloudbees.groovy.cps.impl.PropertyishBlock$ContinuationImpl.get(PropertyishBlock.java:74)
	at com.cloudbees.groovy.cps.LValueBlock$GetAdapter.receive(LValueBlock.java:30)
	at com.cloudbees.groovy.cps.impl.PropertyishBlock$ContinuationImpl.fixName(PropertyishBlock.java:66)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at com.cloudbees.groovy.cps.impl.ContinuationPtr$ContinuationImpl.receive(ContinuationPtr.java:72)
	at com.cloudbees.groovy.cps.impl.ConstantBlock.eval(ConstantBlock.java:21)
	at com.cloudbees.groovy.cps.Next.step(Next.java:58)
	at com.cloudbees.groovy.cps.Continuable.run0(Continuable.java:154)
	at org.jenkinsci.plugins.workflow.cps.SandboxContinuable.access$001(SandboxContinuable.java:18)
	at org.jenkinsci.plugins.workflow.cps.SandboxContinuable$1.call(SandboxContinuable.java:33)
	at org.jenkinsci.plugins.workflow.cps.SandboxContinuable$1.call(SandboxContinuable.java:30)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovySandbox.runInSandbox(GroovySandbox.java:108)
	at org.jenkinsci.plugins.workflow.cps.SandboxContinuable.run0(SandboxContinuable.java:30)
	at org.jenkinsci.plugins.workflow.cps.CpsThread.runNextChunk(CpsThread.java:164)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:324)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.access$100(CpsThreadGroup.java:78)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:236)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:224)
	at org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$2.call(CpsVmExecutorService.java:47)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at hudson.remoting.SingleLaneExecutorService$1.run(SingleLaneExecutorService.java:112)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
```

The problem seems to be that the classloader on the `CpsScript` instance passed to the `GlobalVariable`s now isn't a `GroovyClassLoader`...

abayer:
The hanging issue seems to be at the core that I wanted to just pass `Closure`s to a step as a parameter, not as a Pipeline step block, so the currying that results from `takesImplicitBlockArgument` being true and `GroovyStepExecution.Body` seem to cause problems when the `Closure` contents aren't actually steps at all. When I changed to instead having a `@DataBoundConstructor` parameter of type `Closure`, with a no-args `call()` method instead of `call(Closure body)`, pulling the `Closure` I need from the step itself, I got past the hanging.

abayer:
@kohsuke Another question/issue - I've been [executing steps from a closure like this](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/b385ba5d5015504e91ba9e297283282bd4ce94a5/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy#L81-L86), using `script` as the delegate. Without access to the `CpsScript script`, I'm not sure what the right way to do this would be. Just delegate to `this`? Will that result in inadvertently bypassing script security?

abayer:
@kohsuke The fake `CpsScript` at https://github.com/jenkinsci/workflow-cps-plugin/blob/d535b2951b2b2361356f2c4e194fdc0a95bbe3ff/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java#L115-L115 seems to be the problem for loading classes, fyi.

abayer:
Further experimenting on the loading classes front - I tried adding an override of `start(StepContext)` in my `GroovyStep` extension that did a `CpsThread.get().getExecution().getShell().getClassLoader().loadClass("foo")` and that worked. Like, I think I actually have things functionalish.

abayer:
...but that does still bork running `docker.image('foo').inside { ... }` within a `GroovyStepExecution` (or a body passed to a `GroovyStepExecution`, so far as I can tell) so it's just a workaround for the moment.

abayer:
@kohsuke @jglick First work in progress that actually kinda works is up at https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/18 - tests will *definitely* fail, and in fact, they'll fail horribly unless you've got all three of https://github.com/jenkinsci/workflow-cps-plugin/pull/65, https://github.com/jenkinsci/workflow-cps-plugin/pull/64 and https://github.com/jenkinsci/workflow-cps-plugin/pull/59 merged together and `mvn clean install`'d as `2.18-SNAPSHOT`.

As mentioned above, there are horrific hacks involved in this, `docker.image('foo').inside { ... }` still doesn't work, and, frankly, I haven't run all the tests yet, but `BasicModelDefTest.simplePipeline` passes, so that's a start. =)

jglick:
> having a `@DataBoundConstructor` parameter of type `Closure`

If you find yourself doing that, then you should not be using this facility. This is for creating _steps_, which are things with a meaningful `config` view (with the exception of `parallel` which arguably should never have been defined as a step at all). Whatever you were doing before is what you should stick with.

abayer:
Wellllll, @kohsuke asked me to try converting it, so *shrug*?

jglick:
But what do you gain? The point is having something which _Snippet Generator_ could assist with the syntax of. Obviously that is not going to happen for something which takes a closure, because that is not even remotely something that a form could submit. So you might as well just use a plain old library function in Groovy, which requires no special support.

abayer:
In this case, the most notable thing I can think of off the top of my head is that I don't have to deal with the whitelist, so I don't need gobs of static methods in `Utils.groovy` with `@Whitelisted` on them.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecutionCompiler.java
kohsuke: Fixed

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecutionCompiler.java
kohsuke: Fixed

doc/step-in-groovy.md
kohsuke: Clarified.

kohsuke:
> having a @DataBoundConstructor parameter of type Closure

I haven't read the actual prototyped change by @abayer, but my understanding based on what he told me on chat is that it is an emergency workaround until I fix the way the body block gets passed into the `pipeline` step, so that he can scout ahead and discover other problems in parallel.

I am pretty sure it will not remain in the final shape. So no worries.

jglick:
> I don't have to deal with the whitelist, so I don't need gobs of static methods in `Utils.groovy` with `@Whitelisted` on them.

But that is also true of libraries, which run in a trusted class loader.

Now for your purposes you do not want a “library” per se, since that is something a user defines; rather you want a `GlobalVariable` I think, like `DockerDSL` defines. Currently that is only given a `CpsScript`, from which `script.getClass().getClassLoader()` is untrusted. Probably you could use `script.getClass().getClassLoader().getParent()` to access the trusted (library etc.) loader beneath it, though this seems fragile; I suspect that we need the API to expose the `CpsScript.execution` or at least its `getTrustedShell()` so that a `GlobalVariable` implementation could load trusted Groovy code. But really this is part of [JENKINS-32731](https://issues.jenkins-ci.org/browse/JENKINS-32731). I think JENKINS-37011 is being treated as a catch-all solution for problems which are better handled in other ways.

abayer:
@jglick @kohsuke Putting aside the experimental work over in `pipeline-model-definition`, I'm still deeply concerned about the fact that you can't use `docker` (or any other `GlobalVariable` that does class-loading) from within a `GroovyStep` or a body passed to/called by a `GlobalVariable`. That's a big issue, IMO.

kohsuke:
I'm running out of time on this side of the vacation. I've pushed all my results.

A part of the changes that need to happen to make pipeline-model-plugin really work well with this is to bring back the separate content root for loading `GroovyStepExecution` subtypes and other associated libraries that are (1) loaded & transformed in CPS, (2) run outside sandbox, (3) not visible from untrusted code.  This involves moving `GroovyStepExecution` subtype from `src/main/resources` to somewhere like `src/main/webapp/WEB-INF/*`, which Jesse didn't originally like, but I think it's unavoidable for plugin developers to be able to write a non-trivial amount of StepExecution code, like pipeline-model-plugin does or the Android emulator plugin does.

Just FYI, my sandbox to play with the integration of this with pipeline-model-plugin is [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/tree/step-in-groovy-kk)

jglick:
I still feel this is a bad choice for top-level stuff in `pipeline-model-definition` such as `pipeline {…}`, since these constructs will in general not follow the severely limited syntax permitted to `Step`s, nor do they need to be offered in _Snippet Generator_ for copy & paste into the middle of other `Jenkinsfile`s. Fine for things that would really be steps you might drop in the middle of a `stage`, or a non-declarative Pipeline script.

I am unfamiliar with the Android Emulator code just referred to; PR link?

> a non-trivial amount of StepExecution code

You should not _be_ writing a non-trivial amount of `[Groovy]StepExecution` code to begin with. The actual execution ought to be pretty short and to the point, basically running some binary `Step`s in some order, and calling into Java utility methods for any serious computation. If you want to create a non-steppy DSL, well you can already do that without this change.

Anyway it seems that there is no clear consensus yet on how this stuff is supposed to be used (or even what it is for), so best to hold off on merging.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ingroovy/GroovyStepExecution.java
```diff
@@ -0,0 +1,151 @@
...
+     * <p>
+     * This needs to happen inside CPS VM to work correctly, hence no caching.
+     *
+     * @return null
+     *      if this method is invoked (incorrectly) outside the CPS VM thread.
+     */
+    protected @Nullable GroovyObject getDelegate() throws IOException {
+        if (CpsThread.current()==null)
+            return null;    // invocation outside CPS VM thread
+        return new CpsScript() {
```
abayer: So this is definitely what's causing the classloader problems I mentioned. Since this `CpsScript` instance is being instantiated outside of the `GroovyShell`, it's not getting the right `ClassLoader`, and anything that's trying to use `script.getClass().getClassLoader()`, like [this example in `DockerDSL`](https://github.com/jenkinsci/docker-workflow-plugin/blob/master/src/main/java/org/jenkinsci/plugins/docker/workflow/DockerDSL.java#L47), will not work. 

I haven't been able to figure out how to solve this - there's no way I can find to get at the existing `CpsScript` instance at runtime to just pass that back onward, so...yeah. Confusion. Problems. And a :bug:. =)

jglick:
> Since this `CpsScript` instance is being instantiated outside of the `GroovyShell`, it's not getting the right `ClassLoader`

I think the `ClassLoader` is a bit of a red herring. The problem is that a single build can create multiple `CpsScript`s yet `GlobalVariable`s are passed whichever `CpsScript` is currently referring to the variable, which in your case is not the one which came from `Jenkinsfile`. Setting aside the `ClassLoader` issue here (which probably also affects global libraries—untested I think), in the worst case this can lead to incorrect behavior (if a `GlobalVariable` is stateful), and in the best case to suboptimal performance (loading the same stuff twice or more per build).

What we probably want is a better API for `GlobalVariable` which would let the implementation say that it wants to be truly global, i.e. one per build. That would mean that `CpsScript.getProperty`, rather than passing in `this`, would need to use `CpsThread.current().getGroup()` to look up the `CpsThreadGroup`, which is truly a singleton per build, and which could maintain a single `Binding` just for `GlobalVariable`s. See #29 for context. As noted in JENKINS-33353 this could also let `steps` be treated consistently, with only one `DSL` created per build.

jglick:
At first I thought it would suffice to fix the `CpsScript` to just be `CpsThreadGroup.scripts[0]`. But it turns out this change as written breaks the technique used by `DockerDSL` even *without* any Groovy steps: `GroovyStepExecutionCompiler` injects another `GroovyClassLoader` into the inheritance chain for every Pipeline script, so `Docker.groovy` gets loaded into that new intermediary `ClassLoader` (because we just use the `CpsScript`’s loader as the initiating loader and look for the “lowest” `ClassLoader` which claims to be able to load a `*.groovy` file), and so regular accesses to `docker` fail with a `RejectedAccessException`.

jglick:
@abayer I think I fixed the problem you identified, without needing any broader changes to `GlobalVariable` (though I still think those could be useful for other reasons). @kohsuke please check my changes.

Again this is inappropriate for anything that does not follow the usual design of a `Step`: something with a `config` view and straightforward data-bound parameters that could be run at any point in a script.

jglick:
> the technique used by `DockerDSL`

BTW filed https://github.com/jenkinsci/docker-workflow-plugin/pull/75 along the way.

jglick:
Specifically it was broken by https://github.com/apache/groovy/commit/09e9778e8a33052d8c27105aee5310649637233d in Groovy 2.

jglick:
Being used successfully in Ant plugin. Looking into whether [JENKINS-26635](https://issues.jenkins-ci.org/browse/JENKINS-26635) (external Groovy step) works as nicely.

abayer:
Bug revoked, :bee: - sorry for the delay!

jglick:
Now downstream of #83.

jglick:
https://github.com/cloudbees/groovy-cps/pull/40 was integrated; is there a corresponding change desired here?

jglick:
@orrc was [this](https://github.com/jenkinsci/android-emulator-plugin/compare/pipeline-dsl) intended as a use case?

kohsuke:
@jglick can we get this change finally merged?

jglick:
@kohsuke this is intentionally on hold, and I am having serious doubts about the wisdom of doing it at all, as it introduces new hard dependencies on `workflow-cps` which would complicate the introduction of possible alternate engines.

jglick:
I think this can be closed.

dwnusbaum:
Ok, I will go ahead and close this for now.

PR #60, opened 2016-09-17
[JENKINS-27295] [JENKINS-29952] Overhaul of access to build parameters

jglick:
Implements [JENKINS-27295](https://issues.jenkins-ci.org/browse/JENKINS-27295) and [JENKINS-29952](https://issues.jenkins-ci.org/browse/JENKINS-29952) and also demonstrates that [JENKINS-28447](https://issues.jenkins-ci.org/browse/JENKINS-28447) is fixed by the upstream https://github.com/jenkinsci/workflow-job-plugin/pull/24.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.3</version>
+            <version>2.7-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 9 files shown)


daniel-beck:
I assume everything else (symbols, explicitly defined vars, …) takes precedence over vars from `env`?

jglick:
> I assume everything else […] takes precedence over vars from env?

They should, yes, since if you get to the point of calling `getProperty` it means other bindings have already failed.

abayer:
Whoops, forgot I hadn't bee'd this - so, well, :bee:

PR #62, opened 2016-09-18
[JENKINS-34488] Save an alternate exception when failed to save ErrorAction

ikedam:
[JENKINS-34488](https://issues.jenkins-ci.org/browse/JENKINS-34488) Failing 'assert' hangs the pipeline

`hudson.util.XStream2` rejects to serialize some classes:
* https://github.com/jenkinsci/jenkins/blob/jenkins-2.22/core/src/main/java/hudson/util/XStream2.java#L442
* https://github.com/jenkinsci/remoting/blob/remoting-2.62/src/main/java/hudson/remoting/ClassFilter.java#L56

This results failure in saving `ErrorAction` when pipeline codes failed with some specific exceptions.
`assert` without a message parameter in groovy throws `org.codehaus.groovy.runtime.powerassert.PowerAssertionError`, and it will be rejected by `XStream2`.

This results skip later codes and causes to fail to finish codes.

This change is a workaround for this issue.
It saves `ErrorAction` with alternate exception when failed to save `ErrorAction`.

There're two points causing this problem:
* `CpsFlowExecution#onProgramEnd`
* `CpsThreadGroup#run`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -897,8 +897,20 @@ public class CpsFlowExecution extends FlowExecution {
         // TODO: if program terminates with exception, we need to record it
         // TODO: in the error case, we have to close all the open nodes
         FlowNode head = new FlowEndNode(this, iotaStr(), (FlowStartNode)startNodes.pop(), result, getCurrentHeads().toArray(new FlowNode[0]));
-        if (outcome.isFailure())
-            head.addAction(new ErrorAction(outcome.getAbnormal()));
+        if (outcome.isFailure()) {
+            try {
+                head.addAction(new ErrorAction(outcome.getAbnormal()));
+            } catch (Exception e) {
+                // considered caused for failed to serialize the error.
+                // retry without the original error.
+                LOGGER.log(Level.SEVERE, "Failed to save ErrorAction", e);
+                try {
+                    head.replaceAction(new ErrorAction(new ErrorActionException("Failed to add error action", e)));
+                } catch (Exception e2) {
+                    LOGGER.log(Level.SEVERE, "Failed to save ErrorAction again: Give up!", e2);
+                }
+            }
+        }
 
         // shrink everything into a single new head
         done = true;
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -332,7 +332,19 @@ public final class CpsThreadGroup implements Serializable {
                         result = ((FlowInterruptedException) error).getResult();
                     }
                     execution.setResult(result);
-                    t.head.get().addAction(new ErrorAction(error));
+
+                    try {
+                        t.head.get().addAction(new ErrorAction(error));
+                    } catch (Exception e) {
+                        // considered caused for failed to serialize the error.
+                        // retry without the original error.
+                        LOGGER.log(Level.SEVERE, "Failed to save ErrorAction", e);
+                        try {
+                            t.head.get().replaceAction(new ErrorAction(new ErrorActionException("Failed to add error action", e)));
+                        } catch (Exception e2) {
+                            LOGGER.log(Level.SEVERE, "Failed to save ErrorAction again: Give up!", e2);
+                        }
+                    }
                 }
 
                 if (!t.isAlive()) {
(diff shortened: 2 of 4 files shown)


ikedam:
It may makes sense to modify Jenkins core to allow serialize exceptions even they are in blacklisted packages (it depends on why remoting rejects serializing those classes).

jglick:
> it depends on why remoting rejects serializing those classes

`PowerAssertionError` could be safely removed from the blacklist, I think. But this would require a patch to Remoting. Easier to handle it here.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScriptTest.java
```diff
@@ -113,4 +116,33 @@ private String dumpError() {
...
+            ErrorAction e = n.getAction(ErrorAction.class);
+            if (e != null) {
+                return e;
+            }
+        }
+        return null;
+    }
+
+    @Issue("JENKINS-34488")
+    @WithTimeout(30)
```
jglick: Unnecessary—all tests time out after 3m by default.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScriptTest.java
```diff
@@ -113,4 +116,33 @@ private String dumpError() {
...
+        exec.start();
+        while (!exec.isComplete()) {
+            exec.waitForSuspension();
+        }
+        assertEquals(Result.FAILURE, exec.getResult());
+
+        // extract ErrorAction.
+        ErrorAction e = extractFirstErrorAction();
+        assertNotNull(e);
+        assertThat(e.getError(), Matchers.instanceOf(ErrorActionException.class));
```
jglick: Easier and more general to write this in pretty much any other test suite, using actual `WorkflowJob`s rather than `AbstractCpsFlowTest`, since you can just `assertLogContains` with whatever text you would want the build to print had this bug not existed (`Assertion failed`?).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -332,7 +332,19 @@ private boolean run() throws IOException {
                         result = ((FlowInterruptedException) error).getResult();
                     }
                     execution.setResult(result);
-                    t.head.get().addAction(new ErrorAction(error));
+
+                    try {
+                        t.head.get().addAction(new ErrorAction(error));
+                    } catch (Exception e) {
+                        // considered caused for failed to serialize the error.
```
jglick: Better to consolidate duplicated code into one `ErrorAction.addTo(FlowNode)` utility method.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -897,8 +897,20 @@ public boolean isComplete() {
         // TODO: if program terminates with exception, we need to record it
         // TODO: in the error case, we have to close all the open nodes
         FlowNode head = new FlowEndNode(this, iotaStr(), (FlowStartNode)startNodes.pop(), result, getCurrentHeads().toArray(new FlowNode[0]));
-        if (outcome.isFailure())
-            head.addAction(new ErrorAction(outcome.getAbnormal()));
+        if (outcome.isFailure()) {
+            try {
+                head.addAction(new ErrorAction(outcome.getAbnormal()));
+            } catch (Exception e) {
```
jglick: Possibly useful for corner cases where saving the `FlowNode` genuinely fails for legitimate reasons, like disk I/O problems, though I am not sure we can do much to recover in those cases anyway.

The diagnosis in this issue looks right, but the patch is much too much work, and actually loses the information in the original exception. All you needed to do was add another blacklist to `ErrorAction.isUnserializableException`. Given the Remoting blacklist pattern, and possible future changes to that, perhaps we should just say

```java
private boolean isUnserializableException(Throwable error) {
    try {
        ClassFilter.DEFAULT.check(error.getClass());
    } catch (SecurityException x) {
        return true; // e.g. PowerAssertionError
    }
    return error instanceof MultipleCompilationErrorsException ||
           error instanceof MissingMethodException;
}
```

jglick: > I am not sure we can do much to recover in those cases

Then again, the originally reported problem was less about the assertion failing to be printed, than the whole build hanging, so something does need to be fixed here just in case something else does go terribly wrong.

But I suspect `CpsThreadGroup` is not the best place to do this. Rather, [here](https://github.com/jenkinsci/workflow-api-plugin/blob/8ca82ae8180dc0869f083f2684c291e2faa18222/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java#L301-L305) we are catching `IOException` but not the `RuntimeException`s that are common with XStream. So this whole PR could probably be closed and replaced with one to `workflow-api` which both patches `ErrorAction` to properly save blacklisted power assertions (with full details), *and* handles any other unforeseen problems adding `ErrorAction` or anything else to a node, and also includes a functional test reproducing the bug.

ikedam:
That sounds a better approach.
I'll create a new pull request for workflow-api-plugin.
Thanks!

PR #64, opened 2016-09-20
Detect failure in CPS VM

kohsuke:
I was looking at a problem where a test seemingly hangs forever. I tracked it down to a problem in CPS VM,
but what made this difficult for me is that a RuntimeException happening in some unexpected part of CPS VM went unreported.
If this error happened in the field, I would have never been able to figure this out. So I'm adding a diagnostics here to make sure
we can catch this class of problems.

Just for the record, the particular problem in CPS VM is a NullPointerException in the following place:

      at org.jenkinsci.plugins.workflow.cps.CpsBodyExecution$FailureAdapter.receive(CpsBodyExecution.java:283)
      at org.jenkinsci.plugins.workflow.cps.CpsBodyExecution.launch(CpsBodyExecution.java:161)
      at org.jenkinsci.plugins.workflow.cps.CpsBodyInvoker.launch(CpsBodyInvoker.java:189)
      at org.jenkinsci.plugins.workflow.cps.DSL$ThreadTaskImpl.invokeBody(DSL.java:501)
      at org.jenkinsci.plugins.workflow.cps.DSL$ThreadTaskImpl.eval(DSL.java:471)
      at org.jenkinsci.plugins.workflow.cps.CpsThread.runNextChunk(CpsThread.java:177)
      at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:324)
      at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.access$100(CpsThreadGroup.java:78)
      at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:236)
      at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:224)
      at org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$2.call(CpsVmExecutorService.java:47)
      at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:334)
      at java.util.concurrent.FutureTask.run(FutureTask.java:166)
      at hudson.remoting.SingleLaneExecutorService$1.run(SingleLaneExecutorService.java:112)
      at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
      at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:334)
      at java.util.concurrent.FutureTask.run(FutureTask.java:166)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
      at java.lang.Thread.run(Thread.java:722)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -91,7 +91,7 @@ class CpsVmExecutorService extends InterceptingExecutorService {
         context.restore();
     }
 
-    static ThreadLocal<CpsThreadGroup> CURRENT = new ThreadLocal<CpsThreadGroup>();
+    static ThreadLocal<CpsThreadGroup> CURRENT = new ThreadLocal<>();
     /** {@link Thread#getContextClassLoader} to be used for plugin code, as opposed to Groovy. */
     static ThreadLocal<ClassLoader> ORIGINAL_CONTEXT_CLASS_LOADER = new ThreadLocal<>();
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -7,6 +7,9 @@ import jenkins.util.InterceptingExecutorService;
 
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
+import java.util.logging.Logger;
+
+import static java.util.logging.Level.WARNING;
 
 /**
  * {@link ExecutorService} for running CPS VM.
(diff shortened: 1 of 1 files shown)


abayer:
:bee: Always gotta like better logging. =)

oleg-nenashev:
🐝 even if handling of Throwable may cause another Error and suppress the original one (E.g. in the case of OutOfMemory)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -30,13 +33,26 @@ public void run() {
...
     }
 
+    /**
+     * {@link CpsVmExecutorService} is used to run CPS VM asynchronously by one thread.
+     * None of the submissions of these tasks are monitoring their outcome, and an exception
+     * here usually means something catastrophic to the state of CPS VM.
+     * That makes it worth reporting.
+     */
+    private void reportProblem(Throwable t) {
+        LOGGER.log(WARNING, "Unexpected exception in CPS VM thread: "+cpsThreadGroup.getExecution().toString(), t);
```
jglick: Gratuitous `toString`.

jglick:
retro:bee: assuming you actually verified that this patch *does* print the error. I am curious why it would not be printed before: `CpsVmExecutorService` uses `Computer.threadPoolForRemoting`, which installs an `ExceptionCatchingThreadFactory`. So which part of that failed to report the exception?

PR #65, opened 2016-09-20
Incorrect handling of synchronous abnormal termination of the body block

kohsuke:
Fixing a bug in CPS VM about how it handles an exception in sync body.

When the body of a step is synchronous and explodes, the failure should be recorded and the pipeline job should move on.

But instead, this hangs because `CpsBodyExecution` has a bug in how it handles this case. It tries to launch the body  in a separate CPS thread, and puts the parent CPS thread on hold. Yet when the child CPS thread ends with an exception, it fails to record this result correctly, and it gets into the eternal sleep in which the parent CPS thread expects to be notified of the outcome of the child CPS thread, which never arrives.

Any exception aside from `CpsCallableInvocation` is a sign that the execution has completed abnormally and synchronously. So this code pass should have close parallel to the normal case that immediately follows the `params.body.getBody(currentThread).call()` invocation, yet the thread association is missing from the failure case.

`CpsBodyExecution.onSuccess/onFailure` also assumes that they run in CPS VM thread to be able to grow the flow node graph, as can be seen in their use of `CpsThread.current()`. In the typical case where the body being invoked is also CPS code, this assumption correctly holds as these callbacks as invoked as Continuation while running CPS thread code, but if it executes synchronously, this assumption does not hold, as the call stack originates in `ThreadTask.eval(this)` that's outside the code that sets the `CpsThread.current()` return value.

`ThreadTask.eval()` always operate in the context of single `CpsThread`, it seems reasonable to run this code whiel `CpsThread.current()` is still set.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -139,6 +139,9 @@ class CpsBodyExecution extends BodyExecution {
             }
         } catch (Throwable t) {
             // body has completed synchronously and abnormally
+            synchronized (this) {
+                this.thread = currentThread;
+            }
             onFailure.receive(t);
         }
     }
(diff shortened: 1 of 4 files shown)


abayer:
:bee:

jglick:
retro:bee: (test looks fine, do not understand anything about problem or fix)

PR #66, opened 2016-09-20
When StepExecution.stop throws an exception, forcibly kill the step and record both stack traces

jglick:
@cyrille-leclerc [encountered a case](https://gist.githubusercontent.com/jglick/b906ae79a0dd857dae0098453fe03612/raw/fe90a36ee9311e35f0ad451c4d4e823642486da2/gistfile1.txt) where `StepExecution.stop` was called but threw an exception; the build kept running. Normally `DurableTaskStep.Execution.stop` first checks whether the remoting channel is alive and well, but in this case for some reason that check passed yet an actual attempt to use the channel to run a termination command failed. Best to abandon the step as a failure at that point.

Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/8 for nicer reporting.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,9 +68,9 @@
     </properties>
     <dependencies>
         <dependency>
-            <groupId>${project.groupId}</groupId>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.3</version>
+            <version>2.4-SNAPSHOT</version> <!-- TODO https://github.com/jenkinsci/workflow-step-api-plugin/pull/8 -->
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -68,9 +68,9 @@
     </properties>
     <dependencies>
         <dependency>
-            <groupId>${project.groupId}</groupId>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
```
oleg-nenashev: Will require https://github.com/jenkins-infra/repository-permissions-updater patch

jglick: Huh? Just replacing an expression with the value it would expand to.

oleg-nenashev: then NVM. Just thought the parent had another grou id

PR #67, opened 2016-09-21
Simpler workaround for map iteration JENKINS-27421

jglick:
`ArrayList` iteration works, so might as well use that.

@reviewbybees


abayer:
:bee: assuming tests pass. I'm tempted to add utility methods like this to the DSL somehow, but that's probably a bad idea to bake into the plugin...

jglick:
> I'm tempted to add utility methods like this to the DSL somehow

Throw it up on GitHub and make a library out of it…

jglick:
> assuming tests pass

Currently this one is `@Ignore`d anyway, since its main point was to test the _desired_ direct iteration of `Map.entrySet()`. But yes if you uncomment the annotation and run it, the workaround part works as expected, then it fails in the expected part.

PR #68, opened 2016-09-21
Some flaky tests

jglick:
Trying to solve/diagnose some test flakes noticed in recent release tests.

```
	at org.jenkinsci.plugins.workflow.SerializationTest$1.evaluate(SerializationTest.java:52)
java.lang.InterruptedException: sleep interrupted
	at java.lang.Thread.sleep(Native Method)
	at org.jvnet.hudson.test.JenkinsRule.waitForCompletion(JenkinsRule.java:1231)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecutionTest$7.evaluate(CpsFlowExecutionTest.java:262)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadDumpTest.simple(CpsThreadDumpTest.java:47)
```

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
@@ -49,8 +49,10 @@ public class SerializationTest extends SingleJobTestBase {
                         waitForWorkflowToSuspend();
                     } catch (Exception x) {
                         // ignore persistence failure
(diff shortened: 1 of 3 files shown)


jglick:
The usual “forked VM crashed without saying goodbye” error; not sure how to diagnose these.

svanoort:
🐝

PR #71, opened 2016-09-27
Delegate preferentially to global bindings before applying custom getProperty behaviors

jglick:
Someone on #jenkins claimed to have been using some idiom along the lines of

```groovy
if (SOME_BUILD_PARAM_NAME == '') {
  SOME_BUILD_PARAM_NAME = 'fallback'
}
useSomehow(SOME_BUILD_PARAM_NAME)
```

which I had never intended to support but apparently worked prior to #60 ([release notes](https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Groovy+Plugin#PipelineGroovyPlugin-2.18%28Sep23%2C2016%29)). It seems that the `GlobalVariable` and `env` fallbacks were taking precedence over `Binding`s, which is probably unwanted—generally we want to let scripts override symbols. This should restore the previous behavior in that case.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
@@ -26,10 +26,8 @@ package org.jenkinsci.plugins.workflow.cps;
 
 import com.cloudbees.groovy.cps.SerializableScript;
 import groovy.lang.GroovyShell;
+import groovy.lang.MissingPropertyException;
 import groovy.lang.Script;
-import hudson.EnvVars;
-import hudson.model.ParameterValue;
-import hudson.model.ParametersAction;
 import hudson.model.Queue;
 import hudson.model.Run;
 import java.io.File;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
```diff
@@ -111,27 +110,31 @@ public final Object invokeMethod(String name, Object args) {
...
     @Override
     public Object getProperty(String property) {
-        // cf. CpsWhitelist.permitsMethod
-        Run<?,?> b = $buildNoException();
-        GlobalVariable v = GlobalVariable.byName(property, b);
-        if (v != null) {
-            try {
-                return v.getValue(this);
-            } catch (Exception x) {
-                throw new InvokerInvocationException(x);
```
jglick: [whitespace-reduced diff](https://github.com/jenkinsci/workflow-cps-plugin/pull/71/files?w=1#diff-f4e7d68744ec21328085d86a13da2d4eL111)

abayer:
:bee:

PR #73, opened 2016-10-10
[JENKINS-35698] Make params offer default parameter values from the current definition as a fallback

jglick:
[JENKINS-35698](https://issues.jenkins-ci.org/browse/JENKINS-35698)

@reviewbybees esp. @abayer


abayer:
:bee:

PR #74, opened 2016-10-17
Placate the JDK 9 compiler

jglick:
[JDK-8075793](https://bugs.openjdk.java.net/browse/JDK-8075793)

@reviewbybees


oleg-nenashev:
:bee:

PR #75, opened 2016-10-18
compatibleSinceVersion=2.18

jglick:
IIUC this will ensure that anyone upgrading from [2.17 or earlier](https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Groovy+Plugin) is asked to read release notes according to [this system](https://wiki.jenkins-ci.org/display/JENKINS/Marking+a+new+plugin+version+as+incompatible+with+older+versions).

@reviewbybees esp. @daniel-beck who suggested it

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,13 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.jenkins-ci.tools</groupId>
+                <artifactId>maven-hpi-plugin</artifactId>
+                <configuration>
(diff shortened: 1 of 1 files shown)


daniel-beck:
The message will be a bit misleading, but it's the best we can do :+1:

HRMPW:
Thanks for the change, @jglick . 👍

oleg-nenashev:
🐝

daniel-beck:
Ideally this'll be a .1 release, if nothing else is release-ready, to get the message out ASAP.

PR #76, opened 2016-10-18
[JENKINS-34637] Failure to kill bodies from timeout

jglick:
[JENKINS-34637](https://issues.jenkins-ci.org/browse/JENKINS-34637)

Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/20.

@reviewbybees esp. @kohsuke

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -10,6 +10,7 @@ import com.cloudbees.groovy.cps.impl.TryBlockEnv;
 import com.cloudbees.groovy.cps.sandbox.SandboxInvoker;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.SettableFuture;
 import hudson.model.Action;
 import hudson.model.Result;
 import hudson.util.Iterators;
@@ -180,30 +181,53 @@ class CpsBodyExecution extends BodyExecution {
 
     @Override
     public Collection<StepExecution> getCurrentExecutions() {
-        if (thread==null)   return Collections.emptyList();
-        List<StepExecution> executions = new ArrayList<>();
-        // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
-        Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
-        // TODO access to CpsThreadGroup.threads should be restricted to the CPS VM thread, but the API signature does not allow us to return a promise or throw InterruptedException
-        for (CpsThread t : thread.group.threads.values()) {
-            m.put(t.head, t);
-        }
-        for (CpsThread t : m.values()) {
-            // TODO seems cumbersome to have to go through the flow graph to find out whether a head is a descendant of ours, yet FlowHead does not seem to retain a parent field
-            LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
-            scanner.setup(t.head.get());
-            for (FlowNode node : scanner) {
-                if (node.getId().equals(startNodeId)) {
-                    // this head is inside this body execution
-                    StepExecution execution = t.getStep();
-                    if (execution != null) {
-                        executions.add(execution);
-                    }
-                    break;
-                }
+        CpsThread t;
+        synchronized (this) {
+            t = thread;
+            if (t == null) {
+                return Collections.emptySet();
             }
         }
-        return executions;
+        final SettableFuture<Collection<StepExecution>> result = SettableFuture.create();
+        t.getExecution().runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
+            @Override public void onSuccess(CpsThreadGroup g) {
+                try {
+                    List<StepExecution> executions = new ArrayList<>();
+                    // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
+                    Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
+                    for (CpsThread t : g.threads.values()) {
+                        m.put(t.head, t);
+                    }
+                    for (CpsThread t : m.values()) {
+                        // TODO seems cumbersome to have to go through the flow graph to find out whether a head is a descendant of ours, yet FlowHead does not seem to retain a parent field
+                        LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
+                        scanner.setup(t.head.get());
+                        for (FlowNode node : scanner) {
+                            if (node.getId().equals(startNodeId)) {
+                                // this head is inside this body execution
+                                StepExecution execution = t.getStep();
+                                if (execution != null) {
+                                    executions.add(execution);
+                                }
+                                break;
+                            }
+                        }
+                    }
+                    result.set(executions);
+                } catch (Exception x) {
+                    result.setException(x);
+                }
+            }
+            @Override public void onFailure(Throwable t) {
+                result.setException(t);
+            }
+        });
+        try {
+            return result.get(1, TimeUnit.MINUTES);
+        } catch (ExecutionException | InterruptedException | TimeoutException x) {
+            // TODO access to CpsThreadGroup.threads must be restricted to the CPS VM thread, but the API signature does not allow us to return a ListenableFuture or throw checked exceptions
+            throw new RuntimeException(x);
+        }
     }
 
     @Override
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -8,9 +8,11 @@ import com.cloudbees.groovy.cps.impl.CpsCallableInvocation;
 import com.cloudbees.groovy.cps.impl.FunctionCallEnv;
 import com.cloudbees.groovy.cps.impl.TryBlockEnv;
 import com.cloudbees.groovy.cps.sandbox.SandboxInvoker;
+import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.FutureCallback;
 import hudson.model.Action;
 import hudson.model.Result;
+import hudson.util.Iterators;
 import jenkins.model.CauseOfInterruption;
 import org.jenkinsci.plugins.workflow.actions.BodyInvocationAction;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
@@ -25,9 +27,12 @@ import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
 import javax.annotation.concurrent.GuardedBy;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -36,6 +41,8 @@ import java.util.logging.Logger;
 import static java.util.logging.Level.*;
 import javax.annotation.Nonnull;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.graphanalysis.LinearBlockHoppingScanner;
 
 /**
  * {@link BodyExecution} impl for CPS.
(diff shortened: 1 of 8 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +179,63 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
     }
 
     @Override
-    public synchronized Collection<StepExecution> getCurrentExecutions() {
+    public Collection<StepExecution> getCurrentExecutions() {
```
jglick: This method was not involved in the original bug, but it was wrong too, in the same way.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +179,63 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
...
+                    for (CpsThread t : thread.group.threads.values()) {
+                        m.put(t.head, t);
+                    }
+                    for (CpsThread t : Iterators.reverse(ImmutableList.copyOf(m.values()))) {
+                        LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
+                        scanner.setup(t.head.get());
+                        for (FlowNode node : scanner) {
+                            if (node.getId().equals(startNodeId)) {
+                                t.stop(stopped);
+                                break;
```
jglick: This is the main fix.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -61,6 +61,10 @@
     private final String script;
     private final boolean sandbox;
 
+    /**
+     * @deprecated use {@link #CpsFlowDefinition(String, boolean)} instead
+     */
+    @Deprecated
```
jglick: Not related, just on my to-do list for ages…

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -321,23 +324,46 @@ public CpsBodyInvoker newBodyInvoker() {
...
+            outcome = newOutcome;
+            scheduleNextRun();
+            whenOutcomeDelivered = new Throwable();
+        } else {
+            Throwable failure = newOutcome.getAbnormal();
+            if (failure instanceof FlowInterruptedException) {
+                for (CauseOfInterruption cause : ((FlowInterruptedException) failure).getCauses()) {
+                    if (cause instanceof BodyFailed) {
+                        LOGGER.log(Level.FINE, "already completed " + this + " and now received body failure", failure);
+                        // Predictable that the error would be thrown up here; quietly ignore it.
```
jglick: Otherwise we get an ugly stack trace when timing out.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -231,7 +231,7 @@ void fireCompletionHandlers(Outcome o) {
     /**
      * Finds the next younger {@link CpsThread} that shares the same {@link FlowHead}.
      *
-     * Can be {@code this.}
+     * Cannot be {@code this}.
```
jglick: Did not wind up using this function; another comment, at its sole call site, notes that it is probably wrong to use it there, too (but leaving that for another day).

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -126,7 +126,10 @@ public void onFailure(StepContext context, Throwable t) {
                 if (handler.originalFailure == null) {
                     handler.originalFailure = new SimpleEntry<String, Throwable>(name, t);
                 } else {
-                    handler.originalFailure.getValue().addSuppressed(t);
+                    Throwable originalT = handler.originalFailure.getValue();
+                    if (t != originalT) { // could be the same abort being delivered across branches
```
jglick: Otherwise blows up when using `parallel` inside `timeout`.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -121,4 +134,64 @@ public void stop(@Nonnull Throwable cause) throws Exception {
...
+            }
+        }).get();
+        assertThat(semaphores, Matchers.<SemaphoreStep.Execution>iterableWithSize(2));
+        Collection<StepExecution> currentExecutions1 = execs[1].body.getCurrentExecutions(); // A or B, does not matter
+        assertThat(/* irritatingly, iterableWithSize does not show the collection in its mismatch message */currentExecutions1.toString(),
+            currentExecutions1, Matchers.<StepExecution>iterableWithSize(1));
+        Collection<StepExecution> currentExecutions2 = execs[2].body.getCurrentExecutions();
+        assertThat(currentExecutions2, Matchers.<StepExecution>iterableWithSize(1));
+        assertEquals(semaphores, Sets.union(Sets.newLinkedHashSet(currentExecutions1), Sets.newLinkedHashSet(currentExecutions2)));
+        assertEquals(semaphores, Sets.newLinkedHashSet(execs[0].body.getCurrentExecutions())); // the top-level one
```
jglick: Failed before fix: returned a singleton set of `ParallelStepExecution`.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -121,4 +134,64 @@ public void stop(@Nonnull Throwable cause) throws Exception {
...
+        execs[0].body.cancel();
+        SemaphoreStep.success("c/1", null);
+        jenkins.assertBuildStatus(Result.ABORTED, jenkins.waitForCompletion(b));
```
jglick: Failed before fix: called `ParallelStepExecution.stop`, which delegated to `RetainsBodyStep.Execution.stop`, rather than going straight to `SemaphoreStep.Execution.stop`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +179,63 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
...
         if (thread==null)   return Collections.emptyList();
-
-        StepExecution s = thread.getStep();
-        if (s!=null)        return Collections.singleton(s);
-        else                return Collections.emptyList();
+        List<StepExecution> executions = new ArrayList<>();
+        // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
+        Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
+        // TODO access to CpsThreadGroup.threads should be restricted to the CPS VM thread, but the API signature does not allow us to return a promise or throw InterruptedException
+        for (CpsThread t : thread.group.threads.values()) {
```
rsandell: don't you need to guard against concurrent modifications?

rsandell:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: Well, yes—see the `TODO` comment right above. Unfortunately when @kohsuke defined this API method, besides implementing it incorrectly here, he did not allow it to throw any exceptions—and to implement it correctly (to run in the CPS VM thread) we would need to either throw exceptions, or return a `ListenableFuture`.

At any rate it appears this method was never actually called, so I am not sure it matters. Perhaps the API method should be deprecated and a better replacement introduced if and when there is a need.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +179,63 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
...
         }
 
         if (t!=null) {
             t.getExecution().runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
                 @Override
                 public void onSuccess(CpsThreadGroup g) {
-                    t.stop(stopped);
+                    // Similar to getCurrentExecutions but we want the raw CpsThread, not a StepExecution; cf. CpsFlowExecution.interrupt
+                    Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
+                    for (CpsThread t : thread.group.threads.values()) {
```
jglick: Not here, because we are in the single CPS VM thread.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +180,86 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
...
+                try {
+                    List<StepExecution> executions = new ArrayList<>();
+                    // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
+                    Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
+                    for (CpsThread t : g.threads.values()) {
+                        m.put(t.head, t);
+                    }
+                    for (CpsThread t : m.values()) {
+                        // TODO seems cumbersome to have to go through the flow graph to find out whether a head is a descendant of ours, yet FlowHead does not seem to retain a parent field
+                        LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
```
svanoort: You can simply this with FlowScanningUtils.fetchEnclosingBlocks -- it will just take the t.head.get() and return an iterator over enclosing BlockStartNodes.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +180,86 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
...
+            @Override public void onSuccess(CpsThreadGroup g) {
+                try {
+                    List<StepExecution> executions = new ArrayList<>();
+                    // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
+                    Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
+                    for (CpsThread t : g.threads.values()) {
+                        m.put(t.head, t);
+                    }
+                    for (CpsThread t : m.values()) {
+                        // TODO seems cumbersome to have to go through the flow graph to find out whether a head is a descendant of ours, yet FlowHead does not seem to retain a parent field
```
svanoort: Consideration: would it be worth adding an enclosingBlockStartNodeId field to FlowNode, and setting it with our block-generating steps?

More data to persist and read, but simplifies many things.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -172,29 +180,86 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
...
                 @Override
                 public void onSuccess(CpsThreadGroup g) {
-                    t.stop(stopped);
+                    // Similar to getCurrentExecutions but we want the raw CpsThread, not a StepExecution; cf. CpsFlowExecution.interrupt
+                    Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
+                    for (CpsThread t : thread.group.threads.values()) {
+                        m.put(t.head, t);
+                    }
+                    for (CpsThread t : Iterators.reverse(ImmutableList.copyOf(m.values()))) {
+                        LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
```
svanoort: Same simplification as before applies (FlowScanner.getEnclosingBlocks)

svanoort: Actually, scratch that: reuse the scanner though.  You can create an instance and do setup on it repeatedly since it is stateful.

jglick:
`CpsThreadDump` test failures are trivial, will fix. Need to look at the `ParallelStepTest.suspend` failure closely.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: > adding an `enclosingBlockStartNodeId` field to `FlowNode`

Well it could be `transient` since it recomputable based on existing information. Anyway that would not help here I think, because we are starting with thread information, which does not completely align with flow graph information.

(For `getCurrentExecutions`—which no one besides my new test ever calls—it would probably suffice: you could filter `FlowExecution.currentHeads` to those enclosed by this body. But for `cancel` we need the actual `CpsThread`s, since these have special behavior for `stop` even when not inside a `Step`.)

svanoort:
🐝  AIUI with a suggested small improvement.  Smells potentially risky though due to complexity.   Would suggest deploying on an instance with a few job types to test (I've got one I can toss it on if you don't)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: So what I would really want _here_ is a way to get from a `FlowHead` to an `Iterable<CpsThread>` representing the parts of the call stack that interact with step executions, or conversely to go from a `CpsThread` to any descendants (`getNextInner` does not suffice). For example in

```groovy
timeout(…) {/* #1 */
  parallel a: {
    sleep 9999 /* #2 */
  }, b: {
    while (true) {/* #3 */}
  }
}
```

we would have a straightforward way of going from the `CpsBodyExecution` marked at point 1 to the `CpsThread`s representing points 2 and 3.

svanoort: It simplifies the LinearBlockHoppingScanner use anyway (and might make it mostly redundant), but the catch is you have to have provided it at execution time or do a full scan of the flow to generate the data (persisting as you go). 

Meh, there are bigger optimizations.  Flow scanning will get faster and faster over time anyway.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: I saw that utility method, but it does not simplify much, I think:

```java
for (FlowNode node : FlowScanningUtils.fetchEnclosingBlocks(t.head.get())) {
    if (node.getId().equals(startNodeId)) {
        // …as before
    }
}
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: I could. This code is not performance-critical however.

svanoort: Yeah, it's a micro-optimization here.  Stuff that's on the hot path should do that though wherever possible.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
svanoort: It's a little more terse and does a prefilter to only return block start nodes by instanceof check (maybe slightly faster).  Not a big deal though, there's a half dozen ways to do this, I just want to put it on your radar since this is exactly what it's for.

jglick:
> potentially risky

Well, the changes should only affect calls to `BodyExecution.cancel`…which we already know was badly broken. So I am comfortable with the functional test coverage here. Was hoping for some guidance from @kohsuke on his original intention w.r.t. `StepExecution.stop`, but my understanding of his more recent code in `CpsFlowExecution.interrupt` is that it is expected that we deliver interrupts directly to the innermost running code—which may then propagate throwables up the call stack, log them and continue, process `finally` blocks, etc.

PR #80, opened 2016-10-28
Update Snippet Generator

HRMPW:
Replacing occurrences of "groovy" with "Pipeline Script"

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -31,9 +31,9 @@ THE SOFTWARE.
             <f:form name="config">
                 <f:section title="Overview"/>
                 <f:block>
-                    This <b>Snippet Generator</b> will help you learn the Groovy code which can be used to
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -31,9 +31,9 @@ THE SOFTWARE.
             <f:form name="config">
                 <f:section title="Overview"/>
                 <f:block>
-                    This <b>Snippet Generator</b> will help you learn the Groovy code which can be used to
+                    This <b>Snippet Generator</b> will help you learn the Pipeline Script code which can be used to
                     define various steps. Pick a step you are interested in from the list, configure it,
-                    click <b>Generate Groovy</b>, and you will see a Groovy statement that would call the
+                    click <b>Generate Pipeline Script</b>, and you will see a Pipeline Script statement that would call the
```
jglick: I think it is still useful to say here that it is generating a Groovy statement. But I do not have a strong opinion.

christ66:
🐝

PR #82, opened 2016-11-02
Intern or deduplicate many duplicated strings

svanoort:
Should reduce memory footprint for Jenkins pipeline a bit and simplify garbage collection some. 

See [JENKINS-39456](https://issues.jenkins-ci.org/browse/JENKINS-39456)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -263,7 +263,7 @@ public class CpsFlowExecution extends FlowExecution {
 
     static {
         for(int i = 0; i< ID_LOOKUP_TABLE.length; i++) {
-            ID_LOOKUP_TABLE[i] = String.valueOf(i);
+            ID_LOOKUP_TABLE[i] = String.valueOf(i).intern();  // Interning here allows allows us to just intern on deserialize
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -255,6 +255,18 @@ public class CpsFlowExecution extends FlowExecution {
 
     private final AtomicInteger iota = new AtomicInteger();
 
+    /** Number of node IDs to use in lookup table, 500 covers most common flow graphs  */
+    private static final int ID_LOOKUP_TABLE_SIZE = 500;
+
+    /** Preallocated lookup table for small ID values, used instead of interning for speed & simplicity */
+    private static transient String[] ID_LOOKUP_TABLE = new String[ID_LOOKUP_TABLE_SIZE];
+
+    static {
+        for(int i = 0; i< ID_LOOKUP_TABLE.length; i++) {
+            ID_LOOKUP_TABLE[i] = String.valueOf(i);
+        }
+    }
+
     private transient List<GraphListener> listeners;
 
     /**
@@ -429,7 +441,13 @@ public class CpsFlowExecution extends FlowExecution {
      */
     @Restricted(NoExternalUse.class)
     public String iotaStr() {
-        return String.valueOf(iota());
+        int iotaVal = iota();
+        // We intern this because many, many FlowNodes will have the same ID values
+        if ( iotaVal > 0 && iotaVal < ID_LOOKUP_TABLE_SIZE) {
+            return ID_LOOKUP_TABLE[iotaVal];
+        } else {
+            return String.valueOf(iota()).intern();
+        }
     }
 
     @Restricted(NoExternalUse.class)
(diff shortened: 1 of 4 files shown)


svanoort:
Waiting to see if CI passes, if so, will put up for review.

svanoort:
@reviewbybees esp @jglick

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -255,6 +255,18 @@
 
     private final AtomicInteger iota = new AtomicInteger();
 
+    /** Number of node IDs to use in lookup table, 500 covers most common flow graphs  */
+    private static final int ID_LOOKUP_TABLE_SIZE = 500;
+
+    /** Preallocated lookup table for small ID values, used instead of interning for speed & simplicity */
+    private static transient String[] ID_LOOKUP_TABLE = new String[ID_LOOKUP_TABLE_SIZE];
```
jglick: `static transient` makes no sense. I think you meant `static final`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -255,6 +255,18 @@
 
     private final AtomicInteger iota = new AtomicInteger();
 
+    /** Number of node IDs to use in lookup table, 500 covers most common flow graphs  */
+    private static final int ID_LOOKUP_TABLE_SIZE = 500;
+
+    /** Preallocated lookup table for small ID values, used instead of interning for speed & simplicity */
```
jglick: Probably overkill. A direct call to `String.intern` is going to be so much faster than all the crap we are doing when creating a new `FlowNode` that using a lookup table is just gratuitous code.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepAtomNode.java
```diff
@@ -48,7 +48,7 @@
 
     public StepAtomNode(CpsFlowExecution exec, StepDescriptor d, FlowNode parent) {
         super(exec, exec.iotaStr(), parent);
-        this.descriptorId = d!=null ? d.getId() : null;
+        this.descriptorId = d!=null ? d.getId().intern() : null;
```
jglick: Gratuitous. `Descriptor.getId()` normally calls `Class.getName()`, which uses a cached value; and `Descriptor`s are singletons, so you are not going to have multiple copied of a given `id` to begin with.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepStartNode.java
```diff
@@ -24,7 +24,7 @@
     public StepStartNode(CpsFlowExecution exec, StepDescriptor d, FlowNode parent) {
         super(exec, exec.iotaStr(), parent);
         this.descriptor = d;
-        this.descriptorId = d!=null ? d.getId() : null;
+        this.descriptorId = d!=null ? d.getId().intern() : null;
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepAtomNode.java
svanoort: I'm pretty sure that is *not* being cached, because I definitely see dupes in memory when looking with YourKit and they're gone after the change

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: Transient is redundant, more accurately

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: 1 array lookup vs. hashing string and doing equals against entry in the intern table -- should be exponentially faster, given the array will live almost entirely in CPU cache.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepAtomNode.java
jglick: Well track down why.

svanoort:
Kicking to WIP until I see if readResolve works right

svanoort:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/23

svanoort:
@reviewbybees especially @jglick  -- if tests fail, it's probably having issues with interdependencies b/w PRs

abayer:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: We're creating a ton of garbage strings that must be immediately GC'd if we do that (unless escape analysis catches them, but that's unreliable) -- in any case, valueOf is more expensive than one would think and so is String.intern( ). 

If it was called once per build instead of hundreds of times, I'd agree with you, or if it added any meaningful level of extra complexity.  As is, there's no point in pessimizing it.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepAtomNode.java
jglick: That would be expected during _deserialization_, which you are already fixing elsewhere; but why here? Something is wrong if `Descriptor.getId()` varies from call to call.

rsandell:
:bee:

svanoort:
@jglick I tested in manually with historical data from many builds.

PR #83, opened 2016-11-02
More memory leaks

jglick:
Cf. #41.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -94,6 +94,7 @@ import java.util.logging.Logger;
 
 import static com.thoughtworks.xstream.io.ExtendedHierarchicalStreamWriterHelper.*;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import groovy.lang.GroovyClassLoader;
 import groovy.lang.GroovyCodeSource;
 import hudson.AbortException;
 import hudson.BulkChange;
@@ -109,7 +110,9 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
@@ -593,7 +596,7 @@ public class CpsFlowExecution extends FlowExecution {
             @Override public void onFailure(Throwable t) {
                 LOGGER.log(Level.WARNING, "Failed to set program failure on " + owner, t);
                 onProgramEnd(new Outcome(null, t));
-                cleanUpHeap(Collections.singletonList(scriptClass));
+                cleanUpHeap();
             }
         });
     }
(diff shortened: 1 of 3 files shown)


abayer:
:bee:

jglick:
I also want to check behavior with shared libraries (internal ~ `GroovyShellDecoratorImpl` as well as external ~ `ClasspathAdder`); and #59 ought to augment this test to check that Groovy step definitions are cleared as well.

svanoort:
🐝

jglick:
Hmm, I wonder if I could just use `GroovyClassLoader.getLoadedClasses()`…then I would merely need to enumerate the loaders. The main `GroovyClassLoader` inheritance chain is easy enough to follow; the tricky bit is that each compilation creates a new `InnerLoader`.

jglick:
Hard leak has been fixed, but I wonder if we can also clear the soft references:

```
private static final org.codehaus.groovy.reflection.ClassInfo$ClassInfoSet org.codehaus.groovy.reflection.ClassInfo.globalClassSet->
org.codehaus.groovy.reflection.ClassInfo$ClassInfoSet@48d46169-segments->
[Lorg.codehaus.groovy.util.AbstractConcurrentMapBase$Segment;@5f8ffbd0-[10]->
org.codehaus.groovy.reflection.ClassInfo$ClassInfoSet$Segment@30e6148f-table->
[Ljava.lang.Object;@565a8dc4-[8]->
org.codehaus.groovy.reflection.ClassInfo@29f484c8-ref->
org.codehaus.groovy.util.ReferenceType$SoftRef@27734616-referent->
java.lang.Class@573d4844-<changed>->
groovy.lang.GroovyClassLoader$InnerLoader@6595381a
```

rtyler:
@abayer Do you think these fixes could be part of the PermGen leak behavior [we](http://lists.jenkins-ci.org/pipermail/jenkins-infra/2016-September/000858.html) (and I know others) have seen on JDK7 with Multibranch?

jglick:
This is heap, not permanent generation. And run Java 8!

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1019,6 +1022,23 @@ private static void cleanUpGlobalClassSet(@Nonnull Class<?> clazz) throws Except
         }
     }
 
+    private static void cleanUpObjectStreamClassCaches(@Nonnull Class<?> clazz) throws Exception {
+        Class<?> cachesC = Class.forName("java.io.ObjectStreamClass$Caches");
+        for (String cacheFName : new String[] {"localDescs", "reflectors"}) {
+            Field cacheF = cachesC.getDeclaredField(cacheFName);
+            cacheF.setAccessible(true);
+            ConcurrentMap<Reference<Class<?>>, ?> cache = (ConcurrentMap) cacheF.get(null);
+            Iterator<? extends Entry<Reference<Class<?>>, ?>> iterator = cache.entrySet().iterator();
```
svanoort: How is mutation by concurrent threads handled wrt the iterator in this implementation?

jglick:
Still seeing the test straining to collect loaders, with

```
private static org.codehaus.groovy.reflection.ClassInfo some.arbitrary.class.here.implemented.in.Groovy.$staticClassInfo->
org.codehaus.groovy.reflection.ClassInfo@3b4340f8-segment->
org.codehaus.groovy.reflection.ClassInfo$ClassInfoSet$Segment@391eea7f-table->
[Ljava.lang.Object;@fd26c3b-[28]->
[Ljava.lang.Object;@5855cc7b-[0]->
org.codehaus.groovy.reflection.ClassInfo@4af28a48-weakMetaClass->
org.codehaus.groovy.util.ManagedReference@219c59f9-ref->
org.codehaus.groovy.util.ReferenceType$SoftRef@1e723fcd-referent->
groovy.lang.MetaClassImpl@3c355255-metaMethodIndex->
org.codehaus.groovy.runtime.metaclass.MetaMethodIndex@19f9130d-table->
[Lorg.codehaus.groovy.runtime.metaclass.MetaMethodIndex$Entry;@1771a9cc-[84]->
org.codehaus.groovy.runtime.metaclass.MetaMethodIndex$Entry@16d07783-cachedStaticMethod->
org.codehaus.groovy.runtime.metaclass.MetaMethodIndex$CacheEntry@7b35f458-params->
[Ljava.lang.Class;@606ad775-[0]->
java.lang.Class@9168226-<changed>->
groovy.lang.GroovyClassLoader$InnerLoader@4142b833
```

and on a live instance, `ThreadGroupContext.beanInfoCache` seems to be holding on to a bunch of Pipeline Script classes even though we have called `Introspector.flushFromCaches`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Should be fine, it is a `ConcurrentHashMap`.

jglick:
[JENKINS-32338](https://issues.jenkins-ci.org/browse/JENKINS-32338) seems to report something similar.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -362,6 +362,7 @@ private boolean run() {
         if (ending) {
             execution.cleanUpHeap();
             scripts.clear();
+            closures.clear();
```
jglick: Actually the bug was that it was nonempty at this point. Will correct in #279.

PR #84, opened 2016-11-03
[FIXED JENKINS-38551] Escape invalid chars in gdsl generation

rishi-manidhar:
Some plugins include strings with single-quotes in their `displayName`, causing the workflow plugin to generate a gdsl with invalid groovy. 

This pull request adds a simple test case that includes a single-quote character and also fixes the issue by using a groovy utility method to properly escape that variable before including it in the output. 

This PR fixes https://issues.jenkins-ci.org/browse/JENKINS-38551


src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -308,6 +310,12 @@ public void monomorphicListSymbol() throws Exception {
     }
 
     @Test
+    public void escapeCharacters() throws Exception {
+        DisplaynameWithEscapeCharState displaynameWithEscapeCharState = new DisplaynameWithEscapeCharState(new DisplaynameEscapeCharData("one"));
+        st.assertRoundTrip(displaynameWithEscapeCharState, "displaynameWithEscapeCharState([firstArg: 'one'])");
```
jglick: `assertRoundTrip` does not exercise either of the `dsld` or `gdsl` views AFAICT. Did you try running this test with your code patches temporarily reverted to see if it would fail?

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/DisplaynameEscapeCharData.java
```diff
@@ -0,0 +1,24 @@
+package org.jenkinsci.plugins.workflow.testMetaStep;
+
+import hudson.Extension;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Descriptor;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class DisplaynameEscapeCharData extends AbstractDescribableImpl<DisplaynameEscapeCharData> {
```
jglick: Gratuitous complexity I think; would suffice to have a no-arg `Step`.

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
rishi-manidhar: I tried running the tests without the changes and they did fail.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/DisplaynameEscapeCharData.java
rishi-manidhar: Removed unnecessary stuff, the tests now has just a step with no args and no-op execution.
Removed the DisplaynameEscapeCharData.java file as well.

PR #86, opened 2016-11-04
groovy-cps 1.11

jglick:
[Changes](https://github.com/cloudbees/groovy-cps/compare/groovy-cps-1.10...groovy-cps-1.11)

@reviewbybees esp. @kohsuke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
         <dependency>
             <groupId>com.cloudbees</groupId>
             <artifactId>groovy-cps</artifactId>
-            <version>1.10</version>
+            <version>1.11</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.ui</groupId>


svanoort:
🐝

oleg-nenashev:
Uhm, I'm not sure it is a bug...

`Env#getLocalVariableType()` is a new method introduced in the interface. Smells like a binary compat issue, which may widespread to dependencies. 

I see no usages in OSS packages, but I still feel uncomfortable after the Stapler fun.

oleg-nenashev:
^^ CC @svanoort , who introduced it in https://github.com/cloudbees/groovy-cps/commit/e7368eae7facfeb1676a1d5af93347748e524c0b

jglick:
> binary compat issue, which may widespread to dependencies.
> 
> I see no usages in OSS packages

`groovy-cps` should be used only internally in this plugin.

jglick:
In other words, we need to be concerned about compatibility in the serialized form, but not in Java signatures.

oleg-nenashev:
@jglick If it's not supposed to be used, there are Maven shading features for it. But feel free to go forward as a maintainer

jglick:
> there are Maven shading features for it

A possibility, but not so easy, since we would need to update serialized data too.

jglick:
Anyway in this case the Javadoc states clearly:

> This interface is not to be implemented outside this library.

PR #87, opened 2016-11-05
[JENKINS-39154] Enable plugin parent pom 2.17

tfennelly:
node/npm changes to allow use of parent pom v 2.17

Attn @jglick


tfennelly:
@jglick Sure, I'll look for plugins that might need it.

PR #89, opened 2016-11-08
scripts might be null if loading old builds

jglick:
Regression in #41. Corrects a `NullPointerException` thrown at the end of a very old build noticed in tests after https://github.com/jenkinsci/pipeline-input-step-plugin/pull/9.

@reviewbybees


abayer:
:bee:

PR #91, opened 2016-11-11
Cache the StepDescriptors to elliminate a moderate bottleneck

svanoort:
Profiling showed this is rather expensive, due to repeated iteration needs to find the descriptor. 

This converts this to a simple, fast hashmap lookup, rather than checking against a whole list of items. 

Side point: it probably reduces the garbage objects Jenkins generates.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
@@ -25,7 +25,6 @@
 package org.jenkinsci.plugins.workflow.cps.nodes;
 
 import hudson.Extension;
-import hudson.ExtensionListListener;
 import hudson.ExtensionPoint;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
@@ -44,19 +44,8 @@ import javax.annotation.CheckForNull;
 @Restricted(NoExternalUse.class)
 public class StepDescriptorCache implements ExtensionPoint {
 
-    /** Used ONLY for unit tests where there isn't actually a running Jenkins */
-    private static StepDescriptorCache fallbackSingleton = null;
-
     public static StepDescriptorCache getPublicCache() {
-        Jenkins myJenkins = Jenkins.getInstance();
-        if ( myJenkins == null) {
-            if (fallbackSingleton == null) {
-                 fallbackSingleton = new StepDescriptorCache();
-            }
-            return fallbackSingleton;
-        } else {
-            return myJenkins.getExtensionList(StepDescriptorCache.class).get(0);
-        }
+        return Jenkins.getActiveInstance().getExtensionList(StepDescriptorCache.class).get(0);
     }
 
     public StepDescriptorCache() {}
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
@@ -2,6 +2,7 @@ package org.jenkinsci.plugins.workflow.cps.nodes;
 
 import hudson.Extension;
 import hudson.ExtensionListListener;
+import hudson.util.Memoizer;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
@@ -17,22 +18,17 @@ import java.util.concurrent.ConcurrentHashMap;
 @Restricted(NoExternalUse.class)
 public class StepDescriptorCache {
 
-    // This ensures we use extension loading unless running a pure unit test with no Jenkins
-    private static StepDescriptorCache getCacheInstance() {
+    /** Used ONLY for unit tests where there isn't actually a running Jenkins */
+    private static final StepDescriptorCache fallbackSingleton = new StepDescriptorCache();
+
+    public static StepDescriptorCache getPublicCache() {
         Jenkins myJenkins = Jenkins.getInstance();
         if ( myJenkins == null) {
-            return new StepDescriptorCache();
+            return fallbackSingleton;
         } else {
             return myJenkins.getExtensionList(StepDescriptorCache.class).get(0);
         }
     }
-
-    private static final StepDescriptorCache cacheSingleton = StepDescriptorCache.getCacheInstance();
-
-    public static StepDescriptorCache getPublicCache() {
-        return cacheSingleton;
-    }
-
 
     ExtensionListListener myListener = new ExtensionListListener() {
         @Override
@@ -50,25 +46,14 @@ public class StepDescriptorCache {
         descriptorCache.clear();
     }
 
-    private final ConcurrentHashMap<String, StepDescriptor> descriptorCache = new ConcurrentHashMap<String, StepDescriptor>() {
-        public StepDescriptor get(String descriptorId) {
-            StepDescriptor output = super.get(descriptorId);
-            if (output == null) {
-                output = load(descriptorId);
-                if (output != null) {
-                    this.put(descriptorId, output); // We may get redundant writes but that is ok
-                }
-            }
-            return output;
-        }
+    private final Memoizer<String, StepDescriptor> descriptorCache = new Memoizer<String, StepDescriptor>() {
 
-        public StepDescriptor load(String descriptorId) {
+        public StepDescriptor compute(String descriptorId) {
             Jenkins j = Jenkins.getInstance();
-            if (j != null) {
+            if (descriptorId != null && j != null) {
                 return (StepDescriptor) j.getDescriptor(descriptorId);
-            } else {
-                return null;
             }
+            return null;
         }
     };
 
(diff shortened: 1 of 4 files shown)


svanoort:
@reviewbybees assuming tests pass

svanoort:
Kicking back to WIP since something went wonky on tests until I can resolve it.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -0,0 +1,82 @@
...
+    private static StepDescriptorCache getCacheInstance() {
+        Jenkins myJenkins = Jenkins.getInstance();
+        if ( myJenkins == null) {
+            return new StepDescriptorCache();
+        } else {
+            return myJenkins.getExtensionList(StepDescriptorCache.class).get(0);
+        }
+    }
+
+    private static final StepDescriptorCache cacheSingleton = StepDescriptorCache.getCacheInstance();
```
jglick: Do not hold this in a `static` field. Will be wrong after a restart.

jglick: Just inline `cacheSingleton` and `getCacheInstance` into `getPublicCache`.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -0,0 +1,82 @@
...
+    public StepDescriptorCache() {
+        // Ensures we purge the cache if steps change
+        Jenkins.getInstance().getExtensionList(StepDescriptor.class).addListener(myListener);
+    }
+
+    public void invalidateAll() {
+        descriptorCache.clear();
+    }
+
+    private final ConcurrentHashMap<String, StepDescriptor> descriptorCache = new ConcurrentHashMap<String, StepDescriptor>() {
```
jglick: Easier to use `CacheBuilder`.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepStartNode.java
```diff
@@ -34,11 +33,8 @@ public StepStartNode(CpsFlowExecution exec, StepDescriptor d, FlowNode parent) {
     }
 
     public StepDescriptor getDescriptor() {
-        if (descriptor == null) {
-            Jenkins j = Jenkins.getInstance();
-            if (j != null) {
-                descriptor = (StepDescriptor) j.getDescriptor(descriptorId);
-            }
+        if (descriptor == null && descriptorId != null) {
```
jglick: `descriptorId` cannot be null AFAIK.

svanoort: That's carried over from the original -- no harm in checking it anyway.

svanoort:
Still some wackiness to investigate here, surprisingly (for such a simple change). 
Local runs are getting Permgen (?!) errors.  

Need to check I'm not holding dangling references somewhere (maybe descriptors are not being cleared upon restart, or similar).

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
svanoort: Memoizer ends up being a really nice midpoint between the two, btw

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -1,74 +1,83 @@
...
 import javax.annotation.CheckForNull;
-import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Shared cacheSingleton for the StepDescriptors, extension-scoped to avoid test issues
  */
 @Extension
 @Restricted(NoExternalUse.class)
-public class StepDescriptorCache {
+public class StepDescriptorCache implements ExtensionPoint {
```
jglick: :ant: Not really an `ExtensionPoint`, it is just a singleton which we would not want to advertise [here](https://wiki.jenkins-ci.org/display/JENKINS/Extension+points).

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -1,74 +1,83 @@
...
 
-    // This ensures we use extension loading unless running a pure unit test with no Jenkins
-    private static StepDescriptorCache getCacheInstance() {
+    /** Used ONLY for unit tests where there isn't actually a running Jenkins */
+    private static StepDescriptorCache fallbackSingleton = null;
```
jglick: No actual need to cache this, since it will return null for everything anyway.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -1,74 +1,83 @@
...
-        @Override
-        public void onChange() {
-            invalidateAll();
-        }
-    };
-
-    public StepDescriptorCache() {
-        // Ensures we purge the cache if steps change
-        Jenkins.getInstance().getExtensionList(StepDescriptor.class).addListener(myListener);
+    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED)  // Prevents potential leakage on reload
```
jglick: You should not need this. Every reload will get a fresh set of extensions.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
svanoort: @jglick This is exactly the opposite of what we did (at your insistence) with the stage view caching.  What is different between them that dictates different handling?

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
svanoort: @jglick You're right, I was coding overly defensively here and it doesn't add anything.  Can you re-review with that change please?

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -0,0 +1,80 @@
...
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * Shared cacheSingleton for the StepDescriptors, extension-scoped to avoid test issues
+ */
+@Extension
+@Restricted(NoExternalUse.class)
+public class StepDescriptorCache implements ExtensionPoint {
```
jglick: best not to implement

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public static StepDescriptorCache getPublicCache() {
+        return Jenkins.getActiveInstance().getExtensionList(StepDescriptorCache.class).get(0);
+    }
+
+    public StepDescriptorCache() {}
+
+    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED)  // Prevents potential leakage on reload
```
jglick: probably unnecessary

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Shared cacheSingleton for the StepDescriptors, extension-scoped to avoid test issues
+ */
+@Extension
+@Restricted(NoExternalUse.class)
+public class StepDescriptorCache implements ExtensionPoint {
+
+    public static StepDescriptorCache getPublicCache() {
+        return Jenkins.getActiveInstance().getExtensionList(StepDescriptorCache.class).get(0);
```
jglick: can also use `ExtensionList.lookup`

oleg-nenashev: Not safe if somebody disables the extension point (e.g. via system Groovy), there may be an empty list.
I would rather vote for having explicit handling instead of NPE. Nit

abayer:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
oleg-nenashev: Does reload pass through this state? Do not recall it

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepStartNode.java
```diff
@@ -34,11 +33,8 @@ public StepStartNode(CpsFlowExecution exec, StepDescriptor d, FlowNode parent) {
     }
 
     public StepDescriptor getDescriptor() {
```
oleg-nenashev: Add CheckForNull?

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
svanoort: It sure seems to from what I can tell.   At worst this is a no-op, at best it prevents issues.

svanoort:
Merging but I've got one other goodie I spotted earlier that I'd like to slip in before we cut a release (also a little ATH testing of the whole)

PR #92, opened 2016-12-06
Avoid saving program.dat if we are about to run another chunk anyway

jglick:
Reverts this aspect of behavior to that prior to #46, which should avoid the incompatibility noted in [2.14](https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Groovy+Plugin#PipelineGroovyPlugin-2.14%28Sep07%2C2016%29). Script authors should still of course avoid using unserializable local variables.

Also avoids some load spent saving `program.dat` unnecessarily, though in my tests so far the benefit is typically negligible: while roughly half of the program saves are avoided, each one only took a few milliseconds anyway. Could affect certain scripts more, if they do a lot of (CPS-transformed) Groovy in between steps.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -355,7 +355,7 @@ public final class CpsThreadGroup implements Serializable {
             }
         }
 
-        if (changed) {
+        if (changed && !stillRunnable) {
             try {
                 saveProgram();
             } catch (IOException x) {
@@ -420,6 +420,9 @@ public final class CpsThreadGroup implements Serializable {
 
     @CpsVmThreadOnly
     public void saveProgram(File f) throws IOException {
+        boolean logging = LOGGER.isLoggable(Level.FINER);
+        long start = logging ? System.nanoTime() : 0;
+
         File dir = f.getParentFile();
         File tmpFile = File.createTempFile("atomic",null, dir);
 
@@ -451,6 +454,11 @@ public final class CpsThreadGroup implements Serializable {
         } finally {
             PROGRAM_STATE_SERIALIZATION.set(old);
             Util.deleteFile(tmpFile);
+        }
+
+        if (logging) {
+            long end = System.nanoTime();
+            LOGGER.log(FINER, "saved {0} of size {1}Kb in {2}ms", new Object[] {f, f.length() / 1000, (end - start) / 1000 / 1000});
         }
     }
 


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -452,6 +455,11 @@ public void saveProgram(File f) throws IOException {
             PROGRAM_STATE_SERIALIZATION.set(old);
             Util.deleteFile(tmpFile);
         }
+
+        if (logging) {
+            long end = System.nanoTime();
```
svanoort: Guarded logging 🎆

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -452,6 +455,11 @@ public void saveProgram(File f) throws IOException {
...
+            LOGGER.log(FINER, "saved {0} of size {1}Kb in {2}ms", new Object[] {f, f.length() / 1000, (end - start) / 1000 / 1000});
```
svanoort: Also, logging the I/O size and time :champagne: 

IIRC we can set up logging for just this class to FINER level, and get the cost of that I/O operation right?

svanoort:
I think this will deliver deceptively large gains for complex pipelines (far more than the basic test will accept)... so want this included badly.  Well-spotted, @jglick !

However I am not sure about the impact on pipeline durability -- are we saving program state before beginning a step itself (maybe somewhere I have missed)?  Otherwise we may be losing the ability to survive a restart?

As long as we've got persistence of program state around step executions themselves I see no obstacles to allowing the program to flow freely -- and that's a HUGE potential performance boost.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -420,6 +420,9 @@ void saveProgram() throws IOException {
 
     @CpsVmThreadOnly
     public void saveProgram(File f) throws IOException {
+        boolean logging = LOGGER.isLoggable(Level.FINER);
```
jglick: Could. Anyway I plan to revert this and incorporate it into a more general timing reporting system I am working on for Pipeline builds.

jglick:
> I think this will deliver deceptively large gains for complex pipelines

Well, I _tested_ it on a complex pipeline, and saw no measurable difference.

> are we saving program state before beginning a step itself

Yes, except for `[Abstract]SynchronousStepExecution`s such as `echo`.

Basically this takes us back to the saving behavior in version 1.13. The reason the code here was changed in 1.14 was that we wanted to be able to abort a build running inside synchronous CPS-transformed Groovy; previously the CPS VM would run as much as it could in one uninterruptible batch. But @kohsuke introduced a save at every such intermediate step, I think unintentionally.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
jglick: Anyway `isLoggable` is very cheap—negligible compared to the cost of the rest of the method. And checking this in a `static` initializer would block you from changing the logging level at runtime.

svanoort: True -- I couldn't remember if this was one of the methods overridden to do a deceptively expensive lookup.  Java logging can be terrible sometimes. 

I take back my nit after looking more closely at the code path involved here.

svanoort:
> Well, I tested it on a complex pipeline, and saw no measurable difference.

You gotta get my hopes up and then dash my dreams!  Heh, ah well. 

> Yes, except for [Abstract]SynchronousStepExecutions such as echo

If we can provide that guarantee and you can point me at where KK added the intermediate save then I feel very comfortable bee-ing this.

jglick:
> point me at where KK added the intermediate save

See the line I commented in #46, though you really need to read the whole patch to `CpsThreadGroup` to understand the behavioral change.

You can put in some printlns that trigger when `changed` and print `stillRunnable` to see that we get a series of `true`s followed by one `false`, typically before a step runs (but also at other times, such as the body of a `node` block getting scheduled).

PR #93, opened 2016-12-06
Record timing information for internal processes

jglick:
Useful for determining, both during experimentation and in the field, where the most load is consumed by the Jenkins master when running Pipeline builds: class loading for Groovy compilation, saving `program.dat`, etc.

Supersedes the simple logging in #92.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -32,6 +32,9 @@ import com.cloudbees.groovy.cps.impl.ConstantBlock;
 import com.cloudbees.groovy.cps.impl.ThrowBlock;
 import com.cloudbees.groovy.cps.sandbox.DefaultInvoker;
 import com.cloudbees.groovy.cps.sandbox.SandboxInvoker;
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.Content;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.FutureCallback;
@@ -98,14 +101,21 @@ import groovy.lang.GroovyClassLoader;
 import groovy.lang.GroovyCodeSource;
 import hudson.AbortException;
 import hudson.BulkChange;
+import hudson.Extension;
 import hudson.init.Terminator;
 import hudson.model.Item;
+import hudson.model.Job;
 import hudson.model.Queue;
+import hudson.model.Run;
 import hudson.model.Saveable;
 import hudson.model.User;
 import hudson.security.ACL;
 import hudson.security.AccessControlled;
+import hudson.security.Permission;
 import java.beans.Introspector;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
 import java.lang.ref.Reference;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
@@ -124,6 +134,7 @@ import jenkins.security.NotReallyRoleSensitiveCallable;
 
 import org.acegisecurity.Authentication;
 import org.acegisecurity.userdetails.UsernameNotFoundException;
+import org.apache.commons.io.Charsets;
 import org.jboss.marshalling.reflect.SerializableClassRegistry;
 
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
@@ -241,7 +252,7 @@ public class CpsFlowExecution extends FlowExecution {
     /**
      * Recreated from {@link #owner}
      */
-    /*package*/ transient /*almos final*/ FlowNodeStorage storage;
+    /*package*/ transient /*almos final*/ TimingFlowNodeStorage storage;
 
     /** User ID associated with this build, or null if none specific. */
     private final @CheckForNull String user;
@@ -310,6 +321,41 @@ public class CpsFlowExecution extends FlowExecution {
     /** Actions to add to the {@link FlowStartNode}. */
     transient final List<Action> flowStartNodeActions = new ArrayList<Action>();
 
+    enum TimingKind {
+        /**
+         * Parsing Groovy sources; includes {@link #load}.
+         * @see CpsGroovyShell#parse(GroovyCodeSource)
+         */
+        parse,
+        /**
+         * Loading classes needed during {@link #parse}.
+         * @see ClassLoader#loadClass(String, boolean)
+         * @see ClassLoader#getResource
+         */
+        load,
+        /**
+         * Running inside {@link CpsVmExecutorService}, which includes many other things.
+         */
+        run,
+        /**
+         * Saving the program state.
+         * @see CpsThreadGroup#saveProgram(File)
+         */
+        save,
+        /**
+         * Loading or saving flow nodes.
+         * @see FlowNodeStorage
+         */
+        flow
+    }
+
+    /** accumulated time in ns of a given kind */
+    @GuardedBy("this")
+    @CheckForNull private Map</* for pretty XStream form */String, Long> timings;
+    /** last start time in ns of a given kind; unreliable to just subtract on start and add on end, since the VM might die in between */
+    @GuardedBy("this")
+    @CheckForNull private transient Map<TimingKind, Long> starts;
+
     @Deprecated
     public CpsFlowExecution(String script, FlowExecutionOwner owner) throws IOException {
         this(script, false, owner);
@@ -331,6 +377,40 @@ public class CpsFlowExecution extends FlowExecution {
         if (loadedScripts==null)
             loadedScripts = new HashMap<String,String>();   // field added later
         return this;
+    }
+
+    /**
+     * Record time taken during a certain class of operation in this build.
+     * @param kind what sort of operation is being done
+     * @param end false to start, true at the end (pair in a {@code finally}-block)
+     */
+    synchronized void time(@Nonnull TimingKind kind, boolean end) {
+        if (timings == null) {
+            timings = new HashMap<>();
+        }
+        if (starts == null) {
+            starts = new HashMap<>();
+        }
+        if (end) {
+            Long orig = timings.get(kind.name());
+            if (orig == null) {
+                orig = 0L;
+            }
+            assert starts.containsKey(kind) : kind + " not in " + starts;
+            timings.put(kind.name(), orig + System.nanoTime() - starts.get(kind));
+        } else {
+            starts.put(kind, System.nanoTime());
+        }
+    }
+
+    synchronized void logTimings() {
+        if (LOGGER.isLoggable(Level.FINE)) {
+            Map<String, String> formatted = new TreeMap<>();
+            for (Map.Entry<String, Long> entry : timings.entrySet()) {
+                formatted.put(entry.getKey(), entry.getValue() / 1000 / 1000 + "ms");
+            }
+            LOGGER.log(Level.FINE, "timings for {0}: {1}", new Object[] {owner, formatted});
+        }
     }
 
     /**
@@ -376,8 +456,8 @@ public class CpsFlowExecution extends FlowExecution {
         return owner;
     }
 
-    private SimpleXStreamFlowNodeStorage createStorage() throws IOException {
-        return new SimpleXStreamFlowNodeStorage(this, getStorageDir());
+    private TimingFlowNodeStorage createStorage() throws IOException {
+        return new TimingFlowNodeStorage(new SimpleXStreamFlowNodeStorage(this, getStorageDir()));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


jglick:
(build passed, GH seems to have lost the notification)

svanoort:
Close and reopen to rerun tests -- also, I really want to be able to do a deep review before this is merged.

jglick:
Sure, I do not think there is any rush on this.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
```diff
@@ -78,4 +95,37 @@ protected synchronized String generateScriptName() {
...
+        TimingLoader(ClassLoader parent, CpsFlowExecution execution) {
+            super(parent);
+            this.execution = execution;
+        }
+        @Override protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+            execution.time(CpsFlowExecution.TimingKind.load, false);
+            try {
+                return super.loadClass(name, resolve);
+            } finally {
+                execution.time(CpsFlowExecution.TimingKind.load, true);
```
jglick: Becomes a bit easier in `-source 8`, so long as each body returns at most one kind of checked exception.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -241,7 +252,7 @@
     /**
      * Recreated from {@link #owner}
      */
-    /*package*/ transient /*almos final*/ FlowNodeStorage storage;
+    /*package*/ transient /*almos final*/ TimingFlowNodeStorage storage;
```
svanoort: 🐜  "almos"

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -310,6 +321,41 @@
...
+         * Parsing Groovy sources; includes {@link #load}.
+         * @see CpsGroovyShell#parse(GroovyCodeSource)
+         */
+        parse,
+        /**
+         * Loading classes needed during {@link #parse}.
+         * @see ClassLoader#loadClass(String, boolean)
+         * @see ClassLoader#getResource
+         */
+        load,
```
svanoort: 🐛  would rename to "classLoad" because "load" could just as easily mean reading from disk/initializing (which is potentially something we might benchmark, i.e. loading program.dat read-in on resumption).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -310,6 +321,41 @@
...
+        /**
+         * Running inside {@link CpsVmExecutorService}, which includes many other things.
+         */
+        run,
+        /**
+         * Saving the program state.
+         * @see CpsThreadGroup#saveProgram(File)
+         */
+        save,
```
svanoort: 🐜  Better to call it saveProgram

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -310,6 +321,41 @@
...
+        /**
+         * Loading or saving flow nodes.
+         * @see FlowNodeStorage
+         */
+        flow
```
svanoort: 🐛  better to call it flowNodeIO... again because there are parts of the logic that interact with the flow (iteration, in-memory work) which are separate from I/O routines.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -310,6 +321,41 @@
...
+    }
+
+    /** accumulated time in ns of a given kind */
+    @GuardedBy("this")
+    @CheckForNull private Map</* for pretty XStream form */String, Long> timings;
```
svanoort: 🐛 Do not put comments inline in a generics declaration.  They can go after the ; just fine.  This is bad enough to be more than just a 🐜  I feel.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -310,6 +321,41 @@
...
+    /** last start time in ns of a given kind; unreliable to just subtract on start and add on end, since the VM might die in between */
+    @GuardedBy("this")
+    @CheckForNull private transient Map<TimingKind, Long> starts;
+
```
svanoort: 🐜  but more of a trivial RFE: to make this useful, we should add one more map for call counts.  This lets us distinguish between a few very long calls vs. a lot of short ones. 

Without this we aren't likely to have the information needed to attack bottlenecks.  It will ALSO let us see if a method function simply has not been run enough times to JIT, which makes a huge difference.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: (We may want to instrument those in-memory operations in the future)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -466,13 +464,9 @@ public void saveProgram(File f) throws IOException {
             throw new IOException("Failed to persist "+f,e);
         } finally {
             PROGRAM_STATE_SERIALIZATION.set(old);
+            execution.time(TimingKind.save, true);
             Util.deleteFile(tmpFile);
         }
-
-        if (logging) {
```
svanoort: I actually kind of *liked* the notion of saving some I/O size info -- not via logging, but maybe via the map of info?

svanoort:
@jglick see bugs/comments/review

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: 🐛  For forcing TimingFlowNodeStorage in all cases -- extra overheads for the hot path. Recommend we use a quick check where doing timing and only perform those operations if we are tracking timing.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -334,6 +380,40 @@ private Object readResolve() {
...
+        if (timings == null) {
+            timings = new HashMap<>();
+        }
+        if (starts == null) {
+            starts = new HashMap<>();
+        }
+        if (end) {
+            Long orig = timings.get(kind.name());
+            if (orig == null) {
+                orig = 0L;
```
svanoort: 🐛  - no. If orig is null, we cannot put any timing info.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -334,6 +380,40 @@ private Object readResolve() {
...
+            }
+            assert starts.containsKey(kind) : kind + " not in " + starts;
+            timings.put(kind.name(), orig + System.nanoTime() - starts.get(kind));
+        } else {
+            starts.put(kind, System.nanoTime());
+        }
+    }
+
+    synchronized void logTimings() {
```
svanoort: Perhaps better to only invoke logTimings if log level is appropriate, but make it *always* log?  That way we can use groovy to invoke logTimings explicitly and dump out useful information from a running Jenkins without getting the usual `Level.FINE` spam?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1316,4 +1403,85 @@ private void setField(CpsFlowExecution result, String fieldName, Object value) {
      * this field is set to {@link CpsFlowExecution} that will own it.
      */
     static final ThreadLocal<CpsFlowExecution> PROGRAM_STATE_SERIALIZATION = new ThreadLocal<CpsFlowExecution>();
+
+    class TimingFlowNodeStorage extends FlowNodeStorage {
+        private final FlowNodeStorage delegate;
```
svanoort: 🎆  so glad this is a raw FlowNodeStorage -- this means optimized FlowNodeStorage can use this bad boy.

svanoort:
Make no mistake, I may have landed a ton of bugs or nits on this one, but that's only because the implementation needs some work -- it was and still is a very good and useful idea, and the approach here is quite clever.

svanoort:
Another hard 🐛  with the starts map: it breaks if you have concurrent operations trying to load (for example) load different flowNodes. 

Timing has to take a time value, it can't track start/stop in the map, I'm afraid -- no way around it.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: Looks generally sane though, the delegate approach is probably the best way (once we handle the timings map).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1316,4 +1403,85 @@ private void setField(CpsFlowExecution result, String fieldName, Object value) {
...
+            try {
+                delegate.saveActions(node, actions);
+            } finally {
+                time(TimingKind.flow, true);
+            }
+        }
+    }
+
+    // If we wanted to expose via REST and/or floatingBox, could add a TransientActionFactory to show similar information.
+    @Extension(optional=true) public static class PipelineTimings extends Component {
```
svanoort: 🔖  for me: I need to come back and give the components a deep review once the initial issues are resolved.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
svanoort: Although... potentially that level of fine grain might be better to reserve for OS-level or profiler instrumentation.

svanoort:
🐛 - needs *some* test coverage of the basic timing/logging functionality.   The existing tests will ensure it doesn't break normal operation, but the additional features need coverage.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: Specifically the support component here.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Uh, what is the problem exactly? I am commenting on a particular parameter to the type declaration, so that is where I am putting the comment.

svanoort: It's terrible for readability and on a very short declaration there's absolutely no need.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: We are always tracking timing.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Huh? `orig == null` if this is the first timing of its kind we have performed in this build.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: If the log level is not right, calling `log` is a no-op, so that makes no sense.

I could switch to a more specific logger though, so you could turn just this on.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
jglick: Yeah, I would say _if_ timing information suggests that program saving is a major bottleneck then we go back and see why.

jglick:
```
java.lang.AssertionError: expected:<5> but was:<4>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecutionTest$9.evaluate(CpsFlowExecutionTest.java:296)
```

svanoort:
@jglick Fix is to bump the workflow-support version IIRC.

jglick:
@svanoort said he would try to review soon.

jglick:
tried

* running master
   * creating job to `sleep 99`
   * building
   * restarting
   * copying that job
* running branch
   * building copy job
   * restarting
   * getting support bundle
   * waiting for first build to complete
   * getting support bundle
   * running a second build
   * aborting second build
   * getting support bundle

The build of the original job recorded timings for `flowNode` only.

svanoort:
Testing seems reasonable to me.

PR #94, opened 2016-12-17
[JENKINS-29656] Retry I/O operations due to Windows file locks.

cowwoc:
Attempted fix for [JENKINS-29656](https://issues.jenkins-ci.org/browse/JENKINS-29656).

(I couldn't test this locally because unit tests were failing even before this change, but I've used similar code in one of my other projects and it works fine.)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -443,6 +443,7 @@ public final class CpsThreadGroup implements Serializable {
                 }
                 Files.move(tmpFile.toPath(), f.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
                 LOGGER.log(FINE, "program state saved");
+                break;
             } catch (RuntimeException e) {
                 if (i < MAX_RETRIES - 1) {
                     LOGGER.log(FINE, "Retry [" + (i + 1) + "/" + MAX_RETRIES + "]", e);
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -444,19 +444,6 @@ public final class CpsThreadGroup implements Serializable {
                 Files.move(tmpFile.toPath(), f.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
                 LOGGER.log(FINE, "program state saved");
                 break;
-            } catch (RuntimeException e) {
-                if (i < MAX_RETRIES - 1) {
-                    LOGGER.log(FINE, "Retry [" + (i + 1) + "/" + MAX_RETRIES + "]", e);
-                    long timeout = Math.min(1000, (long) (10 * Math.pow(2, i)));
-                    try {
-                        TimeUnit.MILLISECONDS.sleep(timeout);
-                        continue;
-                    } catch (InterruptedException ignore) {
-                        // give up
-                    }
-                }
-                propagateErrorToWorkflow(e);
-                throw new IOException("Failed to persist "+f,e);
             } catch (IOException e) {
                 if (i < MAX_RETRIES - 1) {
                     LOGGER.log(FINE, "Retry [" + (i + 1) + "/" + MAX_RETRIES + "]", e);
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -57,6 +57,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -428,24 +433,47 @@ public void saveProgram(File f) throws IOException {
             return;
         }
 
-        try {
```
jglick: [Reduced diff](https://github.com/jenkinsci/workflow-cps-plugin/pull/94/files?w=1)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -428,24 +433,47 @@ public void saveProgram(File f) throws IOException {
...
+                RiverWriter w = new RiverWriter(tmpFile, execution.getOwner());
+                try {
+                    w.writeObject(this);
+                } finally {
+                    w.close();
+                }
+                Files.move(tmpFile.toPath(), f.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
+                LOGGER.log(FINE, "program state saved");
+                break;
+            } catch (RuntimeException e) {
```
jglick: No reported case is of anything other than an `IOException` so I would not want to retry in the face of other errors.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -428,24 +433,47 @@ public void saveProgram(File f) throws IOException {
...
+                    }
+                }
+                propagateErrorToWorkflow(e);
+                throw new IOException("Failed to persist "+f,e);
+            } catch (IOException e) {
+                if (i < MAX_RETRIES - 1) {
+                    LOGGER.log(FINE, "Retry [" + (i + 1) + "/" + MAX_RETRIES + "]", e);
+                    long timeout = Math.min(1000, (long) (10 * Math.pow(2, i)));
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(timeout);
```
jglick: We do not really want to sleep; that just blocks the CPS VM thread, usually unnecessarily.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
cowwoc: What's the impact if there is not (another upcoming save operation)? What data gets lost?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
cowwoc: Okay. Agreed.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
cowwoc: What's the alternative? Retrying without a sleep will hog the CPU. We need to give the locking process time to release the lock.

cowwoc:
> I think this whole approach is actually wrong. The only thing we care about in this context is the Files.move call. And if that fails, well, we can log a warning to the log but there is no reason to halt the build—it is quite likely there will be another save operation soon enough anyway.

What's the impact if there is no upcoming save operation? What data is lost?

svanoort:
@cowwoc  Impact if there is no upcoming save operation is either:

1. Build completes (success or failure) and we go our merry way (most likely case, program.dat will then be irrelevant). 
2. Build is interrupted midway through by a Jenkins restart... in this case we are left with something of an undefined state because flownodes may be written but the program.dat will not be correct.  @jglick may want to be aware of this possibility -- my suspicion is that we will see a failure and exception thrown, either in CpsFlowExecution.onLoad or from the CpsThread.

There's a distant possibility that #2 will result in loading an old program.dat... however in cases where we cannot rely on the filesystem writing successfully, all bets are off anyway.

cowwoc:
@svanoort 

> however in cases where we cannot rely on the filesystem writing successfully, all bets are off anyway.

Fair enough, but I've never seen a failure under Windows once a retry loop was added. The file-system is asking you to try later. It's not a fatal failure.

I don't understand @jglick's reluctance to retry (what's the harm?) In my opinion, even limiting retries to under 1 second total is better than no retries at all (though I'd prefer a 30 second timeout.)

svanoort:
@cowwoc Like Jesse said:

> there is no reason to halt the build—it is quite likely there will be another save operation soon enough anyway

It's expected that individual steps will on average complete rapidly (aside from some longer-running shell/batch steps) -- given that average pipelines will trigger a few hundred persistence calls, if you do a retry with any frequency then it will seriously slow the runtime.  

A single retry *might* be reasonable, but on average you're probably better off to just keep going.

cowwoc:
@svanoort Are we able to detect whether we are executing the last persistence call? If so, perhaps we could retry only on this one.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
jglick: Probably you need not retry at all, but if you did want to, do so

* only if the failure is specifically in the `move` call, as opposed to `writeObject` which is expected to be a bug in the program, as in my PR
* do so later (but in the VM thread), as in `CpsStepContext.saveState`

cowwoc: @jglick I understand the first request (catch `move` but not `writeObject`) but I don't fully understand the second request.

I see how `CpsStepContext.saveState` invokes `saveProgram()` from inside `runInCpsVmThread()`, but what are you asking me to do about this in the context of `CpsThreadGroup.saveProgram()`?

dwnusbaum:
AIUI the main issue was fixed with #96. I am going to close this PR since it has been open for a few years with a merge conflict, but feel free to reopen if the conflict is resolved. Thanks!

PR #96, opened 2017-01-12
[JENKINS-29656] Better handle failure renaming program.dat

jglick:
[JENKINS-29656](https://issues.jenkins-ci.org/browse/JENKINS-29656)

Intended to replace #94.

@reviewbybees


jglick:
> inconsistent state between program and flow graph

Possibly. I suspect it would be mostly harmless, though. You might for example wind up rerunning a step (with a new `FlowNode`) because the saved program failed to record the fact that it had been run already.

PR #97, opened 2017-01-13
[JENKINS-33273] Allow lightweight checkouts to be used from CpsScmFlowDefinition

jglick:
Analogue of https://github.com/jenkinsci/workflow-multibranch-plugin/pull/47 for standalone Pipeline projects.

@reviewbybees


pom.xml
```diff
@@ -199,7 +199,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>2.4.1</version>
+            <version>3.0.2-beta-2</version>
```
jglick: Will pick up https://github.com/jenkinsci/git-plugin/pull/467 when available.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -73,18 +76,36 @@ public String getScriptPath() {
...
             }
         }
-        FilePath dir;
         Queue.Executable _build = owner.getExecutable();
         if (!(_build instanceof Run)) {
             throw new IOException("can only check out SCM into a Run");
         }
         Run<?,?> build = (Run<?,?>) _build;
+        if (lightweight) {
+            SCMFileSystem fs = SCMFileSystem.of(build.getParent(), scm);
```
stephenc: use try with resources

jglick:
@reviewbybees done

(but not safe to merge until @stephenc has sorted out the dependencies)

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-lightweight.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    If selected, try to obtain the Pipeline script contents directly from the SCM without performing a full checkout.
+    The advantage of this mode is its efficiency; however, you will not get any changelogs or polling based on the SCM.
```
stephenc: You should be able to get the changelogs with the `SCMFilesystem` API...

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/config.jelly
```diff
@@ -29,6 +29,9 @@ THE SOFTWARE.
     <f:entry field="scriptPath" title="${%Script Path}">
         <f:textbox default="Jenkinsfile"/>
     </f:entry>
+    <f:entry field="lightweight" title="${%Lightweight checkout}">
```
stephenc: I would think you would at most want an advanced option to *not* try lightweight. If you can do lightweight, why wouldn't you want to?

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-lightweight.html
jglick: Would require a significant rewrite. Currently `workflow-job` knows to record changelogs, or consider polling, based on `SCMListener` methods which cannot be called in lightweight mode. I considered such a large change (APIs etc.) out of scope here, especially given that

* many people will not care about changelogs or polling for the SCM containing the script
* if you `checkout scm` the same repo later, you will get those anyway

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/config.jelly
jglick: I preferred to err on the side of leaving existing behavior alone by default for now.

* lack of changelog/polling support (see below)
* new code path with likely bugs compared to original

pom.xml
```diff
@@ -95,7 +95,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>1.1</version>
+            <version>2.0.1</version>
```
jglick: Should ideally be bumped up to whatever release actually makes it to the UC.

jglick:
@reviewbybees done

Though I am still considering introducing an option or at least a flag to disable https://github.com/jenkinsci/workflow-multibranch-plugin/pull/47.

recampbell:
This looks mergable?

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/config.jelly
HRMPW: This box is selected by default. I think it is safer to maintain existing behavior for existing Pipelines.

jglick:
Needs a code review, and would like signoff from @HRMPW as well.

HRMPW:
The UI change is minimal and looks good. :bee:

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/config.jelly
jglick: That is what we are doing—`default` is used for _new_ definitions.

PR #98, opened 2017-01-25
[JENKINS-37324] Store parameter info for steps as an action on the flownode

svanoort:
Tracks the step parameters so you have a record after each step execution of what was actually run. 

Supports masking of inputs for security reasons. 

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/26 -- won't build without that as a snapshot.


jglick:
@svanoort please use forks for pull requests rather than origin branches.

jglick:
> won't build without that as a snapshot

So, `deploy` a snapshot, and edit the dependency here to refer to that timestamp.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -179,6 +191,18 @@ protected Object invokeStep(StepDescriptor d, Object args) {
...
+            if (isKeepStepInfo() && !(s instanceof ParallelStep)) { // Storing ArgumentsActions or not
+                Computer comp = context.get(Computer.class);
+                if (comp != null) {
+                    // Get the environment variables to find ones that might be credentials bindings
+                    // But filter out variables coming from the host itself
+                    EnvVars allEnv = new EnvVars(context.get(EnvVars.class));
+                    allEnv.entrySet().removeAll(comp.getEnvironment().entrySet());
+                    an.addAction(new ArgumentsActionImpl(ps.namedArgs, allEnv));
+                } else {
+                    an.addAction(new ArgumentsActionImpl(ps.namedArgs));  // No EnvVars that can supply credentials bindings
```
abayer: :bug: Not true after https://github.com/jenkinsci/credentials-binding-plugin/pull/34 landed

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,190 @@
...
+            }
+            pattern.append(Pattern.quote(val));
+            count++;
+        }
+        return (count > 0)
+                ? !Pattern.compile(pattern.toString()).matcher(input).find()
+                : true;
+    }
+
+    private static final HashSet<String> SAFE_ENVIRONMENT_VARIABLES = new HashSet<String>(Arrays.asList(
```
abayer: This makes me wish we could identify the provider of an environment variable with more ease, rather than whitelisting. But I don't know how that could be done...ah well.

abayer:
Looks good at first read-through.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: Resolved as of f58c625639f119721a962b8d1b3e4c794fbeb5be

svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -98,6 +104,14 @@ protected Object readResolve() throws IOException {
         return this;
     }
 
+    /** Allows user to disable saving step info, if this imposes too high a performance overhead
+     *   by setting System Property org.jenkinsci.plugins.workflow.cps.DSL.keepStepInfo to 'false'.
+     *  This may be modified at runtime by the user if needed for debugging. */
+    private static boolean isKeepStepInfo() {
+        String prop = System.getProperty(DSL.class.getName()+".keepStepInfo");
```
stephenc: Bad pattern for a run-time property. Normal pattern is a static non-final field initialized to the system property. Jenkins admins will be confused.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+public class ArgumentsActionImpl extends ArgumentsAction {
+
+    /** Arguments to the step, for cases where we cannot simply store the step because masking was applied */
+    @CheckForNull
+    private Map<String,Object> arguments;
+
+
+    private boolean isModifiedBySanitization = false;
+
+    public boolean isModifiedBySanitization() {
```
jglick: Not called; delete.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import static org.jenkinsci.plugins.workflow.actions.ArgumentsAction.NotStoredReason.MASKED_VALUE;
+import static org.jenkinsci.plugins.workflow.actions.ArgumentsAction.NotStoredReason.OVERSIZE_VALUE;
+
+/**
+ * Implements {@link ArgumentsAction} by storing step arguments, with sanitization
+ */
+public class ArgumentsActionImpl extends ArgumentsAction {
```
jglick: Probably should be `@Restricted(NoExternalUse.class)`.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    private boolean isModifiedBySanitization = false;
+
+    public boolean isModifiedBySanitization() {
+        return isModifiedBySanitization;
+    }
+
+    public ArgumentsActionImpl(@Nonnull Map<String, Object> stepArguments, @Nullable EnvVars env) {
+        Map<String,Object> sanitizedArguments = sanitizedStepArguments(stepArguments, env);
+        for (Object o : sanitizedArguments.values()) {
+            if (o != null && (o.equals(MASKED_VALUE) || o.equals(NotStoredReason.OVERSIZE_VALUE))) {
```
jglick: Why not simply

```java
if (o instanceof NotStoredReason) {
```

?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                isModifiedBySanitization = true;
+                break;
+            }
+        }
+        arguments = sanitizedArguments;
+    }
+
+    /** Create a step, sanitizing strings for secured content */
+    public ArgumentsActionImpl(@Nonnull Map<String, Object> stepArguments) {
```
jglick: Unused, delete.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+        this(stepArguments, new EnvVars());
+    }
+
+    /** See if sensitive environment variable content is in a string */
+    public static boolean isStringSafe(@CheckForNull String input, @CheckForNull EnvVars variables, @Nonnull Set<String> safeEnvVariables) {
```
jglick: Could be `private`.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                continue;
+            }
+            if (count > 0) {
+                pattern.append('|');
+            }
+            pattern.append(Pattern.quote(val));
+            count++;
+        }
+        return (count > 0)
+                ? !Pattern.compile(pattern.toString()).matcher(input).find()
```
jglick: I suppose the hope is that compiling & running this regexp is faster than repeated calls to `String.contains`?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: What makes these “safe”? What would make other environment variables “unsafe”? This looks wrong to me, but at least it deserves some careful Javadoc explaining how you arrived at this list and how it would be maintained.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+            "TERM",
+            "USER",
+            "XFILESEARCHPATH"
+    ));
+
+    /**
+     * Does first-level sanitization of a step's params, returning the Describable arguments map
+     * @param stepArguments Arguments provided when creating the step
+     * @param variables Environment variables to remove
+     * @return
```
jglick: Document the return value or delete the tag.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/DescriptorMatchPredicate.java
```diff
@@ -0,0 +1,48 @@
...
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Search predicate for {@link FlowNode}s that have a given descriptor type
+ */
+public class DescriptorMatchPredicate implements Predicate<FlowNode> {
```
jglick: Only used in `ArgumentsActionImplTest` AFAICT, so better moved to `src/test/java/`.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -46,7 +46,7 @@
     /**
      * All the sub-workflows as {@link Closure}s, keyed by their names.
      */
-    /*package*/ final Map<String,Closure> closures;
+    /*package*/ final transient Map<String,Closure> closures;
```
jglick: Why? `ParallelStep` was not even `Serializable`, nor should anything be saving it via XStream, so this makes no sense.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+        List<FlowNode> nodes = new DepthFirstScanner().allNodes(execution.getCurrentHeads());
+        Collections.sort(nodes, FlowScanningUtils.ID_ORDER_COMPARATOR);
+
+        Field nodeExecutionF = FlowNode.class.getDeclaredField("exec");
+        nodeExecutionF.setAccessible(true);
+
+        // Read each node via deserialization from storage, and sanity check the node, the actions, and the ArgumentsAction read back right
+        for (FlowNode f : nodes) {
+            XmlFile file = (XmlFile)(getFileM.invoke(storage, f.getId()));
```
jglick: Do we really need to do this at this level? Rather than getting `FlowNode`s from the execution using the normal API? If you are merely interested in making sure that metadata gets properly round-tripped through the storage system, that is much more easily (and realistically) checked using `RestartableJenkinsRule`.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -0,0 +1,311 @@
...
+    @ClassRule
+    public static BuildWatcher buildWatcher = new BuildWatcher();
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    /** Helper function to test direct file deserialization for an execution */
+    private void testDeserialize(FlowExecution execution) throws Exception {
+        if (!(execution instanceof CpsFlowExecution) || !(((CpsFlowExecution)execution).getStorage() instanceof SimpleXStreamFlowNodeStorage)) {
+            return;  // Test is unfortunately coupled to the implementation -- otherwise it will simply hit caches
```
jglick: I think this could be an assertion.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -0,0 +1,311 @@
...
+    @Test
+    public void testBasicCredentials() throws Exception {
+        String username = "bob";
+        String password = "s3cr3t";
+        UsernamePasswordCredentialsImpl c = new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, "test", "sample", username, password);
+        CredentialsProvider.lookupStores(r.jenkins).iterator().next().addCredentials(Domain.global(), c);
+
+        WorkflowJob job = r.jenkins.createProject(WorkflowJob.class, "credentialed");
+        job.setDefinition(new CpsFlowDefinition(
+                "node{ withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'test',\n" +
```
jglick: Can use symbols on these BTW.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+    @Test
+    public void testArgumentDescriptions() throws Exception {
+        WorkflowJob job = r.jenkins.createProject(WorkflowJob.class, "argumentDescription");
+        job.setDefinition(new CpsFlowDefinition(
+                "echo 'test' \n " +
+                        " node('master') { \n" +
+                        "   retry(3) {\n"+
+                        "     if (isUnix()) { \n" +
+                        "       sh 'whoami' \n" +
```
jglick: Is there a particular reason to be printing the user name here, rather than just some fixed message?

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/MonomorphicStep.java
```diff
@@ -39,7 +42,7 @@ public MonomorphicStep(MonomorphicData data) {
     }
 
     @Extension
-    public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+    public static final class DescriptorImpl extends AbstractStepDescriptorImpl  {
```
jglick: revert

jglick:
BTW test failure due to POM manipulation:

```
Jenkins Git plugin v3.0.5 failed to load.
 - Credentials Plugin v2.1.7 is older than required. To fix, install v2.1.8 or later.
```

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    public static Map<String,Object> sanitizedStepArguments(@Nonnull Map<String, Object> stepArguments, @CheckForNull EnvVars variables) {
+        HashMap<String, Object> output = new HashMap<String, Object>();
+
+        if (variables == null || variables.size() == 0) {
+            // No need to sanitize against environment variables
+            for (Map.Entry<String,Object> entry : stepArguments.entrySet()) {
+                if (entry.getValue() instanceof CharSequence && ((CharSequence)(entry.getValue())).length() > ArgumentsAction.MAX_STRING_LENGTH) {
+                    output.put(entry.getKey(), OVERSIZE_VALUE);
+                } else {
+                    output.put(entry.getKey(), entry.getValue());
```
jglick: May need to flatten `GString` to `String`: [JENKINS-43934](https://issues.jenkins-ci.org/browse/JENKINS-43934). Should be easily testable.

jglick: And as a defense in case something like that was forgotten, you might need to do something like https://github.com/jenkinsci/workflow-api-plugin/pull/18.

svanoort: I like that ProxyException solution -- was going to ask if you had a good approach to guard against unknown serialization issues, thanks.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: @stephenc Normally I'd agree, but there's a strong reason behind this, roughly captured in the comment.  This is explicitly a runtime property because I anticipated the risk of serialization or serialization-linked performance issues and wanted a workaround to solve them without a restart needed -- instafix for things like: https://issues.jenkins-ci.org/browse/JENKINS-43934

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
svanoort: @jglick Yes, this is covering a REAL class of significant issue I saw during testing -- serialization linked problems.  I want to ensure we have some coverage for this because the ArgumentsAction exposes more of this. 

 As proven by the bug you called my attention to this very afternoon, I was correct.  ;-)

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
svanoort: Not if you want to gracefully handle new implementations for storage without breaking the testcase.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
svanoort: We discussed this previously, I believe.  Where it cannot be serialized, it needs to be transient.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: That is why the field is not `final`—so you can modify it (say, from `/script`) without restarting Jenkins.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
jglick: I think if we swap out the storage we would like the test case to fail until we patch it to actually assert the corresponding conditions for the new storage.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
jglick: But who would be trying to serialize it? `DSL` is explicitly skipping `ParallelStep`.

svanoort:
@jglick We already discussed the approach used here in the sanitization and you agreed not to dictate implementation strategy simply because you prefer to take a different route.  Perhaps you have forgotten?

CC @recampbell who will also recall this, and that we explicitly set up a meeting to discuss this in order to avoid this exact thing happening.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: Was used in a testcase, and useful for testing.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: Well you cannot use that here, since the other PR was about a `ErrorAction` whose serial field is a `Throwable`, which is not the case here. But the other PR at least shows an example of explicitly checking in advance for classes which XStream will refuse to serialize.

(Normally all arguments are going to be of safe types as enumerated in the `DescribableModel` Javadoc. But users of pre-1.0 Workflow, and a few other special cases, may still be passing “live” `Describable`s. Generally those are also safe for XStream—that is the point of `Describable`—so perhaps there are no cases other than `GString`, and the bug is really that `DSL` should be doing the flattening, rendering [this](https://github.com/jenkinsci/structs-plugin/blob/a139f02a126515e2bb91208a67c535b44e52fbb8/plugin/src/main/java/org/jenkinsci/plugins/structs/describable/DescribableModel.java#L361-L363) dead code.)

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -179,6 +193,26 @@ protected Object invokeStep(StepDescriptor d, Object args) {
...
+                if (ps.namedArgs != null && !(ps.namedArgs.isEmpty()) && isKeepStepInfo() && !(s instanceof ParallelStep)) {
+                    Map<String, Object> actualArgs = ps.namedArgs;
+                    if (d.isMetaStep() && ps.namedArgs.get("delegate") instanceof Map) {
+                        actualArgs = (Map<String,Object>)ps.namedArgs.get("delegate");
+                    }
+                    // Get the environment variables to find ones that might be credentials bindings
+                    Computer comp = context.get(Computer.class);
+                    EnvVars allEnv = new EnvVars(context.get(EnvVars.class));
+                    if (comp != null && allEnv != null) {
+                        allEnv.entrySet().removeAll(comp.getEnvironment().entrySet());
```
svanoort: @jglick We already discussed this.   To refresh your memory, we agreed that yes, it would be nice to distinguish between env variables sources, but there is not a present mechanism for this. 

Adding such a mechanism would be an excellent future enhancement, but is completely out of scope for this story. By all means, a possible future enhancement.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: @stephenc Also this is a clear :ant:  and not a :bug:

jglick:
> We already discussed the approach used here in the sanitization and you agreed not to dictate implementation strategy simply because you prefer to take a different route.

I explicitly stated that I do not consider the current sanitization code a blocker for merge, though after seeing the actual code for the first time it seems even hairier than I expected and would recommend a simpler and more reliable approach be taken.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: Yeah, you're right.  This was a remnant from some logic around nulls specifically but can be simplified.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
svanoort: Conversely: why would it be NON-transient if unserializable?   If there is not a solid reason than this is a tiny incidental bugfix, and I do not see why that is cause for complaint rather than praise.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
svanoort: Is there a particular reason why this matters?  Something specific that is *explicitly wrong* with it, not just something that you prefer to do differently?

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: Agreed with @svanoort - that's a nice goal but orthogonal to this PR.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -98,6 +104,14 @@ protected Object readResolve() throws IOException {
         return this;
     }
 
+    /** Allows user to disable saving step info, if this imposes too high a performance overhead
+     *   by setting System Property org.jenkinsci.plugins.workflow.cps.DSL.keepStepInfo to 'false'.
+     *  This may be modified at runtime by the user if needed for debugging. */
+    private static boolean isKeepStepInfo() {
+        String prop = System.getProperty(DSL.class.getName()+".keepStepInfo");
+        return (StringUtils.isEmpty(prop)|| Boolean.parseBoolean(prop));
```
abayer: :ant: `return Boolean.parseBoolean(prop, "true");` would be simpler and achieve the same result, right?

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -179,6 +193,26 @@ protected Object invokeStep(StepDescriptor d, Object args) {
             d.checkContextAvailability(context);
             Thread.currentThread().setContextClassLoader(CpsVmExecutorService.ORIGINAL_CONTEXT_CLASS_LOADER.get());
             s = d.newInstance(ps.namedArgs);
+            try {
+                // No point storing empty arguments, and ParallelStep is a special case where we can't store its closure arguments
+                if (ps.namedArgs != null && !(ps.namedArgs.isEmpty()) && isKeepStepInfo() && !(s instanceof ParallelStep)) {
+                    Map<String, Object> actualArgs = ps.namedArgs;
+                    if (d.isMetaStep() && ps.namedArgs.get("delegate") instanceof Map) {
+                        actualArgs = (Map<String,Object>)ps.namedArgs.get("delegate");
```
abayer: I honestly can't remember - where is `delegate` coming from?

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -179,6 +193,26 @@ protected Object invokeStep(StepDescriptor d, Object args) {
...
+                    }
+                    // Get the environment variables to find ones that might be credentials bindings
+                    Computer comp = context.get(Computer.class);
+                    EnvVars allEnv = new EnvVars(context.get(EnvVars.class));
+                    if (comp != null && allEnv != null) {
```
abayer: `allEnv` couldn't ever be null, could it? Empty, sure, but not null, since we just initialized it a line above. =)

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+
+import static org.jenkinsci.plugins.workflow.actions.ArgumentsAction.NotStoredReason.MASKED_VALUE;
+import static org.jenkinsci.plugins.workflow.actions.ArgumentsAction.NotStoredReason.OVERSIZE_VALUE;
+
+/**
+ * Implements {@link ArgumentsAction} by storing step arguments, with sanitization
+ */
+public class ArgumentsActionImpl extends ArgumentsAction {
+
+    /** Arguments to the step, for cases where we cannot simply store the step because masking was applied */
```
abayer: Feel completely free to disregard, but is there a reason for the one-liner javadoc rather than the more common at-least-three-liners? I'm curious. =)

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
abayer: Please no - excessive `private` scope for things like this has resulted in me having to recreate behavior elsewhere for no good reason. I see absolutely no reason this shouldn't be public.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    /** See if sensitive environment variable content is in a string */
+    public static boolean isStringSafe(@CheckForNull String input, @CheckForNull EnvVars variables, @Nonnull Set<String> safeEnvVariables) {
+        if (input == null || variables == null || variables.size() == 0) {
+            return true;
+        }
+        StringBuilder pattern = new StringBuilder();
+        int count = 0;
+        for (Map.Entry<String,String> ent : variables.entrySet()) {
+            String val = ent.getValue();
+            if (val == null || val.isEmpty() || safeEnvVariables.contains(ent.getKey())) {  // Skip values that are safe
```
abayer: Maybe `StringUtils.isEmpty(val)` instead of `val == null || val.isEmpty()`? I always like cutting down on `||`s.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+            "JENKINS_SERVER_COOKIE",
+            "JENKINS_URL",
+            "JOB_BASE_NAME",
+            "JOB_NAME",
+            "JOB_URL",
+            "NODE_LABELS",
+            "NODE_NAME",
+            "WORKSPACE",
+
+            // Normal system variables for posix environments
```
abayer: Since we're hardcoded-whitelisting, let's try to get the equivalent for Windows land in here too?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+     * @return
+     */
+    @Nonnull
+    public static Map<String,Object> sanitizedStepArguments(@Nonnull Map<String, Object> stepArguments, @CheckForNull EnvVars variables) {
+        HashMap<String, Object> output = new HashMap<String, Object>();
+
+        if (variables == null || variables.size() == 0) {
+            // No need to sanitize against environment variables
+            for (Map.Entry<String,Object> entry : stepArguments.entrySet()) {
+                if (entry.getValue() instanceof CharSequence && ((CharSequence)(entry.getValue())).length() > ArgumentsAction.MAX_STRING_LENGTH) {
```
abayer: Itty-bitty-nit - I'm confused by the parentheses in `((CharSequence)(entry.getValue()))` - there's a redundant pair around `entry.getValue()`, I think?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                output.put(param.getKey(), val);
+            }
+        }
+        return output;
+    }
+
+    @Nonnull
+    @Override
+    protected Map<String, Object> getArgumentsInternal() {
+        return (arguments == null)? Collections.EMPTY_MAP : arguments;
```
abayer: Another tiny nit - parens are not needed here, would love a space before the `?`, but yeah, obviously not a blocker.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
abayer: more specifically, this and `isFullArguments()` are identical, so there should really only be one of them. I'd tend to vote for this one since it matches the variable it's returning.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
abayer: ...do we really care about pre-1.0 at this point when we're making changes to a core part of the suite like this?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: This.  I am not locking a useful API behind private for exactly the reasons abayer states.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: IIRC you actually suggested this @jglick

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: For handling of live `Describable`s, yes. While it is generally discouraged as of 1.0, there are some people doing it, for example to copy-and-modify `scm`.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: If and when you find a need to make something `public` (or `protected`), it takes little effort to file an upstream PR proposing that change, paired with a downstream PR demonstrating why it is necessary. But once you open it up, you are on the hook to support it indefinitely. So I stand by this comment.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: I do not see how that is orthogonal. This PR is adding considerable technical debt, introducing a lot of complex code whose correctness is far from clear, apparently to avoid making a very simple change in `credentials-binding`.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: It is the property name used by, for example, `CoreStep`.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: Yes, they will become the same thing -- this was an oversight in refactoring.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: @jglick If you think it's that high a priority and simple to implement, I'll be happy to review your PR accomplishing that since it sounds like you have a clear idea how to proceed.  

As it stands I do not know how to accomplish this, and unfortunately do not have time to take on extra scope while others are waiting for this to be released.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: Aaaaah, gotcha.

svanoort: @jglick One can simply look at the FlowNode/descriptor and see that we have unwrapped a MetaStep.  Failing to unwrap it makes it more work to consume this API because it forces the API consumer to do these inspections to figure out how to get what we actually care about (the arguments given). 

As it stands, we only have CoreStep and CoreWrapperStep that are making use of the metastep approach, and they share the delegate parameter.  AFAIK we do not have any other planned implementations using this besides testcases.   

Based on the above it would be extremely foolish to add new-new APIs to support a use case that does not exist and AFAIK will never exist. 

Instead what I can do is harden this slightly -- we can use existing StepDescriptor methods to confirm we only have a single argument, and then unwrap that (otherwise persisting everything intact).  

In the purely *hypothetical* case that we add another metastep, which has a different delegate parameter name *and* takes another argument then it will be handled seamlessly.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
jglick: Because use of the `transient` keyword should be limited to fields which you specifically wish to exclude from serialization in a class which you do expect to be serialized. Its presence here is misleading and therefore should be removed.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
jglick: Just trying to ascertain if the test could be made simpler. A plain `echo` would allow five lines to be collapsed to one.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: `GenericSCMStep` is expected to become a metastep soon, with a different “delegate” parameter name and additional parameters, so this is not hypothetical.

abayer: Why would it have a different parameter name than the other metasteps?

jglick: Because it is already `scm`.

abayer: Touché, sir. =)

jglick: In a nutshell, the idea is that if we can do [JENKINS-37227](https://issues.jenkins-ci.org/browse/JENKINS-37227) as planned,

```groovy
git 'http://mygit/'
```

will become sugar for

```groovy
checkout(scm: [$class: 'GitSCM', urlOrSomething: 'http://mygit/'])
```

and you would expect `argumentsToString` in this case to return `http://mygit/`, just as it would now with the (crappy) first-class step.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
stephenc:
```
    public static String getProperty(String key) {
        checkKey(key);
        SecurityManager sm = getSecurityManager();
        if (sm != null) {
            sm.checkPropertyAccess(key);
        }

        return props.getProperty(key);
    }
```

vs static non-final field access for code that you have called out as performance critical.

Plus *not following existing patterns* which will significantly increase supportability costs.

For those two reasons I say this is decreasing the quality of the code which makes it a :bug:

abayer:
@svanoort Do you feel like you've got enough information to be able to proceed?

svanoort:
@stephenc I see your point about the potential performance of the property lookup.  However:

> not following existing patterns which will significantly increase supportability costs.

This is a pattern we have used before, for properties that can be modified at runtime.  It's also a straightforward idiom.  

If you can provide a source/benchmark showing that the cost of property lookups at runtime is expensive enough to be an issue on its own I'm happy to switch to a static field with getters/setters -- but without evidence saying otherwise it's likely that the performance impact of a property lookup will disappear next to the cost of the reflective serialization for actions.  

Not saying you're wrong here -- you have extensive knowledge in the performance area -- just saying that this needs some evidence behind it.  And that evidence is useful for future learning opportunities, and for other code decisions.

Yes, I'm aware that in the above context my short-circuit check of null in one of the changes was even more clearly a premature optimization.

svanoort:
@abayer No, we're still somewhat stuck here.  We're making progress and there's things that can be worked on, but ultimately still points that will need to be settled for this set of changes to move forward.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: With this in mind, I'll change the implementation here if there's not just a single parameter (as described above) which will handle this case -- though the StepDescriptor of course need to provide its own argumentsToString (it would anyway).

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: #123 should handle it I think, since that takes effect before you grab data for `ArgumentsActionImpl`.

stephenc:
@svanoort primarily it is a *supportability cost* to follow a different pattern from 95%+ (I suspect 99%+) of Jenkins.

The other cases you allude to, by any chance did you write them that way? (not assigning blame here, rather trying to find out who to educate)

The performance difference depends on the calling pattern. The impact for the normal case (or when the switch is off) should be negligible as there will be IO on the thread immediately after. The impact on the exception case is where one could see a difference as looking up the hash map (esp if running in a servlet container that installs a security manager - hello Tomcat - which would add to the cost) will involve memory cache line pollution... given that one presumes that the switch is on because of CPU resources being under pressure... excessive *avoidable* cacheline pollution should be avoided.

But the real cost is that rather than being able to follow the existing pattern of writing to the field via groovy, we have to tell users "oh this one is different, just set the system property (assuming your running in a servlet container that lets you set system properties) and it will be picked up immediately... but forget that for all the other system properties that Jenkins uses... for them write to the field" which is why I am objecting. Adding unnecessary complexity to supportability is the issue

jglick:
> Please follow existing patterns for runtime mutable settings

seconded

> I would recommend getting very liberal with the @Restricted annotations

seconded

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: Unfortunately there is no such method :(

stephenc:
```
private static boolean KEEP_STEP_INFO = Boolean.parseBoolean(System.getProperty(DSL.class.getName() + ".keepStepInfo", "true"));
```

and then change this to `return KEEP_STEP_INFO;`

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -104,12 +106,15 @@ protected Object readResolve() throws IOException {
...
-        String prop = System.getProperty(DSL.class.getName()+".keepStepInfo");
-        return (StringUtils.isEmpty(prop)|| Boolean.parseBoolean(prop));
+    private static final String KEEP_STEP_ARGUMENTS_PROPERTYNAME = (DSL.class.getName()+".keepStepArguments");
+
+    private static boolean isKeepStepArguments = StringUtils.isEmpty(System.getProperty(KEEP_STEP_ARGUMENTS_PROPERTYNAME))
+            || Boolean.parseBoolean(System.getProperty(KEEP_STEP_ARGUMENTS_PROPERTYNAME));
+
+    /** Tell us if we should store {@link Step} arguments in an {@link org.jenkinsci.plugins.workflow.actions.ArgumentsAction}
+     *  or simply discard them (if set to false, explicitly) */
+    public static boolean isKeepStepArguments() {
```
jglick: Why is this public? Why does it exist at all?

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -147,24 +151,27 @@ public void testBasicCreateAndMask() throws Exception {
         Map<String, Object> arguments = new HashMap<String,Object>();
         arguments.put("message", "I have a secret p4ssw0rd");
 
+        Field maxSizeF = ArgumentsAction.class.getDeclaredField("MAX_RETAINED_LENGTH");
+        maxSizeF.setAccessible(true);
+        int maxSize = maxSizeF.getInt(null);
```
jglick: Oof. Would be better just to have that in this plugin to begin with.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        for (Map.Entry<String,?> param : stepArguments.entrySet()) {
+            Object val = param.getValue();
+            if (ArgumentsAction.isOversized(val)) {
+                output.put(param.getKey(), OVERSIZE_VALUE);
+            } else if (val instanceof String && !isStringSafe((String)val, variables, SAFE_ENVIRONMENT_VARIABLES)) {
+                output.put(param.getKey(), MASKED_VALUE);
+            } else {
+                output.put(param.getKey(), val);
```
jglick: This will not handle string in nested `Describable`s AFAICT. You need to recurse.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
svanoort: It's because it's a test class.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -55,24 +57,23 @@
...
     }
 
     /** Create a step, sanitizing strings for secured content */
     public ArgumentsActionImpl(@Nonnull Map<String, Object> stepArguments) {
         this(stepArguments, new EnvVars());
     }
 
+    /** For testing use only */
+    ArgumentsActionImpl(){
+        this.isUnmodifiedBySanitization = false;
```
jglick: Odd; how could it be modified if it has no arguments at all?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -188,35 +189,105 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
     ));
 
     /**
-     * Does first-level sanitization of a step's params, returning the Describable arguments map,
+     * Sanitize a list recursively
+     */
+    @CheckForNull
+    Object sanitizeListAndRecordMutation(@Nonnull List objects, @CheckForNull EnvVars variables) {
```
jglick: avoid rawtypes

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -188,35 +189,105 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
...
+        return (isMutated) ? output : objects; // Throw away copies and use originals wherever possible
+    }
+
+    /** Sanitize a single non-Map, non-List, non-{@link Step}/non-{@link UninstantiatedDescribable} item */
+    @CheckForNull
+    Object sanitizeObjectAndRecordMutation(@CheckForNull Object o, @CheckForNull EnvVars vars) {
+        // Package scoped so we can test it directly
+        Object tempVal = o;
+        if (tempVal instanceof Step) {
+            // Ugly but functional used for legacy syntaxes with metasteps
```
jglick: Huh? Under what circumstances could this happen? Even legacy metastep syntax

```groovy
step(new JUnitResultArchiver(…))
```

would be passing some `Describable`, but not a `Step`.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -188,35 +189,105 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
...
+     * Returns original input if not modified.
+     * Due to restrictions in {@link org.jenkinsci.plugins.structs.describable.DescribableModel#instantiate(Map)}
+     *  we do not need to sanitize keys, which can only parameters to a {@link Describable}
      * @param stepArguments Arguments provided when creating the step
      * @param variables Environment variables to remove
      * @return Arguments map, with oversized values and values containing bound credentials stripped out.
      */
     @Nonnull
-    public static Map<String,Object> sanitizedStepArguments(@Nonnull Map<String, Object> stepArguments, @CheckForNull EnvVars variables) {
+    Map<String,Object> sanitizeMapAndRecordMutation(@Nonnull Map<String, Object> stepArguments, @CheckForNull EnvVars variables) {
```
jglick: The name `stepArguments` is misleading, because these could be arguments to some nested `Describable`.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -0,0 +1,303 @@
...
+    ));
+
+    /**
+     * Sanitize a list recursively
+     */
+    @CheckForNull
+    Object sanitizeListAndRecordMutation(@Nonnull List objects, @CheckForNull EnvVars variables) {
+        // Package scoped so we can test it directly
+
+        if (isOversized(objects)) {
```
jglick: Does this even still work? I thought it only applies to elements now.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: @jglick While groovy can be used to mutate this since it ignores access permissions, we don't want to expose the field to normal mutation from java -- but we do want to be able to read it potentially.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    @Test
+    public void testUnusualStepInstantiations() throws Exception {
+        WorkflowJob job = r.jenkins.createProject(WorkflowJob.class, "unusualInstantiation");
+        job.setDefinition(new CpsFlowDefinition(
+                " node('master') { \n" +
+                "   writeFile text: 'hello world', file: 'msg.out'\n" +
+                "   step([$class: 'ArtifactArchiver', artifacts: 'msg.out', fingerprint: false])\n "+
+                "   withEnv(['CUSTOM=val']) {\n"+  //Symbol-based, because withEnv is a metastep
```
jglick: Huh? `withEnv` is not a metastep.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
jglick: But last I checked, the corresponding field is only accessed from this class anyway.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: In some cases it's a duplicate check, but that is okay.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
stephenc: Advise a `@Restricted(NoExternalUse.class)` unless you have use cases for reading it from actual Java code

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: There's no way to infer typing here AFAIK.

pom.xml
```diff
@@ -150,6 +150,12 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>credentials-binding</artifactId>
```
jglick: I think this is what caused the failure which I am fixing in #135.

PR #99, opened 2017-01-27
Fixing a flaky test

jglick:
Follows up #85.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsThreadTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsThreadTest.java
@@ -67,7 +67,8 @@ public class CpsThreadTest {
         List<CauseOfInterruption> causes = iba.getCauses();
         assertEquals(1, causes.size());
         assertEquals(CauseOfInterruption.UserInterruption.class, causes.get(0).getClass());
-        r.assertLogContains("never going to stop", b); // TODO under heavy load sometimes fails here; why?
+        r.waitForMessage("Finished: ABORTED", b); // TODO WorkflowRun.isBuilding() can go to false before .finish has completed
+        r.assertLogContains("never going to stop", b);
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsThreadTest.java
```diff
@@ -67,7 +67,8 @@
         List<CauseOfInterruption> causes = iba.getCauses();
         assertEquals(1, causes.size());
         assertEquals(CauseOfInterruption.UserInterruption.class, causes.get(0).getClass());
-        r.assertLogContains("never going to stop", b); // TODO under heavy load sometimes fails here; why?
+        r.waitForMessage("Finished: ABORTED", b); // TODO WorkflowRun.isBuilding() can go to false before .finish has completed
```
bitwiseman: Instead of a TODO, should there be a JIRA for this? I assume this bad product behavior.

jglick: I do not know of any impact on users—we only see it because `waitForCompletion` goes into a fast spin loop and so might proceed with the test during the instant before the log is closed. But I will file a bug for it just in case.

jglick: Scratch that, I am not really sure I know why this flakes out. `waitForCompletion` should have been waiting for `completed` to go to `true`, which should only happen after the log is closed. Perhaps there is some race condition in the log writing vs. reading.

At any rate, `waitForMessage` is useful here since the `Finished` message is printed after the `AbortException.message`.

bitwiseman: @jglick - Either way, you'll file a bug on, right?  Thanks.
That was my only point - that the TODO is noting behavior that is a bug.

PR #101, opened 2017-02-01
Ensure CpsVmExecutorService.reportProblem causes the build to end promptly

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -616,10 +616,16 @@ public class CpsFlowExecution extends FlowExecution {
             }
             @Override public void onFailure(Throwable t) {
                 LOGGER.log(Level.WARNING, "Failed to set program failure on " + owner, t);
-                onProgramEnd(new Outcome(null, t));
-                cleanUpHeap();
+                croak(t);
             }
         });
+    }
+
+    /** Report a fatal error in the VM. */
+    void croak(Throwable t) {
+        setResult(Result.FAILURE);
+        onProgramEnd(new Outcome(null, t));
+        cleanUpHeap();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -616,12 +616,18 @@ private void loadProgramFailed(final Throwable problem, SettableFuture<CpsThread
...
-                onProgramEnd(new Outcome(null, t));
-                cleanUpHeap();
+                croak(t);
             }
         });
     }
 
+    /** Report a fatal error in the VM. */
+    void croak(Throwable t) {
+        setResult(Result.FAILURE);
```
jglick: Previously forgot to set build status.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -916,11 +922,11 @@ public boolean isComplete() {
...
-    /*packgage*/ synchronized void onProgramEnd(Outcome outcome) {
-        // end of the program
-        // run till the end successfully FIXME: failure comes here, too
-        // TODO: if program terminates with exception, we need to record it
-        // TODO: in the error case, we have to close all the open nodes
+    /**
+     * Record the end of the build.
+     * @param outcome success; or a normal failure (uncaught exception); or a fatal error in VM machinery
+     */
+    synchronized void onProgramEnd(Outcome outcome) {
```
jglick: Long-obsolete comments.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -50,7 +50,8 @@ public void run() {
      * That makes it worth reporting.
      */
     private void reportProblem(Throwable t) {
-        LOGGER.log(WARNING, "Unexpected exception in CPS VM thread: "+cpsThreadGroup.getExecution().toString(), t);
+        LOGGER.log(WARNING, "Unexpected exception in CPS VM thread: " + cpsThreadGroup.getExecution(), t);
+        cpsThreadGroup.getExecution().croak(t);
```
jglick: The main fix.

PR #102, opened 2017-02-05
[JENKINS-32986] Apply a timeout to some hang-prone operations in the CPS VM thread

jglick:
Begins to address [JENKINS-32986](https://issues.jenkins-ci.org/browse/JENKINS-32986). Does not go so far as to impose a hard limit. #93 probably required for any sort of quota.

Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/29.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
@@ -38,11 +38,13 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.*;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
 import org.jenkinsci.plugins.workflow.support.concurrent.Futures;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -156,7 +158,7 @@ public StepExecution getStep() {
         final CpsThread old = CURRENT.get();
         CURRENT.set(this);
 
-        try {
+        try (Timeout timeout = Timeout.limit(5, TimeUnit.MINUTES)) {
```
svanoort: Safe to do because any long-running step should be async anyway, and if your next chunk of groovy needs 5 minutes to execution something is horribly, horribly wrong. 

Note: potential for breakage here if there's a valid case where that could happen without user error (or users do something wrong frequently and unknowingly).  If that sparks any sudden concerns, consider another approach.

jglick:
Well, the “valid” use case would be something like

```groovy
def status = new java.net.URL('http://ultraslow.net/api/status').openConnection().responseCode
```

where the server takes six minutes to respond. (During which time other `parallel` branches cannot progress, `input`s will time out, etc.) I am happy to break this script. Things should not explode; if the blocking socket call is interruptible (and last I checked, it was not), then you will get an `InterruptedException` or an `InterruptedIOException` or whatever.

PR #104, opened 2017-02-07
Make StepNode just extend higher-level API in workflow-api

svanoort:
Allows compatibility while permitting workflow-api to expose step info. 

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/28

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.6</version>
+            <!-- Temporary version to pull in upstream PRs, will revert before merge once that is released -->
+            <version>2.9-stepnode-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.9</version>
+            <version>2.10</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -75,7 +75,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.6</version>
+            <!-- Temporary version to pull in upstream PRs, will revert before merge once that is released -->
+            <version>2.9-stepnode-SNAPSHOT</version>
```
jglick: I find it helpful to link to the upstream PR in a comment.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepNode.java
```diff
@@ -7,16 +7,9 @@
 import javax.annotation.CheckForNull;
 
 /**
- * Optional interface for {@link FlowNode} that has associated {@link StepDescriptor}
- *
- * @author Kohsuke Kawaguchi
+ * Optional interface for {@link FlowNode} that has associated {@link StepDescriptor}.
+ * Retained for binary & API compatibility but the interface has been pushed up a level into the
+ *  workflow-api plugin to allow broader access.
```
jglick: Just mark it `@Deprecated`.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepNode.java
```diff
@@ -1,22 +1,18 @@
...
  */
-public interface StepNode {
-    /**
-     * Returns the descriptor for {@link Step} that produced this flow node.
-     *
-     * @return
-     *      null for example if the descriptor that created the node has since been uninstalled.
-     */
-    @CheckForNull StepDescriptor getDescriptor();
+@Deprecated
```
christ66: :NIT: would be nice to java javadoc suggesting to use the workflow-api interface instead.

christ66:
:bee:

christ66:
:bee:

PR #106, opened 2017-02-13
[JENKINS-42499] Removing caching from EnvActionImpl

jglick:
[JENKINS-42499](https://issues.jenkins-ci.org/browse/JENKINS-42499)

Known to cause problems under some circumstances, when plugins dynamically attach actions that affect the environment. For example, some usages of https://github.com/jenkinsci/jenkins/pull/2975 + https://github.com/jenkinsci/jenkins/pull/2993 would not work correctly without this. (But this is independent of that API—also applies to existing APIs like `EnvironmentContributor`.)

- [x] test
- [X] fix

@reviewbybees


svanoort:
@jglick Is this PR dead?

jglick:
Just stalled. Do you need this for anything?

svanoort:
No, just wondering since it's been open a while.

svanoort:
(Though I would prefer to have a second reviewer before merge since the EnvVars providers are like wizards -- subtle and quick to anger -- so another set of eyes may catch something extra).

jglick:
@reviewbybees done

PR #107, opened 2017-02-13
[JENKINS-41945] Ignore null elements in ClassInfo.globalClassSet.items

jglick:
[JENKINS-41945](https://issues.jenkins-ci.org/browse/JENKINS-41945)

Apparent race condition involving garbage collection timing in Groovy 2.x (thus Jenkins 2.x). `ManagedLinkedList.Iter.next` is in fact documented to return null sometimes (unlike, say, `WeakHashMap.KeyIterator.next`), so we need to be prepared for that.

@reviewbybees


recena:
🐝

PR #108, opened 2017-02-17
iotaStr incorrectly skipped every other number after around 500

jglick:
Flaw in #82.

@reviewbybees


jglick:
BTW `plugin-compat-tester` actually caught this in `ScalabilityTest` but for whatever reason @recena & crew did not notify me or follow up.

PR #109, opened 2017-02-23
[JENKINS-42189] Fix ClassInfo cleanup in Groovy 2.4.8+

jglick:
[JENKINS-42189](https://issues.jenkins-ci.org/browse/JENKINS-42189)

Directly solves the soft leak, and probably also solves the deadlock (without waiting for https://github.com/apache/groovy/pull/489 in a 2.4.9 update) by simply skipping the affected code path.

Amends #83. Required by https://github.com/apache/groovy/pull/219 and https://github.com/apache/groovy/pull/427.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -107,6 +107,7 @@ import hudson.security.ACL;
 import hudson.security.AccessControlled;
 import java.beans.Introspector;
 import java.lang.ref.Reference;
+import java.lang.ref.WeakReference;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1027,6 +1040,10 @@ private static void cleanUpGlobalClassSet(@Nonnull Class<?> clazz) throws Except
             globalClassSet.getClass().getMethod("remove", Object.class).invoke(globalClassSet, clazz); // like Map but not
             LOGGER.log(Level.FINER, "cleaning up {0} from GlobalClassSet", clazz.getName());
         } catch (NoSuchMethodException x) { // Groovy 2
+            try {
+                Field classRefF = classInfoC.getDeclaredField("classRef");
+                return; // 2.4.8+, nothing to do here (classRef is weak anyway)
+            } catch (NoSuchFieldException x2) {} // 2.4.7-
```
oleg-nenashev: 2.47 -

oleg-nenashev: NVM

jglick: No, Groovy 2.4.7.

jglick:
> if it got enough manual testing

It got automated testing, in `CpsFlowExecutionTest.loaderReleased`. Prior to the fix, this test does pass against `jenkins.version=2.47`, but only after forcing collection of soft references. (Currently `assertGC` tolerates soft leaks. I could perhaps amend it to fail if regular GC cycles do not suffice to clear everything.)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -996,19 +997,31 @@ private static void cleanUpGlobalClassValue(@Nonnull ClassLoader loader) throws
         Object map = mapF.get(globalClassValue);
         Class<?> groovyClassValuePreJava7Map = Class.forName("org.codehaus.groovy.reflection.GroovyClassValuePreJava7$GroovyClassValuePreJava7Map");
         Collection entries = (Collection) groovyClassValuePreJava7Map.getMethod("values").invoke(map);
-        Field klazzF = classInfoC.getDeclaredField("klazz");
-        klazzF.setAccessible(true);
         Method removeM = groovyClassValuePreJava7Map.getMethod("remove", Object.class);
```
svanoort: Suggest using MethodHandle over raw Method -- as fast as Method in the worst case, an order of magnitude faster in the best:  http://andrewtill.blogspot.com/2011/08/using-method-handles.html

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -996,19 +997,31 @@ private static void cleanUpGlobalClassValue(@Nonnull ClassLoader loader) throws
...
         Class<?> entryC = Class.forName("org.codehaus.groovy.util.AbstractConcurrentMapBase$Entry");
         Method getValueM = entryC.getMethod("getValue");
```
svanoort: Suggest using MethodHandle over Method here.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Could be a nice refinement to cleanup code later. In this case the number of accesses is fairly small so the cost is probably insignificant compared to other aspects of the build. Anyway out of scope for this PR.

svanoort: Unless I misunderstand, it's potentially running for every class with a reference held to it -- I do not think that qualifies as a "small" number

PR #110, opened 2017-02-24
Jenkinsfile

jglick:
And generally updating test infrastructure.

@reviewbybees


src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
```diff
@@ -184,13 +101,6 @@ public void missingContextCheck() throws Exception {
         jenkins.assertBuildStatus(Result.FAILURE, b);
     }
 
-    @Test
-    public void addAction() throws Exception {
```
jglick: See https://github.com/jenkinsci/workflow-job-plugin/pull/39.

jglick:
Waiting for @rtyler to update ci.jenkins.io plugins so that @stephenc’s rate limit fix is live and this PR actually gets built via `Jenkinsfile`.

jglick:
[npm failure](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-110/2/execution/node/64/log/), random?

PR #115, opened 2017-03-08
[JENKINS-42561] Add a param to the timeout on some hang-prone operations in the CPS Vm thread

lucasam:
[JENKINS-42561](http://issues.jenkins-ci.org/browse/JENKINS-42561)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
@@ -24,25 +24,24 @@
 
 package org.jenkinsci.plugins.workflow.cps;
 
+import static java.util.logging.Level.FINE;
+import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.PROGRAM;
+
 import com.cloudbees.groovy.cps.Continuable;
 import com.cloudbees.groovy.cps.Outcome;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.SettableFuture;
-import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
-import org.jenkinsci.plugins.workflow.steps.StepExecution;
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.io.Serializable;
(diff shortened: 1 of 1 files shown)


jglick:
How does this relate to #114?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -314,4 +314,13 @@ public static CpsThread current() {
...
         return "Thread #" + id + String.format(" @%h", this);
     }
+
+    /**
+     * Gets a the timeout collect from a System property or default value.
+     * Use -Dorg.jenkinsci.plugins.workflow.cps.CpsThread.runTimeoutMinutes
+     * @return
+     */
+    public long getRunTimeoutMinutes() {
+        return Long.parseLong(System.getProperty(getClass().getName() + ".runTimeoutMinutes", "5"));
```
jglick: `Long.getLong(String, long)` is what you meant.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -314,4 +314,13 @@ public static CpsThread current() {
         // getExecution().getOwner() would be useful but seems problematic.
         return "Thread #" + id + String.format(" @%h", this);
     }
+
+    /**
+     * Gets a the timeout collect from a System property or default value.
+     * Use -Dorg.jenkinsci.plugins.workflow.cps.CpsThread.runTimeoutMinutes
+     * @return
+     */
+    public long getRunTimeoutMinutes() {
```
jglick: Typically such tuning constants are `public static /* not final */` fields, for ease of manipulation from `/script` etc.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
lucasam: it wasn't. But now it is :+1:

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
lucasam: changed to a class variable

PR #116, opened 2017-03-20
[JENKINS-31967] Specifying decimal numbers for JUnit health factor in Pipeline snippets results in invalid code

artem-fedorov:
https://issues.jenkins-ci.org/browse/JENKINS-31967


jglick:
The `Snippetizer` fix is only cosmetic. The actual fix needs to be done in `DescribableModel` in `structs`.

jglick:
Could probably pick up a snapshot dependency on https://github.com/jenkinsci/structs-plugin/issues/16 to do a proper integration test here.

artem-fedorov:
Can tell how I should do it?

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -102,7 +102,9 @@ static StringBuilder object2Groovy(StringBuilder b, Object o, boolean nestedExp)
             return b;
         }
 
-        if (clazz == Boolean.class || clazz == Integer.class || clazz == Long.class) {
+        if (clazz == Boolean.class || clazz == Integer.class || clazz == Long.class ||
+                clazz == Float.class || clazz == Double.class ||
+                clazz == Byte.class || clazz == Short.class) {
```
jglick: IIRC there is some utility in Commons Lang for this.

jglick:
> Can tell how I should do it?

Just change the dependency version on `structs` to `1.7-20170522.183630-2`, which I just `deploy`ed with your PR.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -66,6 +66,7 @@
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
+import org.springframework.util.ClassUtils;
```
jglick: Mm, rather not use `springframework` please. Better just revert this last change.

jglick:
> Can tell how I should do it?

If you want to do this, you need to not just have a _dependency_ on new `structs` but an actual test relying on it (probably in `SnippetizerTest`). I think we already have a suitable `test`-scoped dependency on `junit`.

pom.xml
```diff
@@ -103,7 +103,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.6</version>
+            <version>1.7-20170522.183630-2</version>
```
jglick: Released; use 1.7.

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: (forget it)

PR #120, opened 2017-04-19
[JENKINS-43019] Suppressing GroovyClassLoader.InnerLoader fixes stuff

jglick:
[JENKINS-43019](https://issues.jenkins-ci.org/browse/JENKINS-43019)

You just do not want to know.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
@@ -40,6 +40,7 @@ class CpsGroovyShell extends GroovyShell {
         super(parent,new Binding(),cc);
         this.execution = execution;
         try {
+            // Apparently there is no supported way to initialize a GroovyShell with a specified GroovyClassLoader.
             Field loaderF = GroovyShell.class.getDeclaredField("loader");
             loaderF.setAccessible(true);
             loaderF.set(this, new CleanGroovyClassLoader(parent, cc));
@@ -51,6 +52,8 @@ class CpsGroovyShell extends GroovyShell {
     /**
      * Disables the weird and unreliable {@link groovy.lang.GroovyClassLoader.InnerLoader}.
      * This is apparently only necessary when you are using class recompilation, which we are not.
+     * We want the {@linkplain Class#getClassLoader defining loader} of {@code *.groovy} to be this one.
+     * Otherwise the defining loader will be an {@code InnerLoader}, and not necessarily the same instance from load to load.
      * @see GroovyClassLoader#getTimeStamp
      */
     private static final class CleanGroovyClassLoader extends GroovyClassLoader {
@@ -60,12 +63,14 @@ class CpsGroovyShell extends GroovyShell {
         }
 
         @Override protected ClassCollector createCollector(CompilationUnit unit, SourceUnit su) {
+            // Super implementation is what creates the InnerLoader.
             return new CleanClassCollector(unit, su);
         }
 
         private final class CleanClassCollector extends ClassCollector {
 
             CleanClassCollector(CompilationUnit unit, SourceUnit su) {
+                // Cannot override {@code final cl} field so have to do it this way.
                 super(null, unit, su);
             }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
@@ -1,23 +1,29 @@
 package org.jenkinsci.plugins.workflow.cps;
 
 import groovy.lang.Binding;
+import groovy.lang.GroovyClassLoader;
 import groovy.lang.GroovyCodeSource;
 import groovy.lang.GroovyShell;
 import groovy.lang.Script;
-import org.codehaus.groovy.control.CompilationFailedException;
-import org.codehaus.groovy.control.CompilerConfiguration;
-
-import javax.annotation.CheckForNull;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


jglick:
Build failures look like npm problems. Cf. [INFRA-1139](https://issues.jenkins-ci.org/browse/INFRA-1139).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
```diff
@@ -28,9 +35,51 @@
     /**
      * Use {@link CpsGroovyShellFactory} to instantiate it.
      */
+    @SuppressFBWarnings(value="DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED", justification="irrelevant")
     CpsGroovyShell(ClassLoader parent, @CheckForNull CpsFlowExecution execution, CompilerConfiguration cc) {
         super(parent,new Binding(),cc);
         this.execution = execution;
+        try {
+            // Apparently there is no supported way to initialize a GroovyShell with a specified GroovyClassLoader.
```
svanoort: A bit confusing -- easier to follow if you say "GroovyShell creates its own GroovyClassLoader using the parent, no way to pass one in"

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
```diff
@@ -28,9 +35,51 @@
...
+    }
+
+    /**
+     * Disables the weird and unreliable {@link groovy.lang.GroovyClassLoader.InnerLoader}.
+     * This is apparently only necessary when you are using class recompilation, which we are not.
+     * We want the {@linkplain Class#getClassLoader defining loader} of {@code *.groovy} to be this one.
+     * Otherwise the defining loader will be an {@code InnerLoader}, and not necessarily the same instance from load to load.
+     * @see GroovyClassLoader#getTimeStamp
+     */
+    private static final class CleanGroovyClassLoader extends GroovyClassLoader {
```
svanoort: Is there something I'm missing here, or is this basically just stripping out the timestamp from InnerLoader?

There's a wrinkle here, which is that InnerLoader seems like it may support nesting, and I'm wondering if there are implications if that is removed.  But perhaps I misunderstand?

svanoort: Ah, okay so the big win here is that we're reusing the loader itself and not creating multiple ones for compilations of different SourceUnits.  

I can't help but feel like this opens some risk that one of the implementations relies on some separation of InnerLoaders (for example using the same class name for different SourceUnits and relying on different loaders  to distinguish them).  Have you looked closely at the implications of this?

But on the other hand: this will probably reduce memory footprint, implement a more "correct" approach and may actually end (or simplify debugging) PermGen memory leaks.

svanoort:
@jglick I see much to like with this and hate the InnerLoader nonsense in the groovy implementation to begin with (it seems very hacky and brittle). 

But I have several questions to make sure we've through through some of the risks with it too (and it *is* somewhat risky if we're modifying classloader behavior) -- don't feel comfortable approving it yet without answers to those.

jglick:
Yes this change drops the `timestamp` field, effectively. There is a fallback implementation based on a constant in the actual class file. I am not sure offhand if that fallback takes effect, or if the implementation simply defaults to “unknown timestamp”, but it should not matter since AFAICT this code is only used by the recompilation feature, which we do not need and do not enable.

The change should have no significant effect on memory usage. The motivation is to prevent different defining class loaders from being used for different classes from the same (non-inner) `GroovyClassLoader`. The effect is best seen by running the test in the downstream PR with and without the dependency update and studying the difference in behavior.

svanoort:
@jglick I'm not worried about the timestamp itself -- I'd already looked at the fallback and it is reasonable, besides which I am pretty sure we do not do anything that cares about the timestamp (after looking around some) -- as noted it applies to recompilation primarily unless you want to poke that API for other reasons (timing data, maybe?). 

> The change should have no significant effect on memory usage

Well, it'll get rid of a bunch of gratuitous and useless InnerLoaders, which may carry more baggage than one would expect being fairly heavy objects.  But probably not heavy enough to be visible to the user except in abusive cases.

> The motivation is to prevent different defining class loaders from being used for different classes from the same (non-inner) GroovyClassLoader

Yes, and what I'm wondering is if that bug may actually be used as a feature in some of the groovy insanity.  Ia Ia dynamic classloading hacks fh'tagn and all that.

jglick:
AFAIK, no, because just as with a normal loader, there should only be one `Class` per (outer) `GroovyClassLoader` x `String name`, unless you are recompiling modified source files.

I cannot swear to anything except to say that the change makes things simpler, fixes a clear bug, and does not regress any current test cases.

jglick:
If you look carefully at the commit message for https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/33/commits/3e74ab1b123ab2c4680498305b1d397baf6dd8ad you will see the issue: while the `GroovyClassLoader` is consistent, the `InnerLoader` used for classes in a given library varies and in certain combinations can yield two copies of a given class. More background is in JIRA.

svanoort:
@jglick I still am feeling less than confident about this one, because it may have cascading impacts. 

Do we have any *urgent* bugs tied to this, or is it just a rare issue?  If the latter, could we deploy a beta of workflow-cps with this fix into Astro and see if it triggers any issues?

Something like this, I just don't have confidence that the unit tests will exercise it properly because... well as we both know, Groovy allows some truly ~~clever~~ **insane** things, and sometimes behaviors of internals are not what one would expect.

rsandell:
:bee:

PR #121, opened 2017-04-25
Delete program.dat at the end of a build

jglick:
Just noticed this file was being left behind after the build completes, when it is no longer needed and is just a waste of disk space.

(Not deleting from `croak`, when we leave it behind for possible analysis.)

@reviewbybees


jglick:
CI failure is from npm I think.

PR #122, opened 2017-05-01
Demonstrating that ConfigSlurper can lead to memory leaks but evaluate does not

jglick:
Tests only.

@reviewbybees


jglick:
npm errors, I am afraid.

PR #123, opened 2017-05-09
[JENKINS-43934] Flatten GString’s found in step arguments, even when present in nested collections

jglick:
[JENKINS-43934](https://issues.jenkins-ci.org/browse/JENKINS-43934)

Never really posed a problem before, probably because there is no existing `StepDescriptor.newInstance` override which would be expecting evaluated `String`s in a collection type, but the previous behavior was at least potentially wrong…and this is needed for https://github.com/jenkinsci/workflow-cps-plugin/pull/98 so we might as well do it anyway.

`DescribableModel` already has `GString` handling, which really should never have been there, but we cannot remove that compatibly.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -58,6 +58,7 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
@@ -70,6 +71,8 @@ import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
+import org.codehaus.groovy.reflection.CachedClass;
+import org.codehaus.groovy.reflection.ReflectionCache;
 import org.jenkinsci.Symbol;
 
 import static org.jenkinsci.plugins.workflow.cps.ThreadTaskResult.*;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/DSLTest.java
```diff
@@ -62,11 +71,48 @@
...
+                return Collections.emptySet();
+            }
+            @Override public Step newInstance(Map<String, Object> arguments) throws Exception {
+                List<?> args = (List<?>) arguments.get("args");
+                StringBuilder b = new StringBuilder();
+                for (Object arg : args) {
+                    if (b.length() > 0) {
+                        b.append(' ');
+                    }
+                    b.append((String) arg);
```
jglick: Admittedly artificial, but at any rate this throws a `ClassCastException` without the fix.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -353,22 +356,51 @@ private NamedArgsAndClosure(Map<?,?> namedArgs, Closure body) {
...
+     */
+    private static Object flattenGString(Object v) {
+        if (v instanceof GString) {
+            return v.toString();
+        } else if (v instanceof List) {
+            boolean mutated = false;
+            List<Object> r = new ArrayList<>();
+            for (Object o : ((List<?>) v)) {
+                Object o2 = flattenGString(o);
+                mutated |= o != o2;
```
svanoort: Creative use of syntax.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -353,22 +356,51 @@ private NamedArgsAndClosure(Map<?,?> namedArgs, Closure body) {
...
+            for (Map.Entry<?,?> e : ((Map<?, ?>) v).entrySet()) {
+                Object k = e.getKey();
+                Object k2 = flattenGString(k);
+                Object o = e.getValue();
+                Object o2 = flattenGString(o);
+                mutated |= k != k2 || o != o2;
+                r.put(k2, o2);
+            }
+            return mutated ? r : v;
+        } else {
```
svanoort: @jglick What about Sets, general collections? 

Probably omitting those types is a bug, but this is still better than before, so I'm happy seeing it merged that way because it is still a million times better.

jglick: I deliberately considered only `List`s and `Map`s because these are the types that `DescribableModel` processes and that are generally used in `Describable` configuration and thus `Step` syntax. (Actually the interpreted `Map`s are more specific than that—only `String`-keyed maps, and only for the purpose of representing nested `Describable`s à la `$class`—but easiest to just process `Map`s generically, and should be harmless.)

jglick:
Argh, npm strikes again! I need to set aside some time to try to use https://github.com/jenkinsci/plugin-pom/pull/57 …

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
svanoort: That's reasonable enough, thanks for explaining.   

Not sure if there's an explicit restriction on Set/Collection types here, but that's good enough for me.

jglick:
@reviewbybees done

Now trying to rerun tests on my laptop since the npm download is busted, sigh; see [INFRA-1139](https://issues.jenkins-ci.org/browse/INFRA-1139).

PR #77, opened 2016-10-20
[JENKINS-27421] Improve behavior of map iteration workaround a bit

jglick:
[JENKINS-27421](https://issues.jenkins-ci.org/browse/JENKINS-27421)

Downstream of https://github.com/jenkinsci/script-security-plugin/pull/92.

@reviewbybees

PR #15, opened 2016-06-06
[JENKINS-26481] More clearly report problem passing closures to binary methods

jglick:
[JENKINS-26481](https://issues.jenkins-ci.org/browse/JENKINS-26481)

Tries to improve diagnostics until https://github.com/cloudbees/groovy-cps/pull/24 can be fixed.

Before this patch, `SerializationTest.eachClosure` fails with cryptic script behavior:

```
Started
[Pipeline] node
Running on master in /tmp/junit3096297553890442745/junit8071484631883149575/workspace/demo
[Pipeline] {
[Pipeline] writeFile
[Pipeline] semaphore
[Pipeline] readFile
[Pipeline] echo
a
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
```

Note that there is no error in the build, it just does not run correctly. After the patch, the build fails:

```
Started
[Pipeline] node
Running on master in /tmp/junit8559386618399120809/junit4301051117364074727/workspace/demo
[Pipeline] {
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
java.lang.UnsupportedOperationException: Calling public static java.lang.Object org.codehaus.groovy.runtime.DefaultGroovyMethods.each(java.lang.Object,groovy.lang.Closure) on a CPS-transformed closure is not yet supported (JENKINS-26481); encapsulate in a @NonCPS method, or use Java-style loops
	at org.jenkinsci.plugins.workflow.cps.GroovyClassLoaderWhitelist.checkJenkins26481(GroovyClassLoaderWhitelist.java:84)
	at org.jenkinsci.plugins.workflow.cps.GroovyClassLoaderWhitelist.permitsStaticMethod(GroovyClassLoaderWhitelist.java:54)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onMethodCall(SandboxInterceptor.java:83)
	at org.kohsuke.groovy.sandbox.impl.Checker$1.call(Checker.java:149)
	at org.kohsuke.groovy.sandbox.impl.Checker.checkedCall(Checker.java:146)
	at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.methodCall(SandboxInvoker.java:15)
	at WorkflowScript.run(WorkflowScript:2)
	at ___cps.transform___(Native Method)
	at com.cloudbees.groovy.cps.impl.ContinuationGroup.methodCall(ContinuationGroup.java:55)
[...]

PR #124, opened 2017-05-11
[JENKINS-26481] [JENKINS-27421] Use GroovyCategorySupport to invoke CpsDefaultGroovyMethods (w/o DGMPatcher) & IteratorHack

jglick:
Downstream of https://github.com/cloudbees/groovy-cps/pull/56.

[JENKINS-26481](https://issues.jenkins-ci.org/browse/JENKINS-26481): supersedes #14 and most of https://github.com/cloudbees/groovy-cps/pull/51. Seems to demonstrate that we never actually needed all the horrifying stuff in `DGMPatcher`. And seems to work in any Groovy version we support. (Pending integration of 2.5.0 in Jenkins core, we will however hit [GROOVY-6263](https://issues.apache.org/jira/browse/GROOVY-6263).)

Also [JENKINS-27421](https://issues.jenkins-ci.org/browse/JENKINS-27421): a more maintainable workaround for serialization issues.

Really the technique is quite general: any time you find a (non-`static`) method somewhere (typically in either the Java Platform or the GDK) which is not amenable to being called from Pipeline script (particularly CPS-transformed parts), we can replace its implementation with something else of our choice.

Tasks, including upstream:
- [X] using `CpsDefaultGroovyMethods`
- [X] adding coverage for most `Closure`-based methods
- [X] fixing `List.iterator`
- [X] fixing `Map.entrySet`
- [x] fixing `Set.iterator`
- [x] fixing `List.listIterator`
- [x] test coverage for miscellaneous new `Closure`-based methods
- [x] `GroovyClassLoaderWhitelist` to check only `Closure`-based methods defined outside the supported receivers, as in #14
- [x] some manual integration checking just because this is all so scary

Subsumes #135.

@reviewbybees if you are feeling adventurous


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -146,10 +157,33 @@ public StepExecution getStep() {
         this.step = step;
     }
 
+    /** TODO pending full coverage in {@link CpsDefaultGroovyMethods} */
```
jglick: Such as is added in https://github.com/cloudbees/groovy-cps/pull/52, but that mixes in other stuff which we do not need.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -160,9 +194,13 @@ public StepExecution getStep() {
 
         try (Timeout timeout = Timeout.limit(5, TimeUnit.MINUTES)) {
             LOGGER.log(FINE, "runNextChunk on {0}", resumeValue);
-            Outcome o = resumeValue;
+            final Outcome o = resumeValue;
             resumeValue = null;
-            outcome = program.run0(o);
+            outcome = GroovyCategorySupport.use(Arrays.<Class>asList(CpsDefaultGroovyMethods.class, CpsDefaultGroovyMethodsExt.class), new Closure<Outcome>(null) {
+                @Override public Outcome call() {
+                    return program.run0(o);
```
jglick: This is just for experimentation; probably we would rather want to have a change in `groovy-cps` which

* fleshes out `CpsDefaultGroovyMethods`
* calls `GroovyCategorySupport.use` from inside `Continuable.run0`

src/main/java/org/jenkinsci/plugins/workflow/cps/GroovyClassLoaderWhitelist.java
```diff
@@ -85,9 +84,11 @@ private boolean checkJenkins26481(Object[] args, /* TODO Java 8: just take Execu
         if (permits(method.getDeclaringClass())) { // fine for source-defined methods to take closures
             return true;
         }
-        for (int i = 0; i < args.length; i++) {
-            if (args[i] instanceof CpsClosure && parameterTypes.length > i && parameterTypes[i] == Closure.class) {
-                throw new UnsupportedOperationException("Calling " + method + " on a CPS-transformed closure is not yet supported (JENKINS-26481); encapsulate in a @NonCPS method, or use Java-style loops");
+        if (false) { // TODO be more discriminating
```
jglick: Suppressing the main change in #15 for the moment; #14 has something closer to what we need, though ideally we would check the actual method and determine if we have CPS coverage for _that_ method (and overload) now.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -146,10 +157,33 @@ public StepExecution getStep() {
...
+        }
+        public static <T> List<T> each(List<T> self, Closure<?> closure) {
+            if (!Caller.isAsynchronous(self, "each", closure) &&
+                    !Caller.isAsynchronous(CpsDefaultGroovyMethodsExt.class, "each", self, closure)) {
+                return DefaultGroovyMethods.each(self, closure);
+            }
+            Builder b = new Builder(loc("each"));
+            CpsFunction f = new CpsFunction(Arrays.asList("self", "closure"), b.block(
+                    b.staticCall(-1, CpsDefaultGroovyMethods.class, "each",
+                            b.staticCall(-1, InvokerHelper.class, "asIterator",
```
jglick: Just delegates to the `Iterator` overload.

src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
```diff
@@ -24,91 +24,86 @@
...
  */
-@Extension public class IteratorHack extends PickleFactory {
+public class IteratorHack {
 
-    private static final Logger LOGGER = Logger.getLogger(IteratorHack.class.getName());
-    private static final Set<String> ACCEPTED_TYPES = ImmutableSet.of("java.util.ArrayList$Itr");
-
-    @Override public Pickle writeReplace(Object object) {
-        if (ACCEPTED_TYPES.contains(object.getClass().getName())) {
-            return new Replacement(object);
```
jglick: Supersedes https://github.com/jenkinsci/pipeline-plugin/pull/372.

src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
```diff
@@ -219,12 +219,8 @@ public void stop(Throwable cause) throws Exception {
                 SemaphoreStep.success("new-one/1", null);
                 SemaphoreStep.success("new-two/1", null);
                 story.j.waitForCompletion(b);
-                /* TODO desired behavior:
```
jglick: Now we can have the “desired behavior” from #77.

jglick:
Hmm, seems `load` step calls and the like fail with

    groovy.lang.MissingMethodException: No signature of method: com.cloudbees.groovy.cps.SandboxCpsTransformer.copy() is applicable for argument types: (java.util.ArrayList) values: [[MethodNode@…[void main([Ljava.lang.String;)], ...]]
Possible solutions: any(), any(groovy.lang.Closure), notify(), wait(), find(), dump()

Looking into that.

jglick:
Seems that merely calling `GroovyCategorySupport.use`—even with no categories—triggers this. So some of the global behavioral switches turned on by the presence of category support seems to be breaking the compiler. I think I have a workaround.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -160,9 +195,13 @@ public StepExecution getStep() {
 
         try (Timeout timeout = Timeout.limit(5, TimeUnit.MINUTES)) {
             LOGGER.log(FINE, "runNextChunk on {0}", resumeValue);
-            Outcome o = resumeValue;
+            final Outcome o = resumeValue;
             resumeValue = null;
-            outcome = program.run0(o);
+            outcome = GroovyCategorySupport.use(Arrays.<Class>asList(CpsDefaultGroovyMethods.class, CpsDefaultGroovyMethodsExt.class, IteratorHack.class), new Closure<Outcome>(null) {
```
abayer: Given the black magic here, it'd be nice to have some comments explaining what exactly is going on. =)

jglick:
Absolutely we would like more test coverage.

As to

    Could not find artifact com.cloudbees:groovy-cps:jar:1.13-20170515.220420-1 in repo.jenkins-ci.org (https://repo.jenkins-ci.org/public/)

seems I forgot to deploy to the Jenkins repo. Will fix.

abayer:
@jglick I'm getting test failures locally - `CpsFlowExecutionTest#trustedShell_control`, `CpsScriptTest#evaluateShallSandbox`, `SnippetizerTest#doDslRef`...are you seeing those too or is this possibly a Java 8 issue? (since I'm using Java 8 locally with #126 merged in)

jglick:
`CpsScriptTest.evaluateShallSandbox` indeed fails for me. Will investigate.

abayer:
Had the same failures on Java 7 (and again with the right Java 8 version) so they look legit.

jglick:
Yes, somehow these changes seem to be suppressing the sandbox, so will need to figure out why that is.

abayer:
`Map.each` is failing with `groovy.lang.MissingMethodException: No signature of method: static com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.$callClosureForMapEntry__groovy_lang_Closure__java_util_Map_Entry() is applicable for argument types: (org.jenkinsci.plugins.workflow.cps.CpsClosure2, java.util.AbstractMap$SimpleImmutableEntry) values: [org.jenkinsci.plugins.workflow.cps.CpsClosure2@3f6cd2de, a=1]`, and it looks like `callClosureForMapEntry` doesn't exist in `CpsDefaultGroovyMethods` for real.

jglick:
npm errors are making CI more or less useless in this repository, so I need to take a detour to try using yarn instead.

jglick:
Seems that as of the last build, all tests passed, and we just bombed out due to npm (cf. #128).

jglick:
Seems to be causing a failure in `LibraryAdderTest.loaderReleased` which I will look into.

jglick:
Never mind, that seems to be a regression in 2.31; I will look into it, but not a blocker for this.

pom.xml
```diff
@@ -62,8 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
+        <jenkins.version>2.7.3</jenkins.version>
```
jglick: Did not seem practical to support Groovy 1.x, since we have to choose a Groovy release to generate `CpsDefaultGroovyMethods` from. Too old, and we do not cover enough methods; too new, and it may not link against an older Groovy core.

pom.xml
```diff
@@ -234,6 +233,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>credentials</artifactId>
```
jglick: This is #135.

src/main/java/org/jenkinsci/plugins/workflow/cps/GroovyClassLoaderWhitelist.java
```diff
@@ -76,21 +68,4 @@ private boolean permits(Class<?> declaringClass) {
         return permits(field.getDeclaringClass()) || delegate.permitsStaticFieldSet(field, value);
     }
 
-    /**
-     * Blocks accesses to some {@link DefaultGroovyMethods}, or any other binary method taking a closure, until JENKINS-26481 is fixed.
```
jglick: Yes there could be other `Closure`-taking binary methods which we are not translating, but it is too cumbersome to track them. Better to spend time implementing the more general, and precise, [JENKINS-31314](https://issues.jenkins-ci.org/browse/JENKINS-31314).

src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -235,8 +235,8 @@
         JenkinsRule.WebClient wc = r.createWebClient();
         String html = wc.goTo(Snippetizer.ACTION_URL + "/html").getWebResponse().getContentAsString();
         assertThat("text from LoadStep/help-path.html is included", html, containsString("the Groovy file to load"));
-        assertThat("SubversionSCM.workspaceUpdater is mentioned as an attribute of a value of GenericSCMStep.delegate", html, containsString("workspaceUpdater"));
-        assertThat("CheckoutUpdater is mentioned as an option", html, containsString("CheckoutUpdater"));
+        assertThat("GitSCM.submoduleCfg is mentioned as an attribute of a value of GenericSCMStep.scm", html, containsString("submoduleCfg"));
+        assertThat("CleanBeforeCheckout is mentioned as an option", html, containsString("CleanBeforeCheckout"));
```
jglick: Just related to `jenkins-test-harness` bump (no more detached plugins).

src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
```diff
@@ -24,91 +24,158 @@
...
+
+    /** Serializable replacement for {@link Map#entrySet}. */
+    public static <K, V> Set<Map.Entry<K, V>> entrySet(Map<K, V> map) {
+        if (!Caller.isAsynchronous(map, "entrySet") && !Caller.isAsynchronous(IteratorHack.class, "entrySet", map)) {
+            // In @NonCPS so no need to bother processing this.
+            return map.entrySet();
+        }
+        Set<Map.Entry<K, V>> entries = new LinkedHashSet<>();
+        for (Map.Entry<K, V> entry : map.entrySet()) {
+            // TODO return an object holding references to the map and the key and delegating all calls accordingly
```
jglick: These impls are not intended to be perfect or complete, just good enough to let typical script usages pass, where previously they would not.

pom.xml
abayer: +1 strongly. Especially given that Blue Ocean and Declarative require 2.7.3, so we've got that as the "default" baseline these days anyway.

src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
abayer: I forget - does this all enable `for (a : listOfA) { ... }` and the like?

jglick: Yes, see `IteratorHackTest`. Prior to this PR, `ArrayList.iterator` was safe but nothing else (and in particular I had no idea how to support iterating `Map`s and the like—the trick did not generalize well). Now most common interface methods in this category should work, and on any implementation class.

jglick:
@reviewbybees done

@kohsuke are you out there?

PR #125, opened 2017-05-16
[JENKINS-43055] Improve diagnostics to include a stack trace

jglick:
[JENKINS-43055](https://issues.jenkins-ci.org/browse/JENKINS-43055)

Was necessary to pinpoint the culprit.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -780,6 +780,10 @@ public class CpsFlowExecution extends FlowExecution {
     }
 
     public synchronized @CheckForNull FlowHead getFlowHead(int id) {
+        if (heads == null) {
+            LOGGER.log(Level.WARNING, null, new IllegalStateException("List of flow heads unset for " + this));
+            return null;
+        }
         return heads.get(id);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -780,14 +780,18 @@ public void waitForSuspension() throws InterruptedException, ExecutionException
     }
 
     public synchronized @CheckForNull FlowHead getFlowHead(int id) {
+        if (heads == null) {
```
jglick: Just in case; not observed.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -780,14 +780,18 @@ public void waitForSuspension() throws InterruptedException, ExecutionException
...
+        }
         return heads.get(id);
     }
 
     @Override
     public synchronized List<FlowNode> getCurrentHeads() {
         List<FlowNode> r = new ArrayList<FlowNode>();
         if (heads == null) {
-            LOGGER.log(Level.WARNING, "List of flow heads unset for {0}, perhaps due to broken storage", this);
+            LOGGER.log(Level.WARNING, null, new IllegalStateException("List of flow heads unset for " + this));
```
jglick: Stack trace needed.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -880,6 +884,10 @@ public CpsThreadDump getThreadDump() {
 
     @Override
     public synchronized boolean isCurrentHead(FlowNode n) {
+        if (heads == null) {
```
jglick: Also just in case.

jglick:
@reviewbybees done

jglick:
also #596

PR #126, opened 2017-05-16
Fix Javadoc error with Java 8

abayer:
The syntax was no longer valid. I'm open to alternatives.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -422,7 +422,7 @@ public class CpsFlowExecution extends FlowExecution {
     /**
      * Returns a groovy compiler used to load the script.
      *
-     * @see "doc/classloader.md"
+     * see "doc/classloader.md"
      * @see GroovyShell#getClassLoader()
      */
     public GroovyShell getShell() {
@@ -432,7 +432,7 @@ public class CpsFlowExecution extends FlowExecution {
     /**
      * Returns a groovy compiler used to load the trusted script.
      *
-     * @see "doc/classloader.md"
+     * see "doc/classloader.md"
      */
     public GroovyShell getTrustedShell() {
         return trusted;


jglick:
And [according to reference documentation](http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIEDI) this syntax is valid.

abayer:
@jglick I'm using `Java(TM) SE Runtime Environment (build 1.8.0-b132)` and am hitting this in a `mvn clean install`. Weeeeeird.

jglick:
Generates

```html
<dl>
<dt><span class="seeLabel">See Also:</span></dt>
<dd>"doc/classloader.md", 
<code>GroovyShell.getClassLoader()</code></dd>
</dl>
```

jglick:
I am using JDK 8u131 and have never seen that.

jglick:
Oh wait. You are not using 8u132; that does not even exist yet. You are using 8, like, 8u0. Please update.

abayer:
f'n OS X...

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -422,7 +422,7 @@ synchronized void logTimings() {
     /**
      * Returns a groovy compiler used to load the script.
      *
-     * @see "doc/classloader.md"
+     * see "doc/classloader.md"
      * @see GroovyShell#getClassLoader()
```
svanoort: Maybe better to just use {@link} in general to avoid accidental issues?

jglick: `@link` would be used for intra-Javadoc references, which this is not.

svanoort: @jglick  For the GroovyShell ref, not the doc.

jglick: Why? That was not the error, and `javadoc` already checks validity of members in this form of `@see`.

PR #127, opened 2017-05-16
[JENKINS-43055] Add support for FlowExecutionListener

abayer:
[JENKINS-43055](https://issues.jenkins-ci.org/browse/JENKINS-43055)

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/36

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -58,6 +58,7 @@ import org.jboss.marshalling.Unmarshaller;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionListener;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
@@ -489,6 +490,9 @@ public class CpsFlowExecution extends FlowExecution {
         g.register(s);
         final SettableFuture<CpsThreadGroup> f = SettableFuture.create();
         programPromise = f;
+
+        FlowExecutionListener.fireRunning(this, false);
+
         g.runner.submit(new Runnable() {
             @Override
             public void run() {
@@ -625,6 +629,7 @@ public class CpsFlowExecution extends FlowExecution {
                                 CpsThreadGroup g = (CpsThreadGroup) u.readObject();
                                 result.set(g);
                                 try {
+                                    FlowExecutionListener.fireRunning(CpsFlowExecution.this, true);
                                     if (g.isPaused()) {
                                         owner.getListener().getLogger().println("Still paused");
                                     } else {
@@ -1023,6 +1028,7 @@ public class CpsFlowExecution extends FlowExecution {
         first.setNewHead(head);
         heads.clear();
         heads.put(first.getId(),first);
+        FlowExecutionListener.fireCompleted(this);
     }
 
     void cleanUpHeap() {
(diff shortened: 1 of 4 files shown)


abayer:
Still needs tests, but I wanted to see if the design was alright before proceeding.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -489,6 +489,7 @@ public void start() throws IOException {
         g.register(s);
         final SettableFuture<CpsThreadGroup> f = SettableFuture.create();
         programPromise = f;
+
```
jglick: ?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -232,13 +233,20 @@ public void unexport(BodyReference ref) {
                 public Void call() throws Exception {
                     Jenkins j = Jenkins.getInstance();
                     if (paused.get() || j == null || j.isQuietingDown()) {
+                        for (FlowExecutionListener listener : FlowExecutionListener.all()) {
+                            listener.onPaused(execution);
```
jglick: This is called whenever we are asked to schedule something, not necessarily when the pause button is pressed.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
abayer: Dunno what happened there. Fixing.

jglick:
Can you merge with `master` please so we can get a new build using yarn, which I hope will be more stable than npm?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -239,6 +241,10 @@ public Void call() throws Exception {
                         return null;
                     }
 
+                    for (FlowExecutionListener listener : ExtensionList.lookup(FlowExecutionListener.class)) {
+                        listener.onRunning(execution);
```
jglick: This is going to be called many times during a build: every time we reënter the CPS VM thread. That does not match the expectations given by its Javadoc, which is one call at build start time and one call per running build per Jenkins restart.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -382,6 +384,53 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
         }
     }
 
+    @Test
+    public void flowExecutionListener() throws Exception {
+        story.addStep(new Statement() {
+                          @Override
+                          public void evaluate() throws Throwable {
+                              WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                              p.setDefinition(new CpsFlowDefinition("echo 'Running for listener'; semaphore 'wait'", true));
```
jglick: Try inserting a `sleep 0` in the middle here and I am betting you will get a count > 2.

abayer: Nope, still got 2...

abayer: Ok, threw in another sleep after the semaphore and did get 3.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
abayer: So where would be the right place to put this?

abayer: Ah, think I figured it out.

jglick:
> So where would be the right place to put this?

`CpsFlowExecution.start` for a new flow vs. `onLoad` for a resumed one (which I guess should just be distinct listener methods).

It is not actually obvious to me that the listener calls should be coming from `workflow-cps` at all. It is `WorkflowRun` which calls `FlowExecution.start` and `.onLoad`, so it would seem logical for `WorkflowRun` to be notifying listeners that these methods have been called. Similarly, `WorkflowRun.finish` would be a logical place to notify listeners that a flow has completed.

abayer:
Yeah, that makes sense.

jglick:
To be closed?

abayer:
Yup, closing.

PR #130, opened 2017-05-18
Support GraphListener ExtensionPoint

abayer:
Subsumes #127, downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/37 - leaving as WIP for now as we wait to see if this is desired.

cc @reviewbybees


abayer:
This was based on the defunct #127. Its functionality is moved off to `workflow-job`, so this is replaced with #137.

PR #131, opened 2017-05-18
Extra test for JENKINS-43934

jglick:
Amends #123: I was afraid I had missed a case involving symbols, but it turns out the original fix works. Came up in the context of #98.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
@@ -22,8 +22,10 @@
  * THE SOFTWARE.
  */
 
-package org.jenkinsci.plugins.workflow;
+package org.jenkinsci.plugins.workflow.cps;
 
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Descriptor;
 import hudson.model.Result;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -22,22 +22,19 @@
  * THE SOFTWARE.
  */
 
-package org.jenkinsci.plugins.workflow;
+package org.jenkinsci.plugins.workflow.cps;
```
jglick: Package did not match `DSL` itself; fixing that while I am here.

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -257,25 +293,35 @@ public void namedSoleParamForStep() throws Exception {
...
     }
 
+    @Issue("JENKINS-37538")
     @Test public void contextClassLoader() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
         p.setDefinition(new CpsFlowDefinition("try {def c = classLoad(getClass().name); error(/did not expect to be able to load ${c} from ${c.classLoader}/)} catch (ClassNotFoundException x) {echo(/good, got ${x}/)}", false));
         r.assertBuildStatusSuccess(p.scheduleBuild2(0));
     }
-    public static class CLStep extends AbstractStepImpl {
+    public static class CLStep extends Step {
```
jglick: Cleaning up some deprecations while I am here.

PR #132, opened 2017-05-22
JENKINS-44406 Return Null Object instead of null for compute

fengxx:
[JENKINS-44406](https://issues.jenkins-ci.org/browse/JENKINS-44406)

when pipeline step get removed or renamed, the REST API  wfapi/runs?fullStages=true will throw NPE. Simply use Null object to fix the issue


svanoort:
A clever idea, but we don't want to have a special NullStepDescriptor, because the descriptor could be installed. 

Since the cache API is already marked as CheckForNull,  #133  offers a more correct approach.

svanoort:
Closing because this we have a working solution now merged (sorry!).

PR #133, opened 2017-05-22
Fix StepDescriptorCache handling of null StepDescriptor lookup [JENKINS-44406]

svanoort:
[JENKINS-44406](https://issues.jenkins-ci.org/browse/JENKINS-44406)

Don't rely on Memoizer to handle nulls correctly - it doesn't handle a computed null because ConcurrentHashMap can't (potential fix to Jenkins core).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
@@ -28,7 +28,7 @@ import hudson.Extension;
 import hudson.ExtensionPoint;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
-import hudson.util.Memoizer;
+import hudson.model.Descriptor;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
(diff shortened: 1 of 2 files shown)


svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
```diff
@@ -55,26 +56,30 @@ public static void invalidateGlobalCache() {
...
-            return descriptorCache.get(descriptorId);
+        if (descriptorId == null) {
+            return null;
         }
-        return null;
 
+        StepDescriptor v = store.get(descriptorId);
+        if(v!=null) return v;
+
+        synchronized (this) {
```
jglick: :bug: double-checked locking. A single check in the store suffices.

src/test/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCacheTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+public class StepDescriptorCacheTest {
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test
```
jglick: missing `@Issue`

jglick:
Amends #91 FTR.

src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepDescriptorCache.java
svanoort: @jglick Look closely - this isn't reliant on field access, it does not suffer from double-checked locking, and is adapted from Memoizer itself.

jglick: The code in `Memoizer` is not good. It is _one_ (bad) workaround for the case where there is a problem if two threads compute the same value, because the cache function is not idempotent (or is so expensive that we would rather block another thread than do it twice). That is not the case here: the lookup is quick and harmless to do in two different threads in the unlikely event they overlap. So you should remove the redundant check and the synchronization.

svanoort:
> BTW Memoizer could probably work if you used Optional<StepDescriptor> as the value type

You'd think that (or a null-marker object, similarly) would do the trick -- but we shouldn't be caching not-present results, because someone could subsequently reinstall the plugin and then we'd be getting false negatives for the plugin.

jglick:
> we shouldn't be caching not-present results, because someone could subsequently reinstall the plugin

True, we should have a mild preference for not caching negative results.

svanoort:
> True, we should have a mild preference for not caching negative results

It's more than that - this would constitute a bug unless we added appropriate invalidation of the cache when new Steps are installed.

PR #135, opened 2017-05-23
InjectedTest failure

jglick:
Introduced in #98.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -234,6 +234,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


jglick:
yarn died with

```
Error: read ECONNRESET
    at exports._errnoException (util.js:837:11)
    at TLSWrap.onread (net.js:544:26)
```

Sigh. I am seriously thinking of ripping out the ACE editor into a separate plugin just so I do not have to keep on dealing with constant test failures from the JavaScript tooling.

PR #136, opened 2017-05-24
[JENKINS-34464] Bump structs version, add test for result roundtrip

abayer:
[JENKINS-34464](https://issues.jenkins-ci.org/browse/JENKINS-34464)

Downstream of https://github.com/jenkinsci/structs-plugin/pull/19, upstream of https://github.com/jenkinsci/workflow-multibranch-plugin/pull/63

The `workflow-multibranch` test was failing, seemingly due to `workflow-cps` having the older `structs` version (plus some other problems, but those are fixed over in that PR anyway). So let's bump the `structs` release.

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.6</version>
+            <version>1.7-20170522.160448-1</version> <!-- TODO: Switch to release after https://github.com/jenkinsci/structs-plugin/pull/19 is merged and released -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.7-20170525.172515-3</version> <!-- TODO: Switch to release after https://github.com/jenkinsci/structs-plugin/pull/19 is merged and released -->
+            <version>1.7</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.7-20170522.160448-1</version> <!-- TODO: Switch to release after https://github.com/jenkinsci/structs-plugin/pull/19 is merged and released -->
+            <version>1.7-20170525.172515-3</version> <!-- TODO: Switch to release after https://github.com/jenkinsci/structs-plugin/pull/19 is merged and released -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
@@ -1,0 +1,72 @@
+package org.jenkinsci.plugins.workflow.testMetaStep;
+
+import hudson.Extension;
+import hudson.model.Result;
+import hudson.model.TaskListener;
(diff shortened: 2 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
```diff
@@ -0,0 +1,72 @@
...
+    public static class EchoResultStepExecution extends StepExecution {
+        private final EchoResultStep step;
+
+        public EchoResultStepExecution(EchoResultStep s, StepContext context) {
+            super(context);
+            this.step = s;
+        }
+
+        @Override
+        public boolean start() throws Exception {
```
jglick: Note that the test is not actually running it.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
```diff
@@ -0,0 +1,72 @@
...
+            TaskListener listener = getContext().get(TaskListener.class);
+
+            if (listener != null)
```
jglick: No null check needed, it is required.

abayer: My IDE complained about a possible NPE without it, so hey, I put in the check.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
abayer: Yup.

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
jglick: What is wrong with your IDE? `StepContext.get` is correctly marked `@Nullable`.

abayer:
`InjectedTest` failure is solved in #135. Trying to figure out why the `resultRoundTrips` test is failing, 'cos it isn't locally so far...

src/test/java/org/jenkinsci/plugins/workflow/testMetaStep/EchoResultStep.java
abayer: It's complaining at the `listener.getLogger()...`.

abayer:
Ok, I am mystified as to why `SnippetizerTest#resultRoundTrips` is failing here when it works for me locally.

jglick:
```
org.junit.ComparisonFailure: expected:<echoResult ['UNSTABLE']> but was:<echoResult [<object of type hudson.model.Result>]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.jenkinsci.plugins.workflow.cps.SnippetizerTester.assertRoundTrip(SnippetizerTester.java:95)
	at org.jenkinsci.plugins.workflow.cps.SnippetizerTest.resultRoundTrips(SnippetizerTest.java:315)
```

specifically. Perhaps you are picking up the wrong snapshot?

abayer:
It's got the right `SNAPSHOT` in the POM - I'm wondering if it's somehow getting messed up and picking up https://repo.jenkins-ci.org/snapshots/org/jenkins-ci/plugins/structs/1.7-SNAPSHOT/structs-1.7-20170522.183630-2.hpi, a SNAPSHOT you deployed a couple hours after https://repo.jenkins-ci.org/snapshots/org/jenkins-ci/plugins/structs/1.7-SNAPSHOT/structs-1.7-20170522.160448-1.hpi, the one in the POM here?

jglick:
(We really need to switch to doing everything inside Docker containers, especially now that we have a repository mirror inside Azure so it should not be too expensive to require every build to download everything.)

pom.xml
```diff
@@ -103,7 +103,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.6</version>
+            <version>1.7-20170525.172515-3</version> <!-- TODO: Switch to release after https://github.com/jenkinsci/structs-plugin/pull/19 is merged and released -->
```
jglick: 1.7

jglick:
Feel free to merge if CI build passes.

PR #137, opened 2017-05-24
Add support for GraphListener as ExtensionPoint

abayer:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/38

cc @reviewbybees


jglick:
How is this different than #130?

PR #138, opened 2017-05-25
Picking up a newer version of git so as to avoid a noisy linkage error

jglick:
See https://github.com/jenkinsci/jenkins-test-harness/pull/35 for background.

@reviewbybees


jglick:
`CpsFlowExecutionTest.quietDown` is known to be flaky.

PR #139, opened 2017-06-01
Tests for DGM running through script security in actual jobs.

abayer:
Downstream of https://github.com/jenkinsci/script-security-plugin/pull/124, post https://github.com/jenkinsci/workflow-cps-plugin/pull/124.

Notes that `sum` doesn't work due to needing `invokeMethod`, which is...obviously a non-starter.

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
+        <groovy-cps.version>1.13.1</groovy-cps.version>
     </properties>
     <dependencies>
         <dependency>
@@ -92,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.27</version>
+            <version>1.28-20170601.214517-1</version> <!-- TODO: Switch to release once https://github.com/jenkinsci/script-security-plugin/pull/124 is merged and released -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -174,7 +175,7 @@
         <dependency>
             <groupId>com.cloudbees</groupId>
             <artifactId>groovy-cps</artifactId>
-            <version>1.13</version>
+            <version>${groovy-cps.version}</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.ui</groupId>
@@ -237,6 +238,13 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
             <version>2.1.8</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.cloudbees</groupId>
+            <artifactId>groovy-cps</artifactId>
+            <version>${groovy-cps.version}</version>
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.29-20170609.183458-2</version> <!-- TODO: Switch to release once https://github.com/jenkinsci/script-security-plugin/pull/124 is merged and released -->
+            <version>1.29-20170614.164511-3</version> <!-- TODO: Switch to release once https://github.com/jenkinsci/script-security-plugin/pull/124 is merged and released -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
-        <groovy-cps.version>1.14</groovy-cps.version>
+        <groovy-cps.version>1.15-20170609.190910-1</groovy-cps.version>
     </properties>
     <dependencies>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
-        <groovy-cps.version>1.15-20170609.190910-1</groovy-cps.version>
+        <groovy-cps.version>1.15-20170609.190910-1</groovy-cps.version> <!-- TODO: Switch to release once https://github.com/cloudbees/groovy-cps/pull/60 is merged and released -->
     </properties>
     <dependencies>
         <dependency>
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.29-20170606.150723-1</version> <!-- TODO: Switch to release once https://github.com/jenkinsci/script-security-plugin/pull/124 is merged and released -->
+            <version>1.29-20170609.183458-2</version> <!-- TODO: Switch to release once https://github.com/jenkinsci/script-security-plugin/pull/124 is merged and released -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
-        <groovy-cps.version>1.16-20170614.172949-1</groovy-cps.version> <!-- TODO: Switch to release once https://github.com/cloudbees/groovy-cps/pull/60 is merged and released -->
+        <groovy-cps.version>1.16-20170614.193138-2</groovy-cps.version> <!-- TODO: Switch to release once https://github.com/cloudbees/groovy-cps/pull/62 is merged and released -->
     </properties>
     <dependencies>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
-        <groovy-cps.version>1.16-20170614.193138-2</groovy-cps.version> <!-- TODO: Switch to release once https://github.com/cloudbees/groovy-cps/pull/62 is merged and released -->
+        <groovy-cps.version>1.16</groovy-cps.version>
     </properties>
     <dependencies>
         <dependency>
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.29-20170614.164511-3</version> <!-- TODO: Switch to release once https://github.com/jenkinsci/script-security-plugin/pull/124 is merged and released -->
+            <version>1.29</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/CpsDefaultGroovyMethodsTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@RunWith(Parameterized.class)
+public class CpsDefaultGroovyMethodsTest {
```
jglick: `@Issue` for reference

src/test/java/org/jenkinsci/plugins/workflow/CpsDefaultGroovyMethodsTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @ClassRule
+    public static JenkinsRule r = new JenkinsRule();
+    @ClassRule
+    public static BuildWatcher buildWatcher = new BuildWatcher();
+
+    private String testName;
+    private String testCode;
+    private Object testResult; // Unused
```
jglick: Cleaner to just process the parameters from the upstream test.

Or, use this parameter, by `assert`ing that the expression equals the expected value.

src/test/java/org/jenkinsci/plugins/workflow/CpsDefaultGroovyMethodsTest.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Test
+    public void dgm() throws Exception {
+        WorkflowJob j = r.createProject(WorkflowJob.class, testName);
+        j.setDefinition(new CpsFlowDefinition(testCode, true));
+
+        WorkflowRun b = j.scheduleBuild2(0).waitForStart();
+
+        r.assertBuildStatusSuccess(r.waitForCompletion(b));
```
jglick:
```java
r.buildAndAssertSuccess(j);
```

pom.xml
```diff
@@ -67,6 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
+        <groovy-cps.version>1.13.1</groovy-cps.version>
```
svanoort: Did you actually make a property for a single use?

src/test/java/org/jenkinsci/plugins/workflow/CpsDefaultGroovyMethodsTest.java
abayer: Yeah, I like the idea of that. Lemme see if I can come up with something.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsWhitelist.java
```diff
@@ -103,6 +104,9 @@ public boolean permitsStaticMethod(Method method, Object[] args) {
             String driverTo = "org.codehaus.groovy.runtime.";
             if (cn.startsWith(driverFrom)) {
                 try {
+                    if (Modifier.isPrivate(method.getModifiers())) {
+                        return true;
```
jglick: Probably OK but I want to think a bit about more why this is necessary. `dgm-builder` is not intentionally generating code that uses `SandboxCpsTransformer` hooks so how are these calls going through the sandbox interceptor to begin with?

abayer: That I have no idea on at all.

src/test/java/org/jenkinsci/plugins/workflow/CpsDefaultGroovyMethodsTest.java
abayer: Whoops, fixing.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsWhitelist.java
abayer: So here's what I'm seeing when it explodes:
```
[collectEntriesKV #1] org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod com.cloudbees.groovy.cps.CpsDefaultGroovyMethods $callClosureForMapEntry__groovy_lang_Closure__java_util_Map_Entry groovy.lang.Closure java.util.Map$Entry
[collectEntriesKV #1] 	at org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist.rejectStaticMethod(StaticWhitelist.java:192)
[collectEntriesKV #1] 	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onStaticCall(SandboxInterceptor.java:142)
[collectEntriesKV #1] 	at org.kohsuke.groovy.sandbox.impl.Checker$2.call(Checker.java:180)
[collectEntriesKV #1] 	at org.kohsuke.groovy.sandbox.impl.Checker.checkedStaticCall(Checker.java:177)
[collectEntriesKV #1] 	at org.kohsuke.groovy.sandbox.impl.Checker.checkedCall(Checker.java:91)
[collectEntriesKV #1] 	at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.methodCall(SandboxInvoker.java:16)
[collectEntriesKV #1] 	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.collectEntries(CpsDefaultGroovyMethods:3489)
[collectEntriesKV #1] 	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.collectEntries(CpsDefaultGroovyMethods:3515)
[collectEntriesKV #1] 	at WorkflowScript.run(WorkflowScript:2)
```

abayer: Ok, I just spent way too long trying to figure this out. We are definitely generating CPS code, so, for example, `CpsDefaultGroovyMethods.$collectEntries__java_util_Iterator__java_util_Map__groovy_lang_Closure` is calling `$addEntry__java_util_Map__java_lang_Object` via `b.staticCall(...)`. Which, I think, means that `ContinuationGroup.methodCall(...)` is getting called at runtime, and *that* calls `Invoker.methodCall(...)`, which is resulting in `SandboxInvoker.methodCall(...)` and hence the rest of the stacktrace. I did some tests and I don't think `SandboxTransformer` is ever being invoked for this method, I think it's just that the `Invoker` is `SandboxInvoker` and so tada? I dunno. A theory.

jglick: So that is not right. The compiled code is carrying over the same `Env` as its caller. I wonder if we need to create a subclass of `CpsFunction` which passes a `ProxyEnv` with a `DefaultInvoker`.

abayer: Sounds like a good idea to me. I can imagine some other cases where we may want to explicitly write CPS code, which would benefit from that as well.

abayer: @jglick I'm wondering if we actually need a new extension of `ProxyEnv` (let's call it `DefaultInvokerProxyEnv` because why not) that works the same as `ProxyEnv` except for returning `DefaultInvoker.INSTANCE` for `getInvoker()`, and then have, say, `CpsInternalFunction.invoke(...)` do `return super.invoke(new DefautInvokerProxyEnv(caller), ...)`? Or some such?

jglick: Yes something along those lines.

Basically, when we are inside one of the transformed methods, it is senseless to be using the sandbox invoker. *But* need to verify that when we invoke a `Closure` it is still using the original invoker.

abayer: Would it possibly make sense to have a check [here](https://github.com/cloudbees/groovy-cps/blob/9e35476a21d6baa801775ce7852d11b4e4a0694d/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java#L62-L62) if `e` is an instance of `DefaultInvokerProxyEnv`, and if so, do `return inv.invoke(((DefaultInvokerProxyEnv)e).originalEnv, loc, k)` or something along those lines?

jglick: Not sure why that would be necessary, though I have not yet studied it in much detail. I was thinking that [this](https://github.com/cloudbees/groovy-cps/blob/63b6aa1aa3b9414b4b2657ffbb5eeea39b21d115/dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java#L151) would instead point to a subclass which creates a `FunctionCallEnv` with a `ProxyEnv` subtype overriding `getInvoker`.

Again, there would need to be an explicit test that, say,

```groovy
[1].each {x -> System.exit(666)}
```

is rejected.

BTW [this Javadoc](https://github.com/cloudbees/groovy-cps/blob/63b6aa1aa3b9414b4b2657ffbb5eeea39b21d115/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallable.java#L27) belongs in somebody’s hall of shame.

jglick: To be clear, my reluctance to accept the current patch is that it is not clear that all `private static` methods in `Cps*` are in fact safe. Just because we decline to whitelist the `public` methods which are supposed to delegate to them does not block a script from calling them directly, since Groovy allows access to `private` methods.

abayer: Agreed wholeheartedly. I'm working on figuring out how to do the proper re-sandboxing now.

abayer: So the above is what I'm trying and the test is failing, 'cos the `ProxyEnv` subtype is cascading further than desired. I will now figure out how to stop that. =)

jglick:
Have some test failures here.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -65,4 +67,16 @@
...
+    @Test
+    public void sandboxInvokerUsed() throws Exception {
+        WorkflowJob job = jenkins.jenkins.createProject(WorkflowJob.class, "p");
+        job.setDefinition(new CpsFlowDefinition("[a: 1, b: 2].collectEntries { k, v ->\n" +
+                "  Jenkins.getInstance()\n" +
+                "  [(v): k]\n" +
+                "}\n", true));
+
+        WorkflowRun r = job.scheduleBuild2(0).waitForStart();
+        jenkins.assertBuildStatus(Result.FAILURE, jenkins.waitForCompletion(r));
```
jglick:
```java
WorkflowRun r = jenkins.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0).get());
```

abayer: Yeah, I keep forgetting that.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -75,8 +75,7 @@ public void sandboxInvokerUsed() throws Exception {
                 "  [(v): k]\n" +
                 "}\n", true));
 
-        WorkflowRun r = job.scheduleBuild2(0).waitForStart();
-        jenkins.assertBuildStatus(Result.FAILURE, jenkins.waitForCompletion(r));
+        jenkins.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0).get());
```
jglick: forgot to assign to the variable

abayer: *facepalm* Fixed.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -65,4 +67,15 @@
         jenkins.configRoundtrip(job);
     }
 
+    @Test
+    public void sandboxInvokerUsed() throws Exception {
```
jglick: Not really the right place for this test, but no matter.

pom.xml
```diff
@@ -67,6 +67,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
+        <groovy-cps.version>1.16-20170614.172949-1</groovy-cps.version> <!-- TODO: Switch to release once https://github.com/cloudbees/groovy-cps/pull/60 is merged and released -->
```
jglick: If you want to try https://github.com/cloudbees/groovy-cps/pull/62 I have deployed it as `1.16-20170614.193138-2`.

abayer: Done.

PR #141, opened 2017-06-05
WaitForConditionStepTest#restartDuringDelay sometimes flaked

jglick:
Because `StepDescriptor.all` was empty. Error:

    No such DSL method 'semaphore' found among steps [] or symbols […OK…] or globals […OK…]

Initially I thought the problem was that `functions` (intrduced in https://github.com/jenkinsci/pipeline-plugin/commit/62ab58d2e0894f160475ac1592e841adb9f2468b) was being cached with an empty list during a call at a bad time, and later a legitimate call then picked up the bad cache. But in fact it seems that sometimes the method is called when the extension list is just empty, and shortly thereafter it is OK. Not really sure what is going on here but I suspect this is specific to `RestartableJenkinsRule`: I do see two sets of Jenkins startup log messages, which looks anomalous, and suggests that `ItemListener.onLoaded` fires in between and things explode.

Have also occasionally seen `CpsFlowExecutionTest.quietDown` flake this way:

```
unexpected build status; build log was:
    ------
    Started
    [Pipeline] semaphore
    Resuming build at Thu Feb 23 16:26:44 EST 2017 after Jenkins restart
    [Pipeline] End of Pipeline
    java.lang.NoSuchMethodError: No such DSL method 'echo' found among steps [] or symbols [CO, california, circle, disableConcurrentBuilds, hawaii, island, junit, monomorphSymbol, multiShape, nevada, newYork, or, pipelineTriggers, polygon] or globals []
    	at org.jenkinsci.plugins.workflow.cps.DSL.invokeMethod(DSL.java:149)
```

I had thought https://github.com/jenkinsci/workflow-cps-plugin/pull/32/files#diff-6452cd1bea46404d75cc1ba4559cc3adR229 did not work reliably or https://github.com/jenkinsci/workflow-cps-plugin/pull/68/files#diff-251f35b4606691f65bd766f8e17b9613L253 was wrong, so Jenkins is shutting down when `echo` is called, but now I am more inclined to blame `RestartableJenkinsRule`.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -30,6 +30,7 @@ import groovy.lang.Closure;
 import groovy.lang.GString;
 import groovy.lang.GroovyObject;
 import groovy.lang.GroovyObjectSupport;
+import groovy.lang.GroovyRuntimeException;
 import hudson.EnvVars;
 import hudson.model.Computer;
 import hudson.model.Describable;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -134,11 +128,19 @@ public Object invokeMethod(String name, Object args) {
             if (exec==null)
                 exec = (CpsFlowExecution) handle.get();
         } catch (IOException e) {
-            throw new Error(e); // TODO
+            throw new GroovyRuntimeException(e);
         }
 
         if (functions == null) {
             functions = new TreeMap<>();
+            while (StepDescriptor.all().isEmpty()) {
```
svanoort: 🐜  Why not fetch StepDescriptor.all() and save to variable (mutated in the while loop), to avoid a double fetch if started, since we're going to reuse it later?

jglick: Could, but it should not really matter since just obtaining the `ExtensionList` instance is cheap—and anyway we are only doing this once per build.

(Really I considered getting rid of the cache altogether, since the list is not very long and `getFunctionName` is generally instantaneous. Was just not sure how much overhead there is in iterating the list.)

PR #142, opened 2017-06-09
Adds option to treat missing script file as an empty file

signed8bit:
When automating project creation along with base Jenkins jobs
for various projects, the current behavior of the Groovy plugin requires
a development team to create an empty script file before they may be
ready to develop a proper pipeline in order to pass CI. This becomes
a barrier to a development team when the failure of the initial CI job
provides a negative code review on a change.

This commit adds an option to treat a missing script file as an empty
file thus allowing the CI job to pass. The default for this option is
"false" thus preserving the original behavior of the plugin.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
@@ -61,6 +61,7 @@ public class CpsScmFlowDefinition extends FlowDefinition {
 
     private final SCM scm;
     private final String scriptPath;
+    private boolean ignoreMissingScript;
     private boolean lightweight;
 
     @DataBoundConstructor public CpsScmFlowDefinition(SCM scm, String scriptPath) {
@@ -74,6 +75,14 @@ public class CpsScmFlowDefinition extends FlowDefinition {
 
     public String getScriptPath() {
         return scriptPath;
+    }
+
+    public boolean isIgnoreMissingScript() {
+        return ignoreMissingScript;
+    }
+
+    @DataBoundSetter public void setIgnoreMissingScript(boolean ignoreMissingScript) {
+        this.ignoreMissingScript = ignoreMissingScript;
     }
 
     public boolean isLightweight() {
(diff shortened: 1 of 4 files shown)


signed8bit:
@jglick, any chance you could have a look at this one? I'd like to get a sense of whether or not you all would be willing to accept the change. I'd be more than happy to file an issue for it as well.

jglick:
I am afraid I do not understand the purpose of this change at all.

signed8bit:
Sure, I'd be happy to try and clarify. In our CI setup we automate the creation of projects in Gerrit along with a base Jenkins job that looks for a Jenkinsfile in the corresponding project.

When a new project is created, the development team may not initially have a Jenkinsfile created but Jenkins will run the configured job automatically. What happens is, any submitted change to a project ends up with a failed verification from Jenkins until a Jenkinsfile is created. In Gerrit, this prevents a change from being merged.

The workaround for this is to instruct the development team to create an empty Jenkinsfile which allows the job to succeed. All this change does is provide an option to treat a missing Jenkinsfile the same as an empty Jenkins file. 

This is mainly aimed at easing the on-boarding of a development team until they are ready to create a proper Jenkinsfile.

jglick:
I would rather just fix whatever automation is creating your Jenkins jobs to not do so until a `Jenkinsfile` exists. That is what a multibranch folder would do, after all. And IIUC your whole infrastructure is just a workaround for the lack of an `SCMSource` (and perhaps `SCMNavigator`) implemented in the `gerrit-trigger` plugin—no one has bothered to write one yet.

rabadin:
This seems like a useful change to me: it removes the dependency between the team in charge of managing/creating the jobs and the team working on the project where the Jenkinsfile resides.  The jobs can be pre-created once and for all.  The separation of concerns is really the kick-ass feature of the whole Jenkinsfile design where the teams working on the projects themselves have full control over what the CI tests... and this changes allows (it's just an option) to push that separation further.

Another benefit of this fix is that it allows to deal nicely with projects with multiple branches where only some branches have Jenkinsfile(s) (for instance when you're transitioning to using Jenkinsfile(s) and have legacy branch which don't use them).  The job doesn't need to have the knowledge of which branches have Jenkinsfile(s) and which branches don't.

Finally, this is just adding a new option in a fully backward-compatible way so I vote for including this change ;).

jglick:
> The job doesn't need to have the knowledge of which branches have Jenkinsfile(s) and which branches don't.

This is already solved by multibranch. Branch projects are created for those projects which have a `Jenkinsfile`.

jglick:
I am not inclined to accept a feature which adds UI weight and maintenance burden and just encourages an incorrect workflow. You should rather fix your job generation system to check for the existence of a `Jenkinsfile` first. If that is not possible, you can always use a `CpsFlowDefinition` with a wrapper script along the lines of

```groovy
def script
node {
  checkout …
  if (fileExists 'Jenkinsfile') {
    script = readFile 'Jenkinsfile'
  }
}
if (script) {
  evaluate script
} else {
  echo 'Create a Jenkinsfile to get started'
}
```

signed8bit:
Thanks for the feedback and for having a look @jglick. Closing.

PR #19, opened 2016-06-08
[JENKINS-34281] Lost FlowExecutionList entries

jglick:
Another symptom of [JENKINS-34281](https://issues.jenkins-ci.org/browse/JENKINS-34281).

Cf. https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/8, https://github.com/jenkinsci/workflow-api-plugin/pull/4, https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/4.

@reviewbybees

PR #144, opened 2017-06-15
Do not allow exceptions from suspendAll to cancel shutdown

jglick:
Observed a problem in the field possibly correlated with

```
java.lang.Error: java.lang.reflect.InvocationTargetException
	at …
	at jenkins.model.Jenkins._cleanUpRunTerminators(Jenkins.java:3284)
	at jenkins.model.Jenkins.cleanUp(Jenkins.java:3206)
	at hudson.lifecycle.UnixLifecycle.restart(UnixLifecycle.java:73)
	at jenkins.model.Jenkins$25.run(Jenkins.java:4159)
Caused by: java.lang.reflect.InvocationTargetException
	at …
Caused by: java.util.concurrent.TimeoutException: Timeout waiting for task.
	at com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:259)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:91)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$8.call(CpsFlowExecution.java:1207)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$8.call(CpsFlowExecution.java:1198)
	at hudson.security.ACL.impersonate(ACL.java:260)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.suspendAll(CpsFlowExecution.java:1198)
	... 19 more
```

Whatever went wrong, we do not wish this to block shutdown.

Also removes workaround introduced in #19, which is no longer necessary now that our baseline is 2.7.x.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -70,6 +70,7 @@ import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
 import org.jenkinsci.plugins.workflow.support.concurrent.Futures;
+import org.jenkinsci.plugins.workflow.support.concurrent.Timeout;
 import org.jenkinsci.plugins.workflow.support.pickles.serialization.PickleResolver;
 import org.jenkinsci.plugins.workflow.support.pickles.serialization.RiverReader;
 import org.jenkinsci.plugins.workflow.support.storage.FlowNodeStorage;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1305,24 +1305,24 @@ public void pause(final boolean v) throws IOException {
     }
 
     @Restricted(DoNotUse.class)
-    @Terminator public static void suspendAll() throws Exception {
-        ACL.impersonate(ACL.SYSTEM, new NotReallyRoleSensitiveCallable<Void,Exception>() { // TODO Jenkins 2.1+ remove JENKINS-34281 workaround
-            @Override public Void call() throws Exception {
-                LOGGER.fine("starting to suspend all executions");
-                for (FlowExecution execution : FlowExecutionList.get()) {
```
jglick: [diff for reviewers](https://github.com/jenkinsci/workflow-cps-plugin/pull/144/files?w=1#diff-3eec43b20e92e7596aa70d1c7dacc1c6L1308)

PR #147, opened 2017-06-20
[JENKINS-39194] Implement SCM retry count

mmitche:
Honors the retry count of the Jenkins server, and also allows for a custom retry if desired.  Only has effect in non-lightweight cases.


mmitche:
@jglick Fixes https://issues.jenkins-ci.org/browse/JENKINS-39194

mmitche:
@jglick I was having trouble with the custom retry.  The job config save doesn't go through the descriptor's configure, so ensuring that the jenkins backend doesn't always set the retry count to 0 for a job was impossible.

To unblock for now I just removed the advanced custom retry setting.

jglick:
> The job config save doesn't go through the descriptor's configure

Not sure what you mean by that.

> the jenkins backend doesn't always set the retry count to 0

Sorry, what?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -127,15 +134,38 @@ public boolean isLightweight() {
         delegate.setPoll(true);
         delegate.setChangelog(true);
         try (WorkspaceList.Lease lease = computer.getWorkspaceList().acquire(dir)) {
-            delegate.checkout(build, dir, listener, node.createLauncher(listener));
```
jglick: [diff for review](https://github.com/jenkinsci/workflow-cps-plugin/pull/147/files?w=1#diff-a2f3f524db6fdc8eb18f8f3130eae7f6R137)

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/config.jelly
```diff
@@ -24,7 +24,7 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:st="jelly:stapler">
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:st="jelly:stapler" xmlns:p="/lib/hudson/project">
```
jglick: Why? Revert if unused.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -79,6 +82,10 @@ public String getScriptPath() {
     public boolean isLightweight() {
         return lightweight;
     }
+    
+    public int getScmCheckoutRetryCount() {
```
jglick: Currently used only internally, so better inlined.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -127,15 +134,38 @@ public boolean isLightweight() {
...
+                    }
+                } catch (InterruptedIOException e) {
+                    throw (InterruptedException)new InterruptedException().initCause(e);
+                } catch (IOException e) {
+                    // checkout error not yet reported
+                    listener.error(e.toString());
+                }
+                
+                if (retryCount == 0)   // all attempts failed
+                    throw new RunnerAbortedException();
```
jglick: Mentally I had to rearrange this into something like

```java
for (int x = 0; x <= getScmCheckoutRetryCount(); x++) {
    // …
    script = scriptFile.readToString();
    break;
    // …
    listener.getLogger().println("Retrying after 10 seconds");
    Thread.sleep(10000);
}
if (script == null {
    throw …;
}
```

to follow the logic, though I think it is right as written.

jglick: Use `AbortException` not `RunnerAbortedException`. Can probably rearrange the logic a bit so that if we are on the last iteration (the normal case), the original `AbortException` or whatever is rethrown unmodified.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -127,15 +134,38 @@ public boolean isLightweight() {
...
+                    script = scriptFile.readToString();
+                    break;
+                } catch (AbortException e) {
+                    // abort exception might have a null message.
+                    // If so, just skip echoing it.
+                    if (e.getMessage() != null) {
+                        listener.error(e.getMessage());
+                    }
+                } catch (InterruptedIOException e) {
+                    throw (InterruptedException)new InterruptedException().initCause(e);
```
jglick: Why wrap it?

mmitche:
@jglick So if you look at the way that the config-retry.jelly is implemented, it wants to call hasCustomScmCheckoutRetryCount() on the plugin to determine the checkbox, then binds the value of the text box to scmCheckoutRetryCount.  scmCheckoutRetryCount is an Integer which is supposed to be null if the box isn't checked.  This is accomplished for AbstractProject by overriding submit() on AbstractBuild and determining whether the checkbox was checked, otherwise it resets scmCheckoutRetryCount to null.

In our case, if you don't implement similar logic, then when you save the job, scmCheckoutRetryCount gets assigned a default value, which ends up being 0, rather than null.  The right way to fix this I think would be to override Descriptor's configure method, parse the json like AbstractBuild does, and then set the appropriate values.  Saving the job doesn't appear to go through configure() in this case.

Anyways, I could be totally wrong on how to implement the custom retry count UI.  I played around with it for a while but wasn't able to get it to work.  If you have some suggestions, let me know.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
```diff
@@ -157,9 +163,28 @@
...
+    @Test public void retry() throws Exception {
+        // We use an un-initialized repo here to test retry
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        GitStep step = new GitStep(invalidRepo.toString());
+        CpsScmFlowDefinition def = new CpsScmFlowDefinition(step.createSCM(), "flow.groovy");
+        def.setLightweight(false);
+        p.setDefinition(def);
+        // Initially check for the global Jenkins retry count
+        assertEquals(r.jenkins.getScmCheckoutRetryCount(), def.getScmCheckoutRetryCount());
+        // Set the jenkins global retry count to 3 and ensure that they are still equal
```
jglick: uh, 1?

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
```diff
@@ -157,9 +163,28 @@
...
+        r.jenkins.setScmCheckoutRetryCount(1);
+        assertEquals(r.jenkins.getScmCheckoutRetryCount(), def.getScmCheckoutRetryCount());
+        WorkflowRun b = r.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
+        r.assertLogContains("Could not read from remote repository", b);
+        r.assertLogContains("Retrying after 10 seconds", b);
```
jglick: Could I suppose use `waitForMessage`, initialize the repository, then assert then the build completes.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
mmitche: I followed the same form as AbstractBuild: https://github.com/jenkinsci/jenkins/blob/d13b1361e1650494528a7b46f82fe25d7fb5e3c3/core/src/main/java/hudson/model/AbstractBuild.java#L589

I think this is because InterruptedIOException is IOException derived, and InterruptedException is Exception derived.  Maybe plain InterruptedException

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
mmitche: I followed the same form as AbstractBuild: https://github.com/jenkinsci/jenkins/blob/d13b1361e1650494528a7b46f82fe25d7fb5e3c3/core/src/main/java/hudson/model/AbstractBuild.java#L589

I can change it, just want to make sure there is a good reason to do so.

mmitche: Fixed

SpencerMalone:
So what's missing for this to get a merge, besides resolving conflicts?

mmitche:
test this please

mmitche:
Can we get a merge here?  It's a big issue in certain cases (like GitHub being flaky)

mmitche:
@jglick Anything else?

svanoort:
Close/re-open to try to retrigger CI

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -83,7 +86,7 @@ public boolean isLightweight() {
     @DataBoundSetter public void setLightweight(boolean lightweight) {
         this.lightweight = lightweight;
     }
-
+
```
jglick: BTW please configure your text editor to not make gratuitous whitespace changes to otherwise unmodified lines.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
jglick: But _this_ method `throws Exception`, so there is no need for a wrapper.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -127,15 +130,38 @@ public boolean isLightweight() {
...
+                    // abort exception might have a null message.
+                    // If so, just skip echoing it.
+                    if (e.getMessage() != null) {
+                        listener.error(e.getMessage());
+                    }
+                } catch (InterruptedIOException e) {
+                    throw (InterruptedException)new InterruptedException().initCause(e);
+                } catch (IOException e) {
+                    // checkout error not yet reported
+                    listener.error(e.toString());
```
jglick: Should be something like

```
listener.error("Checkout failed").println(Functions.printThrowable(e).trim()); // TODO 2.43+ use Functions.printStackTrace
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -127,15 +130,38 @@ public boolean isLightweight() {
...
+                }
+                
+                if (retryCount == 0)   // all attempts failed
+                    throw new AbortException();
```
jglick: Should specify a message.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
```diff
@@ -157,9 +163,24 @@
...
+    @Issue("JENKINS-39194")
+    @Test public void retry() throws Exception {
+        // We use an un-initialized repo here to test retry
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        GitStep step = new GitStep(invalidRepo.toString());
+        CpsScmFlowDefinition def = new CpsScmFlowDefinition(step.createSCM(), "flow.groovy");
+        def.setLightweight(false);
+        p.setDefinition(def);
+        r.jenkins.setScmCheckoutRetryCount(1);
+        WorkflowRun b = r.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
```
jglick: Would be nice to extend this test to run a second build, this time using `r.waitForMessage("Retrying after 10 seconds", b)` and then initializing the repo and then asserting that the build does succeed with the second checkout.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
mmitche: Okay sounds good, will remove.

svanoort:
Looks like CI issues?

> Installing /home/jenkins/tools/hudson.model.JDK/jdk8/jdk.sh
Can not call launch on a dummy launcher.

kofwang:
Why is this PR not merged?

abayer:
@mmitche Can you resolve the merge conflict, please?

mmitche:
Done

mmitche:
Anything else @svanoort?

svanoort:
Nope, merging now, thanks for the contribution @mmitche (and the reminder now that it is passing)

PR #149, opened 2017-06-24
[JENKINS-45109] Also adjust StepNode.getTypeDisplayName to handle metasteps

jglick:
Amends #148 to address the first bullet point of [JENKINS-45101](https://issues.jenkins-ci.org/browse/JENKINS-45101), which it turns out is not the fault of Blue Ocean, and can in fact be seen in the original **Pipeline Steps** view as well.

@reviewbybees


i386:
@jglick good catch!
@abayer mind if you review?

PR #151, opened 2017-06-29
[JENKINS-31582] Using ArgumentsAction.getResolvedArguments to better handle JENKINS-45109

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/44.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
@@ -225,6 +225,7 @@ public class ArgumentsActionImpl extends ArgumentsAction {
      *   (including the Step or UninstantiatedDescribable)
      */
     @CheckForNull
+    @SuppressWarnings("unchecked")
     Object sanitizeObjectAndRecordMutation(@CheckForNull Object o, @CheckForNull EnvVars vars) {
         // Package scoped so we can test it directly
         Object tempVal = o;
@@ -254,7 +255,13 @@ public class ArgumentsActionImpl extends ArgumentsAction {
         if (modded != tempVal) {
             // Sanitization stripped out some values, so we need to record that and return modified version
             this.isUnmodifiedBySanitization = false;
-            return modded;
+            if (o instanceof UninstantiatedDescribable) {
+                // Need to retain the symbol.
+                UninstantiatedDescribable ud = (UninstantiatedDescribable) o;
+                return new UninstantiatedDescribable(ud.getSymbol(), ud.getKlass(), (Map<String, ?>) modded);
+            } else {
+                return modded;
+            }
         } else {  // Any mutation was just from exploding step/uninstantiated describable, and we can just use the original
             return o;
         }
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
@@ -54,6 +54,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
 
 /**
  * Tests the input sanitization and step persistence here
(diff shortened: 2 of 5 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -394,7 +395,7 @@ public void testUnusualStepInstantiations() throws Exception {
                 " node('master') { \n" +
                 "   writeFile text: 'hello world', file: 'msg.out'\n" +
                 "   step([$class: 'ArtifactArchiver', artifacts: 'msg.out', fingerprint: false])\n "+
-                "   withEnv(['CUSTOM=val']) {\n"+  //Symbol-based, because withEnv is a metastep
+                "   withEnv(['CUSTOM=val']) {\n"+  //Symbol-based, because withEnv is a metastep; TODO huh? no it is not
```
abayer: That i don't 😖 .

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -423,4 +424,25 @@ public void testUnusualStepInstantiations() throws Exception {
         Assert.assertEquals("CUSTOM=val", (String)((ArrayList) ob).get(0));
         testDeserialize(run.getExecution());
     }
+
+    @Test
+    public void testReallyUnusualStepInstantiations() throws Exception {
```
abayer: Ok, I really like this test name. =)

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -254,7 +255,13 @@ Object sanitizeObjectAndRecordMutation(@CheckForNull Object o, @CheckForNull Env
         if (modded != tempVal) {
             // Sanitization stripped out some values, so we need to record that and return modified version
             this.isUnmodifiedBySanitization = false;
-            return modded;
+            if (o instanceof UninstantiatedDescribable) {
+                // Need to retain the symbol.
```
jglick: This is a fix to what is _stored_. Without this, even a fleshed-out version of `getResolvedArguments` would have no hope, because previously the storage would be something like

    delegate → [<anonymous> → 'TEST-*.xml']

with no record being kept that the original delegate was in fact `junit` ~ `JUnitResultArchiver`.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
jglick: Not sure what @svanoort thought he was testing in this case. If the idea was to test a symbol-based block-scoped metastep (like `configFileProvider`), it is not doing that.

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -423,4 +424,25 @@ public void testUnusualStepInstantiations() throws Exception {
...
+        WorkflowJob job = r.jenkins.createProject(WorkflowJob.class, "unusualInstantiation");
+        job.setDefinition(new CpsFlowDefinition(
+                " node() {\n" +
+                "   writeFile text: 'hello world', file: 'msg.out'\n" +
+                "   step(new hudson.tasks.ArtifactArchiver('msg.out'))\n" + // note, not whitelisted
```
jglick: Seems that this case was implicitly handled in the code but not tested.

src/test/java/org/jenkinsci/plugins/workflow/cps/nodes/StepNodeTest.java
```diff
@@ -78,4 +83,87 @@
...
+            "node {\n" +
+            "  wrap([$class: 'ConfigFileBuildWrapper', managedFiles: []]) {\n" +
+            "    writeFile text: '''<testsuite name='a'><testcase name='c'><error>failed</error></testcase></testsuite>''', file: 'x.xml'\n" +
+            "    step([$class: 'JUnitResultArchiver', testResults: 'x.xml'])\n" +
+            "  }\n" +
+            "}", true));
+        WorkflowRun b = r.assertBuildStatus(Result.UNSTABLE, p.scheduleBuild2(0));
+        List<FlowNode> coreStepNodes = new DepthFirstScanner().filteredNodes(b.getExecution(), new NodeStepTypePredicate("step"));
+        assertThat(coreStepNodes, hasSize(1));
+        assertEquals("junit", coreStepNodes.get(0).getDisplayFunctionName());
```
jglick: We may as well show the nice view even if you have not yet updated your script to take advantage of simplified syntax.

src/test/java/org/jenkinsci/plugins/workflow/cps/nodes/StepNodeTest.java
```diff
@@ -78,4 +83,87 @@
...
+            "node {\n" +
+            "  wrap(new org.jenkinsci.plugins.configfiles.buildwrapper.ConfigFileBuildWrapper([])) {\n" +
+            "    writeFile text: '''<testsuite name='a'><testcase name='c'><error>failed</error></testcase></testsuite>''', file: 'x.xml'\n" +
+            "    step(new hudson.tasks.junit.JUnitResultArchiver('x.xml'))\n" +
+            "  }\n" +
+            "}", false));
+        WorkflowRun b = r.assertBuildStatus(Result.UNSTABLE, p.scheduleBuild2(0));
+        List<FlowNode> coreStepNodes = new DepthFirstScanner().filteredNodes(b.getExecution(), new NodeStepTypePredicate("step"));
+        assertThat(coreStepNodes, hasSize(1));
+        assertEquals("junit", coreStepNodes.get(0).getDisplayFunctionName());
```
jglick: A corner case but we get it for free anyway.

PR #154, opened 2017-06-30
The argumentsToString parameter was not, in fact, checked for null

jglick:
Cf. https://github.com/jenkinsci/workflow-step-api-plugin/pull/27.

@reviewbybees


jglick:
@reviewbybees done

jglick:
This is just in test code.

PR #158, opened 2017-07-27
[JENKINS-44548] NPE from loadProgramFailed

jglick:
[JENKINS-44548](https://issues.jenkins-ci.org/browse/JENKINS-44548)

`CpsFlowExecution.storage` and `startNodes` are initialized after `onLoad` by `initializeStorage`. But `flowStartNodeActions`, normally initialized from the constructor, will not be reinitialized after a load from disk, so this could be null in case we are in `loadProgramFailed`.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
@@ -94,13 +94,15 @@ final class FlowHead implements Serializable {
     }
 
     void newStartNode(FlowStartNode n) throws IOException {
-        for (Action a : execution.flowStartNodeActions) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
```diff
@@ -94,13 +94,15 @@ public CpsFlowExecution getExecution() {
     }
 
     void newStartNode(FlowStartNode n) throws IOException {
-        for (Action a : execution.flowStartNodeActions) {
```
jglick: [diff for review](https://github.com/jenkinsci/workflow-cps-plugin/pull/158/files?w=1)

jglick:
@reviewbybees done

PR #159, opened 2017-08-04
Add various plugin dependency version POM properties

abayer:
I'm working on tooling to do testing of core + groovy-sandbox +
groovy-cps + workflow-cps + script-security + the rest of the main
Pipeline plugins against bleeding-edge versions of each other and
different versions of Groovy. To do this, it'd be a lot easier if the
dependencies between the various things I'm building/testing had their
versions controlled by properties.

cc @reviewbybees


jglick:
ditto https://github.com/jenkinsci/maven-hpi-plugin/pull/66

PR #160, opened 2017-08-08
[JENKINS-38052] Test verifying that JENKINS-38052 is fixed.

abayer:
[JENKINS-38052](https://issues.jenkins-ci.org/browse/JENKINS-38052)

This works now, so let's add a test to make sure it keeps working.

cc @reviewbybees


abayer:
Replaced by #163

PR #161, opened 2017-08-08
[JENKINS-27916] Test verifying use of GStrings in map keys

abayer:
[JENKINS-27916](https://issues.jenkins-ci.org/browse/JENKINS-27916)

This works properly now, so let's test it.

cc @reviewbybees


jglick:
This is testing something unrelated to the originally reported issue, which was about `Step` arguments—which should have been fixed in #123.

abayer:
Replaced by #163 - and ah, ok. I'll leave the test in anyway. Can't hurt. =)

PR #162, opened 2017-08-08
[JENKINS-41248] Verify that explicit setter methods take priority.

abayer:
[JENKINS-41248](https://issues.jenkins-ci.org/browse/JENKINS-41248)

This works now, let's test it.

cc @reviewbybees


abayer:
Replaced by #163

PR #165, opened 2017-08-15
Deleting some obsolete Groovy 1.x code

jglick:
@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1135,33 +1130,28 @@ private static void cleanUpGlobalClassSet(@Nonnull Class<?> clazz) throws Except
         Field globalClassSetF = classInfoC.getDeclaredField("globalClassSet");
         globalClassSetF.setAccessible(true);
         Object globalClassSet = globalClassSetF.get(null);
-        try { // Groovy 1
-            globalClassSet.getClass().getMethod("remove", Object.class).invoke(globalClassSet, clazz); // like Map but not
-            LOGGER.log(Level.FINER, "cleaning up {0} from GlobalClassSet", clazz.getName());
-        } catch (NoSuchMethodException x) { // Groovy 2
-            try {
```
jglick: [diff for review](https://github.com/jenkinsci/workflow-cps-plugin/pull/165/files?w=1#diff-3eec43b20e92e7596aa70d1c7dacc1c6L1138)

PR #168, opened 2017-08-22
[JENKINS-46358] Tests for StringGroovyMethods transformations

abayer:
[JENKINS-46358](https://issues.jenkins-ci.org/browse/JENKINS-46358)

Downstream of https://github.com/cloudbees/groovy-cps/pull/70 and
https://github.com/jenkinsci/script-security-plugin/pull/145

cc @reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/CpsStringGroovyMethodsTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/CpsStringGroovyMethodsTest.java
@@ -5,8 +5,10 @@ import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 
+@Issue("JENKINS-46358")
 public class CpsStringGroovyMethodsTest {
     @ClassRule
     public static JenkinsRule r = new JenkinsRule();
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/CpsStringGroovyMethodsTest.java
```diff
@@ -0,0 +1,67 @@
+package org.jenkinsci.plugins.workflow;
+
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.JenkinsRule;
+
+public class CpsStringGroovyMethodsTest {
```
jglick: `@Issue`

src/test/java/org/jenkinsci/plugins/workflow/CpsStringGroovyMethodsTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @ClassRule
+    public static JenkinsRule r = new JenkinsRule();
```
jglick: Trying to keep the build time for this plugin to a sane level I see…

abayer: Ahyup. Since this is all purely self contained tests with no side effects, why not?

PR #169, opened 2017-08-23
[JENKINS-46391] Test verifying CPS-specific ~/foo/ behavior

abayer:
[JENKINS-46391](https://issues.jenkins-ci.org/browse/JENKINS-46391)

Downstream of https://github.com/jenkinsci/script-security-plugin/pull/146

cc @reviewbybees


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -224,4 +224,11 @@ public void traitsSandbox() throws Exception {
         jenkins.assertLogContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance", b);
     }
 
+    @Issue("JENKINS-46391")
+    @Test
+    public void tildePattern() throws Exception {
```
jglick: Not a task for this PR, but sometime this test suite needs to be split up and rearranged more logically, since of course it is not testing `CpsFlowDefinition` per se.

PR #170, opened 2017-09-01
Defer writing FlowNode actions

svanoort:
Finally we have the payoff!  This reduces the amount of disk serialization calls significantly. 

Next up:
- [x] Deploy SNAPSHOT for downstream use, as 2.41-deferactionwrite-20170908.150443-1
- [x] Downstream of this, in workflow-job plugin, defer write of timing action
- [x] Ensure that persist is invoked eventually in most cases (finally, etc)

Downstream of:

- https://github.com/jenkinsci/workflow-step-api-plugin/pull/29
- https://github.com/jenkinsci/workflow-api-plugin/pull/49
- https://github.com/jenkinsci/workflow-support-plugin/pull/43

# Result: 
**In IO-~50% more FlowNode throughput, or put another way, ~33% reduction in build time in IO-bound situations.**

Benchmark: running the follow pipeline, ramping up from 1 to 6 concurrent runs evenly over a minute, and starting a new build when one completes.   Plateau on the left is with a modern core and latest pipeline plugin versions and on the right is with downstream plugin versions + this one.  

screen shot 2017-09-01 at 5 59 25 pm

screen shot 2017-09-01 at 6 00 02 pm

Pipeline: 

```groovy
/*
Simple echo steps
*/
properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '100')), pipelineTriggers([])])

for (int i=0; i<10; i++) {
	stage ("Stage $i") {
		// Some echos.
		for (int j=0; j < 20; j++) {
			// Put in some logging.
            echo "Here goes echo number $j"
		}
	}
}
```

Combined with finishing off the log handling rewrite work, we will probably end up cutting the IO burden of pipeline in half AKA double performance in IO-bound situations. 

**TODO:**
- [ ] Create a test utility that can report on which FlowNodes are persisted to disk and which are not, and compare persisted version to actual version. (maybe look to see if actions are trivially equal)
    - Consider that we may have different persistence models for FlowNodes in the future, i.e. nonpersistent, fully persistent, etc
- [ ] Instrument testcases for error in pipeline syntax, error in pipeline, normal run of steps, and running steps which defer writing their actions and do it inside the StepExecution.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.31</version>
+        <version>2.29</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-cps</artifactId>
-    <version>2.40-SNAPSHOT</version>
+    <version>2.40-deferactionwrite-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Groovy</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Groovy+Plugin</url>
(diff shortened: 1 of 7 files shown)


svanoort:
Closed in deference to #188

PR #171, opened 2017-09-04
[JENKINS-26148] Using default implementation of StepExecution.stop

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/30.

@reviewbybees


jglick:
@reviewbybees done

PR #173, opened 2017-09-18
[FIXED JENKINS-44619] Don't allow replays of nonbuildable jobs

abayer:
[JENKINS-44619](https://issues.jenkins-ci.org/browse/JENKINS-44619)

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,7 +139,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.10</version>
+            <version>2.11.2</version>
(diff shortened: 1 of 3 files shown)


abayer:
@jglick What error would you recommend?

jglick:
> What error would you recommend?

Could use a 409, as [here](https://github.com/jenkinsci/jenkins/blob/f09d00dd32ea351105ef99456f3af1883d3ca8da/core/src/main/java/jenkins/model/ParameterizedJobMixIn.java#L196-L198).

abayer:
@svanoort Updated to throw an http error.

svanoort:
@abayer Looking good here!

PR #174, opened 2017-09-18
Use block info cache

svanoort:
Consumes https://github.com/jenkinsci/workflow-api-plugin/pull/50

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -20,6 +20,7 @@ import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
+import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.steps.BodyExecution;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -194,15 +195,21 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
         t.getExecution().runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
             @Override public void onSuccess(CpsThreadGroup g) {
                 try {
+                    if (g.threads.size() == 1) {
```
svanoort: I can't optimize this with the block lookup cache since the body isn't necessarily enclosing (boo), but at least can make it faster in the common case.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -247,13 +254,20 @@ public boolean cancel(final CauseOfInterruption... causes) {
             t.getExecution().runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
                 @Override
                 public void onSuccess(CpsThreadGroup g) {
+                    if (g.threads.size() == 1) {
```
svanoort: I can't optimize this with the block lookup cache since the body isn't necessarily enclosing (boo), but at least can make it faster in the common case.

svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -194,15 +195,21 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
         t.getExecution().runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
             @Override public void onSuccess(CpsThreadGroup g) {
                 try {
+                    if (g.threads.size() == 1) {
+                        result.set(Collections.singletonList(g.threads.firstEntry().getValue().getStep()));
```
jglick: First of all, you are failing to check for a null return value from `getStep`. So you might be returning `[null]`, which is illegal.

That aside, I doubt this optimization is correct. You are making no effort to verify that the thread’s `head` is in fact a child of this body’s start node. Now if the program in fact has only one active thread, then its single head should be a child of any _active_ body execution. But there is no documented guarantee that this method can be called only on an active body.

This method seems like a good candidate for consuming the new API rather than doing a linear scan, so I am not sure why you are not doing so.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -247,13 +254,20 @@ public boolean cancel(final CauseOfInterruption... causes) {
             t.getExecution().runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
                 @Override
                 public void onSuccess(CpsThreadGroup g) {
+                    if (g.threads.size() == 1) {
+                        // Early exit for trivial case
+                        g.threads.firstEntry().getValue().stop(stopped);
```
jglick: Again this is potentially wrong.

(Anyway how often is `cancel` called? Not often at all. So not a priority for little optimizations.)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -561,6 +561,7 @@ int approximateNodeCount() {
 
     protected void initializeStorage() throws IOException {
         storage = createStorage();
+        this.internalGraphLookup = super.getInternalGraphLookup();
```
jglick: Why? The getter initializes it on demand.

svanoort:
Closing out until I can come back and re-assess parallel bottlenecks again with an attached profiler.

PR #176, opened 2017-09-21
Consume temporary version of groovy-cps that reduces memory

svanoort:
Demonstrates that https://github.com/cloudbees/groovy-cps/pull/74 works and does not introduce errors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
         <!-- Temporary until https://github.com/cloudbees/groovy-cps/pull/74 is released and we can use final-->
-        <groovy-cps.version>1.20-20170921.213551-1</groovy-cps.version>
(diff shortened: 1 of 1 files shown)


svanoort:
Sigh, rebuild for what looks like builder issues

svanoort:
Passes locally, doing something to trigger a rebuild so we can see if it goes 💥  in CI.

jglick:
Regarding the Surefire error attributed (perhaps falsely) to `WorkflowJobNonRestartingTest`: not sure what went on.

    lineNumber: 2524917

suggests perhaps a huge list of repeating error messages. The [relevant log](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-176/1/execution/node/94/log/) ends with

```
00:29:52.616 [INFO] Running org.jenkinsci.plugins.workflow.cps.DSLTest
00:34:15.943 Sending interrupt signal to process
```

which suggests a hung test. In that case you would want to pick up https://github.com/jenkinsci/jenkins-test-harness/pull/67, most easily by just updating to the 2.35 parent POM.

abayer:
Closing in favor of #177

PR #177, opened 2017-09-22
[JENKINS-47064] Verify fix for BooleanClosureWrapper

abayer:
[JENKINS-47064](https://issues.jenkins-ci.org/browse/JENKINS-47071)

Also bumps parent POM for Intellij.

Downstream of https://github.com/cloudbees/groovy-cps/pull/75

cc @reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
@@ -282,7 +282,7 @@ public class CpsFlowDefinition2Test extends AbstractCpsFlowTest {
     @Test
     public void booleanClosureWrapperFromDGM() throws Exception {
         WorkflowJob job = jenkins.jenkins.createProject(WorkflowJob.class, "p");
-        job.setDefinition(new CpsFlowDefinition("['a', 'b'].every { sleep 1; return it != null }\n", true));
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -277,4 +277,12 @@ public void rhsOfDeclarationSandboxedInCPS() throws Exception {
         jenkins.assertBuildStatus(Result.FAILURE, b);
         jenkins.assertLogContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance", b);
     }
+
+    @Issue("JENKINS-47064")
+    @Test
+    public void booleanClosureWrapperFromDGM() throws Exception {
+        WorkflowJob job = jenkins.jenkins.createProject(WorkflowJob.class, "p");
+        job.setDefinition(new CpsFlowDefinition("['a', 'b'].every { sleep 1; return it != null }\n", true));
```
jglick: Maybe you also want to assert the return value of `every`?

abayer: Wouldn't hurt - we've already got tests over in `CpsDefaultGroovyMethodsTest` that verify the expected behavior without the `sleep`, but hey.

PR #178, opened 2017-09-25
[JENKINS-47105] Fixing test failure in loaderReleased

jglick:
[JENKINS-47105](https://issues.jenkins-ci.org/browse/JENKINS-47105)

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'], failFast: false)
+buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'])
+buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'], failFast: false)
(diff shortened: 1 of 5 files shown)


jglick:
Probably a good candidate for squashing, since some of the intermediate commits are deliberately junk.

jglick:
Hmm, [this](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-178/3/execution/node/102/log/) does not look good…

jglick:
Nor does [this](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-178/3/execution/node/60/log/).

svanoort:
AFAICT it looks reasonable, assuming tests pass -- though it does seem to be running awfully long.

svanoort:
I take it back, that last one looks scary though it *could* be some flakiness in CI.  The tests shouldn't take so long to run though -- something is off.

jglick:
Indeed something is off. I thought I had diagnosed the issue but it does not seem that I did.

jglick:
I now think https://github.com/jenkinsci/jenkins-test-harness/pull/67 is responsible for the problem, but have not yet figured out why.

jglick:
Hmm. Locally, the test passes if _either_ I stop running `trustedShell` first, _or_ I revert that test timeout change. But on CI it continues to fail _sometimes_ even when I stop running `trustedShell` first.

jglick:
Looking better. Seems that whatever leaked, the leaks persist across Surefire reused forks of the JVM. Safer to have each test suite run completely independently, even if it adds a bit of overhead.

jglick:
@reviewbybees done

PR #185, opened 2017-10-23
Partial support for lightweight checkout changelogs

stephenc:


stephenc:
@jglick @abayer @svanoort Here is a Proof Of Concept WIP on getting the lightweight checkout changelog.

*THIS IS NOT COMPLETE*

I do not know how to wire up the changelog file to the WorkflowRun as we do not have access to the  `WorkflowRun.SCMCheckout` class from here and hence there is no way to track.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -96,10 +101,26 @@ public boolean isLightweight() {
         }
         Run<?,?> build = (Run<?,?>) _build;
         if (isLightweight()) {
-            try (SCMFileSystem fs = SCMFileSystem.of(build.getParent(), scm)) {
+            try (SCMFileSystem fs = SCMFileSystem.of(build.getParent(), scm, SCMRevisionAction.getRevision(build))) {
```
stephenc: Technically unrelated, but signals a bug waiting to happen... if there are commits concurrent with the build, we could end up checking out the wrong (newer) revision of the `Jenkinsfile` than the checkout will check out if we do not at least attempt to resolve the revision

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
```diff
@@ -161,6 +167,44 @@
...
+        sampleRepo.git("add", "flow.groovy");
+        sampleRepo.git("commit", "--message=update");
+        // we add the SCMRevisionAction because we are hijacking a plain SCM as a proxy for a SCMSource backed SCM
+        b = r.assertBuildStatusSuccess(p.scheduleBuild2(0,
+                new SCMRevisionAction(new AbstractGitSCMSource
+                .SCMRevisionImpl(new SCMHead("master"), sampleRepo.head()
+                ))
+        ));
+        changeSets = b.getChangeSets();
+        assertEquals(1, changeSets.size());
```
stephenc: So this is currently 0 not 1 because `WorkflowRun` is unaware of the additional `changelog0.xml`. What probably needs to happen is that there would be a higher level action that could be attached in a more generic fashion and then `WorkflowRun` would be aware of that action and use it in addition to `WorkflowRun.SCMCheckout` in order to build the changeSets

stephenc:
FTR I am expecting one of @jglick @svanoort @abayer to pick this up. I do not intend on driving this further myself

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
jglick: `CpsScmFlowDefintion` does not support exact revision matching between `Jenkinsfile` and `checkout scm`. It _could_, if using `scm-api` (there is a filed RFE for it), but such a change does not belong here I think; should be done separately, with test coverage.

jglick: [JENKINS-36453](https://issues.jenkins-ci.org/browse/JENKINS-36453) FTR

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -96,10 +101,26 @@ public boolean isLightweight() {
...
                 if (fs != null) {
                     String script = fs.child(scriptPath).contentAsString();
                     listener.getLogger().println("Obtained " + scriptPath + " from " + scm.getKey());
+                    // a heavyweight checkout will populate the changelog only if there is a previous build
+                    // to compare with
```
jglick: That would be true of a lightweight checkout as well. A changelog is only meaningful when there is some previous build.

stephenc: Which is why I added the comment... so that the `if (prev != null)` makes sense... we could choose to say this is the first build, so "oh here's the last 1000 changes"... but that would deviate from the behaviour of a heavyweight checkout

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
stephenc: Well if you attach an `SCMRevisionAction` to the build in this step then you get [JENKINS-36453](https://issues.jenkins-ci.org/browse/JENKINS-36453) "for free"... but as this PR is mostly a PoC to show what should be done...

stephenc:
> Is this solving some filed RFE/bug?

A number of issues in the issue tracker with users complaining about this, e.g. 

https://issues.jenkins-ci.org/browse/JENKINS-46431

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
jglick: Cannot `SCMRevisionAction` be used for this purpose? `WorkflowRun` would need to use `scm-api` and compare to the last build, instead of that being done here.

dwnusbaum:
I am going to close this PR, if anyone wants to pick it back up and resolve the conflicts feel free to reopen.

PR #186, opened 2017-11-06
Rename ReplayCommand into ReplayPipelineCommand

batmat:
Small consistency improvement: remove then also getName() override to rely on default class name to CLI command name conversion.
(which will stay "replay-pipeline").

I have come across this trying to assess how often people were overriding getName() and apparently this is the only occurrence.
As I think there's no reason anyone would inherit/depend on this, I guess we can safely just rename it for consistency.
(Good thing being when crawling codebase, one will be able to translate the command name into a probable class name, and find it right away using the type-search feature of one's IDE).

@reviewbybees esp. @jglick


batmat:
@reviewbybees done

batmat:
@jglick do you need anything else from my side before we can merge? Thanks a lot.

batmat:
Thanks @abayer & @jglick !

PR #187, opened 2017-11-07
[FIXED JENKINS-47339] Add build-implied REBUILD permission.

abayer:
[JENKINS-47339](https://issues.jenkins-ci.org/browse/JENKINS-47339)

Allows rebuilding a Pipeline run with the exact same script.

This will be used directly by Blue Ocean for its rebuild functionality, which doesn't need to be able to make changes to the script, so shouldn't require `Item.CONFIGURE` or `ReplayAction.REPLAY`. Instead, it'll require either `Item.BUILD` or `ReplayAction.REBUILD`.

cc @reviewbybees @i386

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction/index.jelly
@@ -4,27 +4,38 @@
     <l:layout title="Replay ${it.owner.displayName}" norefresh="true">
         <st:include page="sidepanel.jelly" it="${it.owner}"/>
         <l:main-panel>
-            <j:if test="${it.enabled}">
+            <j:if test="${it.enabled or it.rebuildEnabled}">
                 <h1>Replay ${it.owner.displayName}</h1>
                 <p>
                     <j:out value="${%blurb}"/>
                 </p>
-                <f:form action="run" method="POST" name="config">
-                    <f:entry field="mainScript" title="Main Script">
-                      <wfe:workflow-editor script="${it.originalScript}" checkUrl="${rootURL}/${it.owner.url}${it.urlName}/checkScript" checkDependsOn=""/>
-                    </f:entry>
-                    <j:forEach var="loadedScript" items="${it.originalLoadedScripts.entrySet()}">
-                        <f:entry field="${loadedScript.key.replace('.', '_')}" title="${loadedScript.key}">
-                            <wfe:workflow-editor script="${loadedScript.value}" checkUrl="${rootURL}/${it.owner.url}${it.urlName}/checkScript" checkDependsOn=""/>
-                        </f:entry>
-                    </j:forEach>
-                    <f:block>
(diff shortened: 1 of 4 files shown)


jtnord:
> This will be used directly by Blue Ocean for its rebuild functionality, which doesn't need to be able to make changes to the script

Why does it just not trigger a build then?
Most of the times bo does this it is in fact wrong and should just ask for a build. 
https://issues.jenkins-ci.org/browse/JENKINS-42353

abayer:
@jtnord To retain the parameters and SCM revision.

abayer:
Seems like repo.azure.jenkins.io has a busted cache of the 2.73.1 war file. Fun!

abayer:
@jglick Fair enough.

src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -189,6 +195,15 @@ private static boolean canReplay(WorkflowRun b, String user) {
         });
     }
 
+    private static boolean canRebuild(WorkflowRun b, String user) {
+        final ReplayAction a = b.getAction(ReplayAction.class);
+        return ACL.impersonate(User.get(user).impersonate(), new NotReallyRoleSensitiveCallable<Boolean,RuntimeException>() {
```
jglick: BTW `ACL.as` is generally easier to use.

src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -271,4 +286,51 @@ private static boolean canReplay(WorkflowRun b, String user) {
...
+                    story.j.waitUntilNoActivity();
+                    b2 = p.getBuildByNumber(2);
+                    assertNotNull(b2);
+                }
+                story.j.assertLogContains("script to rebuild", story.j.assertBuildStatusSuccess(b2));
+                ReplayCause cause = b2.getCause(ReplayCause.class);
+                assertNotNull(cause);
+                assertEquals(1, cause.getOriginalNumber());
+                assertEquals(b1, cause.getOriginal());
+                assertEquals(b2, cause.getRun());
```
jglick: Might be interesting to verify that, say, parameters are the same as in the original.

src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction/index.jelly
```diff
@@ -4,27 +4,38 @@
     <l:layout title="Replay ${it.owner.displayName}" norefresh="true">
         <st:include page="sidepanel.jelly" it="${it.owner}"/>
         <l:main-panel>
-            <j:if test="${it.enabled}">
+            <j:if test="${it.enabled or it.rebuildEnabled}">
                 <h1>Replay ${it.owner.displayName}</h1>
                 <p>
                     <j:out value="${%blurb}"/>
                 </p>
-                <f:form action="run" method="POST" name="config">
```
jglick: [diff for reviewers](https://github.com/jenkinsci/workflow-cps-plugin/pull/187/files?w=1#diff-bea9feb20818b01fa16e0538ec82cccfR12)

PR #189, opened 2017-11-27
[FIXED JENKINS-46597] IteratorHack needs to special-case SortedMap

abayer:
[JENKINS-46597](https://issues.jenkins-ci.org/browse/JENKINS-46597)

For whatever reason that I can't exactly determine yet, Groovy
categories seem to, at least in some scenarios, treat `Map` differently
than `SortedMap`. I know, weird. Anyway, `IteratorHack#entrySet(Map)`
and friends would fire for `HashMap` but not for `TreeMap`. Adding
these shim methods fixes `entrySet`, `keySet`, and `values`.

Replaces https://github.com/jenkinsci/workflow-scm-step-plugin/pull/23

cc @reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHackTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHackTest.java
@@ -127,6 +127,45 @@ public class IteratorHackTest {
         });
     }
 
+    @Issue("JENKINS-46597")
+    @Test public void treeMapIterator() {
+        rr.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = rr.j.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                        "def map = new TreeMap<String,Integer>()\n" +
+                                "map.put('one', 1)\n" +
+                                "map.put('two', 2)\n" +
+                                "@NonCPS def entrySet(m) {m.collect {k, v -> [key: k, value: v]}}; for (def e in entrySet(map)) {echo \"running flattened loop on ${e.key} -> ${e.value}\"; semaphore \"C-${e.key}\"}\n" +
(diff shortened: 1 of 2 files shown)


abayer:
I'm also wondering if we should be moving `IteratorHack` to `groovy-cps` alongside the `CpsDefaultGroovyMethods` and friends, since we use them in basically the same way and context...

svanoort:
I kind of like them here, but there's not a specific technical justification, it's just easier to work with that and test it in the workflow-cps codebase.

jglick:
> wondering if we should be moving `IteratorHack` to `groovy-cps`

Yes I have considered that too in the past, was just too lazy to do it.

src/test/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHackTest.java
```diff
@@ -127,6 +127,45 @@
...
+    @Issue("JENKINS-46597")
+    @Test public void treeMapIterator() {
+        rr.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = rr.j.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                        "def map = new TreeMap<String,Integer>()\n" +
+                                "map.put('one', 1)\n" +
+                                "map.put('two', 2)\n" +
+                                "@NonCPS def entrySet(m) {m.collect {k, v -> [key: k, value: v]}}; for (def e in entrySet(map)) {echo \"running flattened loop on ${e.key} -> ${e.value}\"; semaphore \"C-${e.key}\"}\n" +
```
jglick: Could probably drop this copy-pasted clause, not really testing anything new.

jglick:
> easier to work with

Well. If in `groovy-cps`, we might need to add a test utility that tried serializing the program after every operation, and I am not sure how much work that would be. The payoff would be that fixes would be closer to the source, and would be true unit tests, much faster than `JenkinsRule` tests.

PR #194, opened 2017-12-18
CpsScmFlowDefinition should add a WorkspaceActionImpl corresponding t…o the directory it actually acquired

jglick:
Buglet I just noticed.

@reviewbybees


jglick:
@reviewbybees done

PR #196, opened 2018-01-03
Warning about misuse of this extension point

jglick:
After https://github.com/jenkinsci/docker-workflow-plugin/pull/127, another attempt to steer the innocent away from a commonly abused API.

@reviewbybees


jglick:
> those cases are and should be rare

Right, there are legitimate reasons for using `GlobalVariable` when you know what you are doing and why. I am just trying to discourage people from grabbing at it as “oh this is the API for defining new Pipeline features”. For example, not sure if you were pulled into diagnosing the awful problems surrounding designs like [this one](https://github.com/jenkinsci/artifactory-plugin/blob/f99ee1194db987d4f45be32bc79f8806763538e4/src/main/java/org/jfrog/hudson/pipeline/dsl/ArtifactoryPipelineGlobal.java#L95-L100).

Anyway, @reviewbybees done.

abayer:
What in the actual hell.

PR #201, opened 2018-02-01
Pull in new Structs to resolve test failures

svanoort:
May replace https://github.com/jenkinsci/workflow-cps-plugin/pull/200 if it only needed the structs fix. 

PR to hit it with CI tests.

PR #202, opened 2018-02-02
Optimize fetching the sole parameter to avoid needless reflection and possible classloading

svanoort:
Removes yet another cause of needless runtime Reflection/classloading when running pipeline.  Should result in running steps much faster generally and with less disk hits from classloading + less lock contention.

Similar to https://github.com/jenkinsci/structs-plugin/pull/31 but on the DSL side (using the DescribableModel more efficiently with what is provided) -- also related to https://github.com/jenkinsci/structs-plugin/pull/30. 

Incidentally includes https://github.com/jenkinsci/workflow-cps-plugin/pull/201 for the fixed Structs.

Justification: see attached image showing Object monitor thread contention when running 40x concurrent trivial pipelines -- this is one of the top 10 leading bottlenecks.

screen shot 2018-02-02 at 11 15 04 am

**Remaining locking bottlenecks for future PRs:**

* Script Security - ProxyWhiteList.delegates ArrayList during permitsXXXX calls -- addressed by https://github.com/jenkinsci/script-security-plugin/pull/179 which uses caching so locks are only held very briefly (rather than while doing O(n) a full scan of all possible Whitelists). 
* Reusing an XStream instance (mandatory internal locking) - [JENKINS-19561](https://issues.jenkins-ci.org/browse/JENKINS-19561) - a larger bottleneck when your storage is fast, but less trivial to solve (I have plans to do a limited solution in the near future). 

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -458,7 +458,13 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         boolean singleArgumentOnly = false;
         try {
             DescribableModel<?> stepModel = DescribableModel.of(d.clazz);
-            singleArgumentOnly = stepModel.hasSingleRequiredParameter() && stepModel.getParameters().size() == 1;
+            if (stepModel != null) {
+                singleArgumentOnly = stepModel.hasSingleRequiredParameter() && stepModel.getParameters().size() == 1;
+                if (singleArgumentOnly) {  // Can fetch the one argument we need
+                    String paramName = stepModel.getSoleRequiredParameter().getName();
+                    return parseArgs(arg, d.takesImplicitBlockArgument(), paramName, singleArgumentOnly);
+                }
+            }
         } catch (NoStaplerConstructorException e) {
             // Ignore steps without databound constructors and treat them as normal.
         }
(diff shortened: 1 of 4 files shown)


svanoort:
@jglick Apologize for requesting yet another review from you - you're the only person likely to catch some subtle issue here if present (unlikely).  And thankfully it's a very short review.

svanoort:
Odd test failure that didn't occur locally -- seems *probably* unrelated but investigating.

svanoort:
And... resolved.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -458,7 +458,14 @@ static NamedArgsAndClosure parseArgs(Object arg, StepDescriptor d) {
         boolean singleArgumentOnly = false;
         try {
             DescribableModel<?> stepModel = DescribableModel.of(d.clazz);
-            singleArgumentOnly = stepModel.hasSingleRequiredParameter() && stepModel.getParameters().size() == 1;
+            if (stepModel != null) {
```
jglick: Why? `of` never returns null.

svanoort: FindBugs.

svanoort: Eh, gratuitous

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
```diff
@@ -7,6 +7,8 @@
 import java.util.Arrays;
 import static java.util.Arrays.*;
 import java.util.List;
+
+import jenkins.model.Jenkins;
```
jglick: gratuitous

PR #204, opened 2018-02-18
Edge Case: CpsScript invokeMethod does not execute closures defined in the script binding

steven-terrana:
Method calls made in a pipeline get routed to CpsScript's invokeMethod before being executed.  

**Currently**:
The method checks to see if the method in question has been defined as a GlobalVariable and if not - checks the DSL where if not present, an error will be thrown. 

Given this flow - there is no room for user defined closures injected into the script binding to get executed in the pipeline. 

**Fix**:  
Check the script binding for variables named the same as the method name attempting to be invoked.  If the variable exists and it is of type CpsClosure2, execute the closure and pass in the arguments.  

User defined closures should take precedence over library methods or pipeline DSL steps as they were defined in the most local context to the script execution. 

**Reproduce:**
~~~ groovy
my_closure = { println "my closure!" }
my_closure()
~~~
fails with:  _java.lang.NoSuchMethodError: No such DSL method 'my_closure' found among steps_

but 
~~~ groovy
def my_closure = { println "my closure!" }
my_closure()
~~~
works as expected.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
@@ -91,30 +91,30 @@ public abstract class CpsScript extends SerializableScript {
     @Override
     public final Object invokeMethod(String name, Object args){
 
-    		// check for user defined closures in the script binding
-    		if (getBinding().hasVariable(name)){
-    		    Object local_var = getBinding().getVariable(name);
-    		    if (local_var instanceof CpsClosure2){
-                    CpsClosure2 local_closure = (CpsClosure2) local_var;
-    		        return local_closure.call(args);
-    		    }
-    		}
+        // check for user defined closures in the script binding
+        if (getBinding().hasVariable(name)){
+            Object local_var = getBinding().getVariable(name);
+            if (local_var instanceof CpsClosure2){
+                CpsClosure2 local_closure = (CpsClosure2) local_var;
+                return local_closure.call(args);
+            }
+        }
     		
-			// if global variables are defined by that name, try to call it.
-		    // the 'call' convention comes from Closure
-	        GlobalVariable v = GlobalVariable.byName(name, $buildNoException());
-	        if (v != null) {
-	            try {
-	                Object o = v.getValue(this);
-	                return InvokerHelper.getMetaClass(o).invokeMethod(o, "call", args);
-	            } catch (Exception x) {
-	                throw new InvokerInvocationException(x);
-	            }
-	        }
+        // if global variables are defined by that name, try to call it.
+        // the 'call' convention comes from Closure
+        GlobalVariable v = GlobalVariable.byName(name, $buildNoException());
+        if (v != null) {
+            try {
+                Object o = v.getValue(this);
+                return InvokerHelper.getMetaClass(o).invokeMethod(o, "call", args);
+            } catch (Exception x) {
+                throw new InvokerInvocationException(x);
+            }
+        }
 		
-	        // otherwise try Step impls.
-	        DSL dsl = (DSL) getBinding().getVariable(STEPS_VAR);
-	        return dsl.invokeMethod(name,args);
+        // otherwise try Step impls.
+        DSL dsl = (DSL) getBinding().getVariable(STEPS_VAR);
+        return dsl.invokeMethod(name,args);
 	             
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
@@ -27,6 +27,7 @@ package org.jenkinsci.plugins.workflow.cps;
 import com.cloudbees.groovy.cps.SerializableScript;
 import groovy.lang.GroovyShell;
 import groovy.lang.MissingPropertyException;
+import groovy.lang.MissingMethodException;
 import groovy.lang.Script;
 import hudson.model.Queue;
 import hudson.model.Run;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
```diff
@@ -91,6 +91,15 @@ public CpsScript() throws IOException {
     @Override
     public final Object invokeMethod(String name, Object args) {
         // TODO probably better to call super method and only proceed here incase of MissingMethodException:
+        
+        // check for user defined closures in the script binding
+        if (getBinding().hasVariable(name)){
+            if (getBinding().getVariable(name) instanceof CpsClosure2){
+		CpsClosure2 local_closure = (CpsClosure2) getBinding().getVariable(name);
```
svanoort: Nitpick: why not store `getBinding().getVariable(name)` to a local variable so we don't repeatedly fetch?

steven-terrana:
@svanoort I appreciate the feedback - changes made and unit test added & passing to DSLTest.java.  Admittedly, the test should probably go in CpsScriptTest, but I was having difficult getting the jenkins harness to work appropriately.  When I added the test to DSLTest it performed as expected. 

To reproduce the bug without the potential fix introduced in this PR, the following Jenkinsfile would suffice:

~~~groovy
my_closure = { println "my closure!" }
my_closure()
~~~

resulting in java.lang.NoSuchMethodError: No such DSL method 'my_closure' found among steps

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -314,6 +315,18 @@ public void namedSoleParamForStep() throws Exception {
...
         r.assertBuildStatusSuccess(p.scheduleBuild2(0));
     }
+
+    /**
+     *  Test user defined closure execution
+     */
+    @Test public void userDefinedClosureExecution() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        ScriptApproval.get().approveSignature("method groovy.lang.Binding setVariable java.lang.String java.lang.Object");
+        p.setDefinition(new CpsFlowDefinition("binding.setVariable(\"my_closure\", { echo \"my closure!\" })\n my_closure() ", true));
```
svanoort: Nitpick: can just use "false" for the setDefinition argument 2 - that turns off sandbox for the test and you won't need the ScriptApproval.   Seems like a reasonable testcase though.

steven-terrana:
Thanks!  Please let me know if there's anything else I can do on this.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
```diff
@@ -91,6 +91,16 @@ public CpsScript() throws IOException {
     @Override
     public final Object invokeMethod(String name, Object args) {
         // TODO probably better to call super method and only proceed here incase of MissingMethodException:
```
jglick: The comment outlines what I suspect would be the cleaner fix.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScriptTest.java
```diff
@@ -113,4 +113,4 @@ private String dumpError() {
         assertEquals(dumpError(), Result.SUCCESS, exec.getResult());
     }
 
-}
+}
```
jglick: revert please

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -314,6 +314,17 @@ public void namedSoleParamForStep() throws Exception {
         p.setDefinition(new CpsFlowDefinition("try {def c = classLoad(getClass().name); error(/did not expect to be able to load ${c} from ${c.classLoader}/)} catch (ClassNotFoundException x) {echo(/good, got ${x}/)}", false));
         r.assertBuildStatusSuccess(p.scheduleBuild2(0));
     }
+
+    /**
+     *  Test user defined closure execution
+     */
+    @Test public void userDefinedClosureExecution() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("binding.setVariable(\"my_closure\", { echo \"my closure!\" })\n my_closure() ", false));
```
jglick: Two things that are weak about this test:

* It does not verify that `CpsClosure2.call(Object)` properly handles arguments, which is not at all obvious to me at least from code inspection—Groovy is weird this way. Ought to be testing closures defining zero, one, multiple, and varargs just to be sure.
* It does not verify that `CpsCallableInvocation` is being handled naturally. This would be most easily tested by, say,
  ```groovy
  binding['my_closure'] = {
    sleep 1 // exercises CPS interpreter
    echo 'ran'
  }
  my_closure()
  ```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java
jglick: (FWIW. Fine to keep kicking that can down the road, so long as we have test coverage for the stuff we think works.)

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
steven-terrana: Thanks for the feedback @jglick, i'll modify the test case.

steven-terrana:
@jglick @svanoort 

varags fails, though i'm hesitant to say it's because of the implementation of invoking the method. 

It seems to be a limitation of CpsClosure.  When using varargs, only the first instance of the variatic input parameter is passed regardless of whether it's in the script binding or defined as a local variable.

For instance, 
~~~ groovy
def my_closure = { String message, Integer... n ->
    println message
    n.each{ num ->
        println num
    }
}

my_closure.call("hey!", 2, 2, 3)
~~~

results in 
```
hey!
2
```

I'm currently trying to track down why this is the case, particularly looking at the CpsClosure definition [here](https://github.com/cloudbees/groovy-cps/blob/master/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsClosure.java)

steven-terrana:
bump.  Given that this PR works as expected, I feel as if this could be merged and a separate ticket regarding the varargs problem with CpsClosure's should be opened (which I would be happy to do, outlining the steps to reproduce outlined above). 

thoughts?

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -314,6 +314,70 @@ public void namedSoleParamForStep() throws Exception {
...
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        r.assertLogContains("my closure!", b); 
+    }
+
+    @Test public void userDefinedClosureShellInvocationExecution() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("my_closure = { \n" +
+                                              " sleep 1 \n" + 
+                                              " node{ \n" + 
+                                              "   sh \"pwd\" \n" +
```
svanoort: This will fail tests under Windows agents -- please see some of the other examples that use the Hudson utility function to determine if we're on Windows vs. Unix and run the appropriate step (sh or bad).

steven-terrana:
@svanoort - switched to pwd instead of sh to avoid OS specifics. primarily wanted to test the ability to resolve pipeline DSL methods within the user defined closure. 

The unit tests are failing but it seems to be a different thing each time? The previous run had tests fail because git.exe could not be found.  Some tests failed for other reasons this time. 

none of the tests that are failing are coming from DSLTest - which are the only tests updated. 

Any insights?

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -314,6 +315,91 @@ public void namedSoleParamForStep() throws Exception {
...
+    
+    /**
+    * Tests the ability to resolve DSL steps within a user defined closure
+    */
+    @Test public void userDefinedClosurePwdInvocationExecution() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("my_closure = { \n" +
+                                              " sleep 1 \n" + 
+                                              " node{ \n" + 
+                                              "   echo pwd() \n" +
```
svanoort: Uh, I'm not really sure the pwd() is needed/helpful there?  A simple echo 'some stuff' would probably suffice.

svanoort:
@steven-terrana Have you verified all tests pass locally with a clean build env?  I did see tests passing earlier and this could have had a subtle effect.

@abayer Any chance this might be related to the env issues introduced with durable tasks on windows, i.e. https://issues.jenkins-ci.org/browse/JENKINS-44334?

steven-terrana:
@svanoort all checks have passed! no changes on my part. 

anything left to be done for this?

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
steven-terrana: ah, just saw this.  I can modify it.  I though echo was defined in CpsScript and was trying to test that DSL steps resolve within user defined closures - but that's not the case. 

I can remove this test as it isn't testing anything additional.

steven-terrana:
@svanoort @jglick any progress on this? appreciate it!

steven-terrana:
I've updated this on my branch to be more generic.  Any object in the binding that responds to the call method with the arguments passed should be invoked. 

For some reason I'm not seeing the changes getting picked up by this PR?  I've pulled the latest changes since this PR was originally opened.  Tests still pass as expected. 

FWIW, I've been using this fix in production without issue for a month now.

svanoort:
IOU another round of review -- I went to review this before cutting a release and realized it needs a closer look than expected for the final changes

svanoort:
Out-of-band conversation with @jglick indicates he doesn't see a need to block this for his review so I'm going to go ahead and merge

svanoort:
Thanks for your contribution and your hard work on this @steven-terrana

steven-terrana:
Thank you!

svanoort:
Dang, I swear I'd merged it, oops, fixing that now.

PR #209, opened 2018-03-06
Switch Snippetizer sidepanel links to use an extension point

abayer:
This will allow Declarative to contribute its own links to the sidepanel as well.

Upstream of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/250

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
@@ -27,7 +27,10 @@ package org.jenkinsci.plugins.workflow.cps;
 import hudson.Extension;
 import hudson.ExtensionPoint;
 import hudson.Functions;
+import hudson.model.Job;
 import hudson.model.RootAction;
+import org.kohsuke.stapler.Stapler;
+import org.kohsuke.stapler.StaplerRequest;
 
 import javax.annotation.Nonnull;
 
@@ -54,23 +57,32 @@ public abstract class SnippetizerLink implements ExtensionPoint {
     public abstract String getUrl();
 
     /**
-     * Get the actual URL to use in sidepanel.jelly. If {@link #getUrl()} is not absolute, this will prepend "../" to
-     * {@link #getUrl()} to ensure it will work from the snippetizer and other similar pages. Otherwise it will just
-     * return {@link #getUrl()}.
+     * Get the actual URL to use in sidepanel.jelly. If {@link #getUrl()} is not absolute, this will try to get the
+     * current Job context and return a url starting with that job's {@link Job#getUrl()} appended with {@link #getUrl()}.
      */
     @Nonnull
     public final String getDisplayUrl() {
         String u = getUrl();
+
         try {
             if (new URI(u).isAbsolute()) {
                 return u;
-            } else {
-                return "../" + u;
             }
         } catch (URISyntaxException e) {
-            LOGGER.log(Level.WARNING, "Failed to parse URL for {0}: {1}", new Object[] {u, e});
+            LOGGER.log(Level.WARNING, "Failed to parse URL for {0}: {1}", new Object[]{u, e});
             return "";
         }
+
+        StaplerRequest req = Stapler.getCurrentRequest();
+        if (req == null) {
+            return u;
+        }
+
+        Job ancestor = req.findAncestorObject(Job.class);
+        if (ancestor == null) {
+            return u;
+        }
+        return "/" + ancestor.getUrl() + u;
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
@@ -26,9 +26,8 @@ package org.jenkinsci.plugins.workflow.cps;
 
 import hudson.Extension;
 import hudson.ExtensionPoint;
-import hudson.Functions;
+import hudson.model.Item;
 import hudson.model.Job;
-import hudson.model.RootAction;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
 
@@ -78,11 +77,13 @@ public abstract class SnippetizerLink implements ExtensionPoint {
             return u;
         }
 
-        Job ancestor = req.findAncestorObject(Job.class);
-        if (ancestor == null) {
+        Item i = req.findAncestorObject(Item.class);
+
+        if (i == null) {
             return u;
         }
-        return "/" + ancestor.getUrl() + u;
+
+        return req.getContextPath() + "/" + i.getUrl() + u;
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
@@ -1,0 +1,187 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,187 @@
...
+    private static final Logger LOGGER = Logger.getLogger(SnippetizerLink.class.getName());
+
+    /**
+     * Get the URL this link should point to.
+     */
+    @Nonnull
+    public abstract String getUrl();
+
+    /**
+     * Get the actual URL to use in sidepanel.jelly. If {@link #getUrl()} is not absolute, this will prepend "../" to
```
jglick: Why would you not just require `getUrl` to be absolute?

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Get the display name for the link.
+     */
+    @Nonnull
+    public abstract String getDisplayName();
+
+    /**
+     * Check whether the link should target a new window - this defaults to false;
```
jglick: I would think this would be the decision of the GUI embedding the link. If from a configuration screen, `target="_blank"`. If from a sidepanel, no.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: Good question. Will change to absolute.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: I was trying to ape the behavior currently present in the sidepanel, where the GDSL and external docs link always go to a new window.

abayer:
@jglick - Correct, **Pipeline Syntax** will still go to the Snippet Generator page, with links off to everything else on the sidepanel.

abayer:
@jglick @svanoort - would appreciate a review on this in the next couple days so I can merge/release and move ahead with the Declarative changes that depend on this. Thanks!

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,200 @@
...
+ * A link that will show up on the side panel of the snippet generator and other similar pages.
+ * Display order is determined by extension ordinal - highest ordinal first.
+ * 
+ * @author Andrew Bayer
+ */
+public abstract class SnippetizerLink implements ExtensionPoint {
+    private static final Logger LOGGER = Logger.getLogger(SnippetizerLink.class.getName());
+
+    /**
+     * Get the URL this link should point to.
```
jglick: Should document permitted formats, as in `getDisplayUrl`.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,200 @@
...
+    @Nonnull
+    public final String getDisplayUrl() {
+        String u = getUrl();
+
+        try {
+            if (new URI(u).isAbsolute()) {
+                return u;
+            }
+        } catch (URISyntaxException e) {
+            LOGGER.log(Level.WARNING, "Failed to parse URL for {0}: {1}", new Object[]{u, e});
```
jglick: or

```java
LOGGER.log(Level.WARNING, "Failed to parse URL for " + u, e);
```

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,200 @@
...
+            return "";
+        }
+
+        StaplerRequest req = Stapler.getCurrentRequest();
+        if (req == null) {
+            return u;
```
jglick: Hmm…

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        Item i = req.findAncestorObject(Item.class);
+
+        if (i == null) {
+            return u;
```
jglick: Should this not be

```java
return req.getContextPath() + "/" + u;
```

? Check behavior of `http://jenkins/pipeline-syntax` (yes this is a valid URL).

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,200 @@
...
+            return Messages.SnippetizerLink_GlobalsReferenceLink_displayName();
+        }
+    }
+
+    @Extension(ordinal = 700L)
+    public static class OnlineDocsLink extends SnippetizerLink {
+        @Override
+        @Nonnull
+        public String getUrl() {
+            return "https://jenkins.io/doc/pipeline/";
```
jglick: BTW we should be using a redirect for this.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
svanoort: With context path, check for a trailing '/' character before appending btw (have seen issues with relative URLs sometimes -- shouldn't be an issue but it happens).

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
svanoort: +1

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: Realistically, this won't ever happen, but I decided to play it a bit safe and not return empty here.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: Good catch.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: @jglick Not sure what you mean?

abayer:
@svanoort Tests added for making sure the links show up with the expected URL both when we're in a job and when we're not. I think that suffices, along with the manual testing I've done for folders, e.g.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: and fwiw, all I'm doing here is replicating the existing behavior of the rendered HTML, so if that should be changed, that should probably be handled separately from this.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,217 @@
...
+        if (!req.getContextPath().endsWith("/")) {
+            toAppend.append("/");
+        }
+
+        if (i == null) {
+            toAppend.append(u);
+        } else {
+            toAppend.append(i.getUrl());
+
+            if (!i.getUrl().endsWith("/")) {
```
jglick: This would be illegal. So, could all be simplified to something like

```java
return req.getContextPath() + '/' + (i == null ? '' : i.getUrl()) + u;
```

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
jglick: Yes it could be handled separately, merely noting it as an aside. By redirect I mean `https://jenkins.io/redirect/some-short-phrase-here/`, the target of which can be controlled by a simple PR to `jenkins-infra/jenkins.io`, so that you can rearrange web properties later without touching the product.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        Item i = req.findAncestorObject(Item.class);
+
+        StringBuilder toAppend = new StringBuilder();
+
+        toAppend.append(req.getContextPath());
+
+        if (!req.getContextPath().endsWith("/")) {
+            toAppend.append("/");
```
jglick: The context path is either blank or something like `/jenkins`. It will _never_ end in a slash.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
jglick: I think this could just be an `IllegalStateException`. Returning a relative URL here would never actually work anyway.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
abayer: Aaaah, good idea.

PR #210, opened 2018-03-06
Fix [JENKINS-49961] by avoiding an unsafe save operation when toggling ResumeEnabled if the FlowExecutionOwner is not yet set

svanoort:
Fixes [JENKINS-49961](https://issues.jenkins-ci.org/browse/JENKINS-49961) and includes test for same.

Save is not needed *here* when flag is mutated because it will be re-evaluated at onLoad.  Besides, the run will probably be saved momentarily due the results of resuming the build.

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.73.1'])
+buildPlugin(jenkinsVersions: [null, '2.73.3'])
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'])
+buildPlugin(jenkinsVersions: [null, '2.73.1'])
```
jglick: Better to just use 2.73.x as the baseline. No one cares about people using dailies prior to a published LTS line.

PR #211, opened 2018-03-09
[JENKINS-31314] Integrate and test a groovy-cps diagnostic trick

jglick:
[JENKINS-31314](https://issues.jenkins-ci.org/browse/JENKINS-31314). Downstream of https://github.com/cloudbees/groovy-cps/pull/85.

- [X] catching cases I could think of (incl. one inspired by a question from @quivers)
- [X] false positives on default Groovy methods
- [x] false positive on metasteps (e.g., `DSLTest.dollar_class_must_die3`)
- [x] false positive on global variables (e.g., `RestartingLoadStepTest.accessToSiblingScripts`, or more simply `DSLTest.userDefinedClosure0ArgsExecution`)
- [x] false positive on `evaluate` (e.g., `CpsFlowExecutionMemoryTest`)
- [x] false positives on stuff in `CpsDefaultGroovyMethodsTest` (`CpsBooleanClosureWrapper`?)
- [x] false positives on `get` in `CpsDefaultGroovyMethodsTest.withDefaultList`/`map`
- [x] look for false positives in `workflow-cps-global-lib` tests
- [x] look for false positives in `pipeline-model-definition` tests


abayer:
So the `CpsBooleanClosureWrapper` one actually doesn't surprise me, after my latest failed attempt at dealing with the `.sort {}` stuff a week or two ago: I honestly was surprised that the `CpsBooleanClosureWrapper` stuff even *works* in the first place. In fact, I'd bet that there are permutations you can create that wouldn't work (i.e., closures defined within the `map.any { k, v -> ... }` closure would probably short-circuit). So the fix there is probably for me to grind out an actual solution for `DefaultGroovyMethods` closure methods that perform/expect coercion into a `boolean`. Yay.

`withDefaultList`/`withDefaultMap` seems similar - `[].withDefaultList { i -> i * 2}.get(1)` ends up creating an instance of `groovy.lang.ListWithDefault` with the provided closure in its `initClosure` field. And then when we call `get(1)`, it ends up routing through `getDefaultValue(int idx)`, which returns `initClosure.call(new Object[]{idx})`. Again, a nested closure within the block would probably break this. So...again, a generalized solution is needed. Fun.

fwiw, the only `CpsDefaultGroovyMethods` that spit out an `expected to call ...` are the `withDefault...` methods and `Map` methods that need a boolean condition - so `any`, `count`, `every`, `find`, and `findAll`. So `BooleanClosureWrapper` interaction and `ListWithDefaults` interaction seem to be the only currently functioning false positives.

abayer:
...the fun in both those cases and the general incorrect behavior of `.sort {}` and friends is that they all end up having the actual `call` of the closure happening in another class, not in `DefaultGroovyMethods` - i.e., `BooleanClosureWrapper#call`, `ListWithDefaults#getDefaultValue`, which are the simple cases (i.e., in the naive case, they *work* but do so in such a way that it's actually cheating, hence the `expected to call ...` message), and...the `sort` hell. 

`sort` and friends are particularly hellish because their `call` of the closure is actually a couple levels deep: the closure ends up getting passed to an instance of a `Comparator` (either `groovy.util.OrderBy` or `groovy.util.ClosureComparator`), which themselves are passed to a `Collections.sort` call...so the actual `call` ends up happening when something deep in the Java sorting/comparing logic calls `compareTo` on the `OrderBy` or `ClosureComparator` instance. I got stuck on that because I really couldn't figure out what the right solution would be - honestly, it almost seemed like what we needed there was to have our own `CpsOrderBy` and `CpsClosureComparator` replacements that do the `call` in a new CPS thread or something.

jglick:
> the only currently functioning false positives

No, see `RestartingLoadStepTest.accessToSiblingScripts`. I suspect the same applies to any library `vars/x.groovy` with a `call` method (i.e., the most common use case for libraries): a source call `x(…)` will throw `CpsCallableInvocation` from a `call` method via `CpsClosure`, but I am not yet sure how we detect that. OTOH I was unable to reproduce a similar mismatch merely from

```groovy
def x = {-> 2 + 2}; println(x())
```

so I am not sure exactly where the problem lies or what form the solution would take.

jglick:
> I got stuck on [handling `sort`]

Well, if we can get this PR working, at least those cases would become more obvious to users and there might be fewer complaints and confusion.

abayer:
(note - looks like the `BooleanClosureWrapper` and `ListWithDefault` cases are more robust than I thought - or at least I haven't managed to break them with nested closures yet)

abayer:
Sorry, I meant the only current false positives in `CpsDefaultGroovyMethods`.

src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
```diff
@@ -228,23 +227,4 @@ public boolean start() throws Exception {
         });
     }
 
-    @Ignore("TODO JENKINS-31314: calls writeFile just once, echoes null (i.e., return value of writeFile), then succeeds")
-    @Test public void nonCpsContinuable() {
```
jglick: Introduced in https://github.com/jenkinsci/pipeline-plugin/pull/233 FTR.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -392,7 +393,7 @@ protected Object invokeDescribable(String symbol, Object _args) {
                 ud = new UninstantiatedDescribable(symbol, null, dargs);
                 margs.put(p.getName(),ud);
 
-                return invokeStep(metaStep,new NamedArgsAndClosure(margs,args.body));
+                return invokeStep(metaStep, symbol, new NamedArgsAndClosure(margs, args.body));
```
jglick: https://github.com/jenkinsci/workflow-cps-plugin/pull/239#discussion_r277671055

pom.xml
```diff
@@ -70,7 +70,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>3.2</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <groovy-cps.version>1.26</groovy-cps.version>
+        <groovy-cps.version>1.27-20190424.002713-11</groovy-cps.version> <!-- TODO https://github.com/cloudbees/groovy-cps/pull/85 -->
```
jglick: Would love to set up Incrementals upstream, but that would probably require getting it moved into @jenkinsci, and would require some help from infrastructure people.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -105,9 +107,42 @@ private ThreadContext setUp() {
...
+                LOGGER.log(Level.FINE, null, x);
+            }
+        }
+    }
+
+    private static @CheckForNull String className(@CheckForNull Object receiver) {
+        return receiver == null ? null : receiver.getClass().getName();
+    }
+
+    static boolean FAIL_ON_MISMATCH = Main.isUnitTest;
```
jglick: Making false positives be fatal makes it much easier to detect mistakes here and in PCT. But we would rather have this be nonfatal in production, since there are sure to be some false positives in exotic situations not yet covered by tests. (As well as some false negatives, but this is a best effort after all.)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -105,9 +107,42 @@ private ThreadContext setUp() {
...
+
+    static String mismatchMessage(@CheckForNull String expectedReceiverClassName, String expectedMethodName, @CheckForNull String actualReceiverClassName, String actualMethodName) {
+        // TODO reference something like https://jenkins.io/redirects/pipeline-cps-method-mismatches/ sending you to a wiki page with commonly attempted idioms and the working equivalents
```
jglick: Can easily do this if the main PR is accepted.

dwnusbaum:
`CpsThreadDumpTest.nativeMethods` failed in the CI build. I'm not sure if its related to the change or is just flaky:

> [2019-04-25T02:12:58.277Z] [ERROR]   Run 5: CpsThreadDumpTest.nativeMethods:132->assertStackTrace:140 expected:<[WorkflowScript.helper(WorkflowScript:2), WorkflowScript.run(WorkflowScript:3)]> but was:<[DSL.sleep(sleeping for another 0.97 sec), WorkflowScript.helper(WorkflowScript:2), WorkflowScript.run(WorkflowScript:3)]>

jglick:
I will check it.

jglick:
Seems like a flake: the CI builds actually passed on the original branch head here, only failing on this merge build; and I get the same error locally even on the revisions the CI build accepted! Will investigate.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -24,22 +24,97 @@
...
+                return null;
+            });
+            errors.checkSucceeds(() -> {
+                p.setDefinition(new CpsFlowDefinition("@NonCPS def bad() {polygon(17) {}}; bad()", true));
+                WorkflowRun b = r.buildAndAssertSuccess(p);
+                r.assertLogContains("wrapping in a 17-gon", b);
+                r.assertLogContains(CpsVmExecutorService.mismatchMessage("WorkflowScript", "bad", null, "polygon"), b);
+                return null;
+            });
+        } finally {
```
dwnusbaum: I was hoping that a warning would be printed when CPS-transformed code is used inside a constructor, for example in something like this:

```
errors.checkSucceeds(() -> {
    p.setDefinition(new CpsFlowDefinition("class Foo { Foo(script) { script.sleep(1) } }; new Foo(this)", true));
    WorkflowRun b = r.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
    r.assertLogContains(CpsVmExecutorService.mismatchMessage("Foo", "<init>", null, "sleep"), b);
    return null;
});
```

For now though, that code appears to just fail directly with `hudson.remoting.ProxyException: CpsCallableInvocation{methodName=sleep, call=com.cloudbees.groovy.cps.impl.CpsFunction@3d58963, receiver=null, arguments=[org.jenkinsci.plugins.workflow.cps.DSL$ThreadTaskImpl@19cb8536]}` and no additional message is printed.

I think this is because [FunctionCallBlock.dispatchOrArg](https://github.com/cloudbees/groovy-cps/blob/1d00d440688d6320f81c57c00e5320e63fa05727/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java#L96) handles constructor invocation directly so none of the code added downstream in `ContinuationGroup.methodCall` applies. Would it make sense to add similar logic [here](https://github.com/cloudbees/groovy-cps/blob/1d00d440688d6320f81c57c00e5320e63fa05727/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java#L98) to give more context to this kind of error?

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -24,22 +24,97 @@
...
+                WorkflowRun b = r.buildAndAssertSuccess(p);
+                r.assertLogContains("oops got -1", b);
+                r.assertLogContains(CpsVmExecutorService.mismatchMessage("java.util.ArrayList", "sort", "org.jenkinsci.plugins.workflow.cps.CpsClosure2", "call"), b);
+                return null;
+            });
+            errors.checkSucceeds(() -> {
+                p.setDefinition(new CpsFlowDefinition("node {[1, 2, 3].each {x -> sleep 1; echo(/no problem got $x/)}}", true));
+                WorkflowRun b = r.buildAndAssertSuccess(p);
+                r.assertLogContains("no problem got 3", b);
+                r.assertLogNotContains("expected to call", b);
```
dwnusbaum: When I test this PR locally against the latest version of the groovy-cps PR, this assertion fails, because the following line is present in the logs: `expected to call java.util.ArrayList.each but wound up catching ?`. Perhaps the build is passing here because the main snapshot is outdated? I am a little surprised that the expected receiver is `java.util.ArrayList.each` and not something like `$each__java_util_Iterable__groovy_lang_Closure`, since I thought this was one of the transformed methods from DefaultGroovyMethods.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -24,22 +24,97 @@
...
+                return null;
+            });
+            errors.checkSucceeds(() -> {
+                p.setDefinition(new CpsFlowDefinition("class C {@Override String toString() {'never used'}}; def gstring = /embedding ${new C()}/; echo(/oops got $gstring/)", true));
+                WorkflowRun b = r.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0)); // JENKINS-27306: No such constructor found: new org.codehaus.groovy.runtime.GStringImpl java.lang.String java.lang.String[]
+                r.assertLogContains(CpsVmExecutorService.mismatchMessage("java.lang.Class", "asType", "C", "toString"), b);
```
dwnusbaum: The expected receiver class name seems a bit weird here. If you patch `CpsVmExecutorService.className` to return the class name directly when the expected receiver is a class and not a method rather than calling `getClass` on the class then you get the following, which seems better:

> expected to call org.codehaus.groovy.runtime.ScriptBytecodeAdapter.asType but wound up catching C.toString

```
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
index d094278e..7b7691d5 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -125,6 +125,9 @@ class CpsVmExecutorService extends InterceptingExecutorService {
     }
 
     private static @CheckForNull String className(@CheckForNull Object receiver) {
+        if (receiver instanceof Class) {
+            return ((Class) receiver).getName();
+        }
         return receiver == null ? null : receiver.getClass().getName();
     }
```

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
dwnusbaum: This patch to `FunctionCallBlock` over in groovy-cps along with the patch to `CpsVmExecutorService.className` makes the test work, although I don't know if it might cause some false positives:

```
diff --git a/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java b/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
index 94f71e2..03e4c32 100644
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
@@ -94,6 +94,10 @@ public class FunctionCallBlock extends CallSiteBlockSupport {
                     Object v;
                     try {
                         v = e.getInvoker().contextualize(FunctionCallBlock.this).constructorCall((Class)lhs,args);
+                    } catch (CpsCallableInvocation inv) {
+                        inv.checkMismatch(lhs, Arrays.asList("<init>"));
+                        Throwable t = new IllegalStateException("Calling CPS-transformed methods inside a constructor is not supported.");
+                        return throwException(e, t, loc, new ReferenceStackTrace());
                     } catch (Throwable t) {
                         return throwException(e, t, loc, new ReferenceStackTrace());
                     }
```

jglick:
Last test failure was [INFRA-2075](https://issues.jenkins-ci.org/browse/INFRA-2075).

jglick:
> Is there a PR anywhere for the content for the redirect link?

https://github.com/jenkins-infra/jenkins.io/pull/2290 was already merged. Note that the wiki page is incomplete—I left some stubbed-out sections. Could try to flesh those out if you need me to.

dwnusbaum:
Oops, I should have actually looked at that PR. I gave the wiki page a quick read and it looks good to me, thanks! We can flesh out the final sections as needed before releasing the change.

PR #214, opened 2018-03-16
[FIXED JENKINS-50171] Don't use package names in loadedScripts keys.

abayer:
[JENKINS-50171](https://issues.jenkins-ci.org/browse/JENKINS-50171)

It just makes things go badly.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
@@ -110,7 +110,7 @@ class CpsGroovyShell extends GroovyShell {
     public Script parse(GroovyCodeSource codeSource) throws CompilationFailedException {
         Script s = doParse(codeSource);
         if (execution!=null)
-            execution.loadedScripts.put(s.getClass().getName(), codeSource.getScriptText());
+            execution.loadedScripts.put(s.getClass().getSimpleName(), codeSource.getScriptText());
         if (this.execution != null && !this.execution.getDurabilityHint().isPersistWithEveryStep()) {
             // Ensure we persist new scripts
             this.execution.saveOwner();
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
@@ -1,6 +1,8 @@
 package org.jenkinsci.plugins.workflow.cps.steps;
 
 import javax.inject.Inject;
+
+import hudson.FilePath;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
```diff
@@ -110,7 +110,7 @@ public void prepareScript(Script script) {
     public Script parse(GroovyCodeSource codeSource) throws CompilationFailedException {
         Script s = doParse(codeSource);
         if (execution!=null)
-            execution.loadedScripts.put(s.getClass().getName(), codeSource.getScriptText());
+            execution.loadedScripts.put(s.getClass().getSimpleName(), codeSource.getScriptText());
```
svanoort: Um, if the issue is the fact that the key is a String and is duplicated, won't keying by simple name make it worse?  As in, rather than key collisions on *all* components (package and class simple name) it'll happen if just the class's base name matches?   Or am I missing something here about the impact of the keying?  For reference: https://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
```diff
@@ -157,4 +159,55 @@ public void pauseInsideLoad() throws Exception {
         });
     }
 
+    @Issue("JENKINS-50172")
+    @Test public void loadAndUnnamedClassesInPackage() {
```
svanoort: This testcase is fine for just *package* collisions but should be extended to cover classname collisions too.

abayer:
These classes are just the `Script` classes - one per `load` step. The only thing I was really concerned about was collision after resume with more `load` calls, but the logic for next script name is persistent, so that’s not a problem (and is tested). `loadedScripts`, in practice, never contains anything but `ScriptX` classes from `load`, so I do think it’s ok. I had to think about it a bunch to be sure, admittedly. =)

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
abayer: Yeah, like in the JIRA, that isn’t actually a thing. =)

svanoort:
@abayer I do think it's necessary to have testing for Anononymous/Inner classes IIUC though because of the `getSimpleName()` - but have a bit of a complex about this logic, so maybe that's prevented already somewhere that you can point to?  Similarly class name collisions.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
```diff
@@ -157,4 +159,55 @@ public void pauseInsideLoad() throws Exception {
...
+                        "node('master') {\n" +
+                        "    config = load 'src/org/foo/devops/JenkinsEnvironment.groovy'\n" +
+                        "    util = load 'src/org/foo/devops/Utility.groovy'\n" +
+                        "    config.loadProdConfiguration()\n" +
+                        "}\n" +
+                        "util.isValueExist(\"\")\n" +
+                        "semaphore 'wait'\n" +
+                        "node('master') {\n" +
+                        "    util2 = load 'src/org/foo/devops/Utility.groovy'\n" +
+                        "    util = load 'src/org/foo/devops/Utility.groovy'\n" +
```
jglick: You do not seem to actually be calling any of these utilities to make sure they can run…?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
jglick: It is a user error to include a `package` declaration in a `load`ed script, so maybe we should just throw an explanatory error right here.

svanoort: Warning at least would make sense -- but that need not be part of this PR.

abayer:
@svanoort - added anonymous and inner class tests.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
jglick: Yes a warning would be a good idea. Could be in this PR but does not need to be.

abayer: Yeah, I'm thinking of doing that in another PR - I'm debating whether the text should be just a pure "don't do that" warning or something more like "We're gonna error on this eventually, so *really* don't do that" warning, and I want to ponder that for a bit. =)

PR #215, opened 2018-03-26
[JENKINS-50407] Trying to diagnose an NPE

jglick:
[JENKINS-50407](https://issues.jenkins-ci.org/browse/JENKINS-50407)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/SandboxContinuable.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/SandboxContinuable.java
@@ -31,15 +31,15 @@ class SandboxContinuable extends Continuable {
         try {
             CpsFlowExecution e = thread.group.getExecution();
             if (e == null) {
-                throw new IllegalStateException("no loaded execution");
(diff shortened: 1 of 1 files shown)


svanoort:
I'm also 75% sure that this is a symptom of the same issues since it crops up with the https://issues.jenkins-ci.org/browse/JENKINS-49686 and https://issues.jenkins-ci.org/browse/JENKINS-50199 issues (and the other related bugs) in some variants and the fix *should* cover this anyway.  

The other 25% is because the upstream JIRA doesn't have as much context so it could be part of the same cluster of issues but triggered differently.

jglick:
> I should directly integrate this into my in-progress mega-fix

Note that you can just

```bash
git checkout my-mega-fix
git pull https://github.com/jglick/workflow-cps-plugin.git diag-JENKINS-50407
```

to subsume this one, if you agree with it. GitHub will not get confused, whether or not this is merged early.

svanoort:
@jglick Yep, I was planning on a direct git merge.

PR #216, opened 2018-03-30
Fix a whole wad of Resume issues and add a bunch of diagnostic logging for FlowHead problems etc

svanoort:
The main fixes are here https://github.com/jenkinsci/workflow-job-plugin/pull/93 but this helps harden against the following and adds a bunch of diagnostic logging:

* [JENKINS-50199](https://issues.jenkins-ci.org/browse/JENKINS-50199)
* [JENKINS-46986](https://issues.jenkins-ci.org/browse/JENKINS-46986)
* [JENKINS-50407](https://issues.jenkins-ci.org/browse/JENKINS-50407)

It also subsumes: https://github.com/jenkinsci/workflow-cps-plugin/pull/215

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18-20180404.011344-6</version>
+            <version>2.18-20180404.152425-7</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,8 @@
         <workflow-support-plugin.version>2.17</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
         <groovy-cps.version>1.24</groovy-cps.version>
-        <jenkins-test-harness.version>2.37-20180405.183013-2</jenkins-test-harness.version>
+        <!-- Pending https://github.com/jenkinsci/jenkins-test-harness/pull/101 -->
+        <jenkins-test-harness.version>2.37-20180405.213807-3</jenkins-test-harness.version>
     </properties>
     <dependencies>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <workflow-support-plugin.version>2.17</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
         <groovy-cps.version>1.24</groovy-cps.version>
-        <jenkins-test-harness.version>2.33</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2.37-20180405.183013-2</jenkins-test-harness.version>
     </properties>
     <dependencies>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18-20180403.233137-5</version>
+            <version>2.18-20180404.011344-6</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18-20180406.172304-8</version>
+            <version>2.18</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -69,8 +69,7 @@
         <workflow-support-plugin.version>2.17</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
         <groovy-cps.version>1.24</groovy-cps.version>
-        <!-- Pending https://github.com/jenkinsci/jenkins-test-harness/pull/101 -->
-        <jenkins-test-harness.version>2.37-20180405.213807-3</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2.37</jenkins-test-harness.version>
     </properties>
     <dependencies>
         <dependency>
@@ -142,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18-20180404.152425-7</version>
+            <version>2.18-20180406.172304-8</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.17</version>
+            <version>2.18-20180329.215807-4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18-20180329.215807-4</version>
+            <version>2.18-20180403.233137-5</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1820,6 +1820,13 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         boolean persistOk = true;
         FlowNodeStorage storage = getStorage();
         if (storage != null) {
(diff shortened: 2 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -313,7 +316,7 @@
      * {@link FlowExecution} gets loaded into memory for the build records that have been completed,
      * and for those we don't want to load the program state, so that check should be efficient.
      */
-    private boolean done;
+     Boolean done; // Only non-private for unit test use.
```
abayer: Extraneous leading space after indent? =)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -598,14 +601,35 @@ int approximateNodeCount() {
         return iota.get();
     }
 
+    /** For diagnostic purposes only. */
```
abayer: So I personally appreciate it when diagnostics methods like this have javadoc or comments giving a quick explanation of what their diagnostics are meant to show - not a blocker by any means, just a casual request.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -598,14 +601,35 @@ int approximateNodeCount() {
...
             // something went catastrophically wrong and there's no live head. fake one
+            LOGGER.log(Level.WARNING, "Failed to load pipeline heads, so faking some up for execution " + this.toString());
             if (this.startNodes == null) {
                 this.startNodes = new Stack<BlockStartNode>();
             }
-            this.heads.clear();
+
+            if (this.heads != null && this.heads.size() > 0) {
+                if (LOGGER.isLoggable(Level.INFO)) {
+                    LOGGER.log(Level.INFO, "Resetting heads to rebuild the Pipeline structure, tossing existing heads: "+getHeadsAsString());
```
abayer: oooo, this is a nice addition.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -687,6 +713,7 @@ public boolean canResume() {
     }
 
     @Override
+    @SuppressFBWarnings(value = "RC_REF_COMPARISON_BAD_PRACTICE_BOOLEAN", justification = "We want to explicitly check for boolean not-null and true")
```
abayer: Is it wrong that I love that with `Boolean`, you effectively can treat it as a, I dunno, `troolean`? What'd be the term for a boolean but for three possible states... =)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -827,8 +857,14 @@ private void loadProgramFailed(final Throwable problem, SettableFuture<CpsThread
     /** Report a fatal error in the VM. */
     void croak(Throwable t) {
         setResult(Result.FAILURE);
+        done = true;
```
abayer: Why `Boolean.TRUE` above but `true` here? Might be handy to mention why you're using one vs the other in one case (probably the `Boolean.TRUE` case, since that's not something you see as often).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1523,6 +1566,9 @@ public void marshal(Object source, HierarchicalStreamWriter w, MarshallingContex
                 for (BlockStartNode st : e.startNodes) {
                     writeChild(w, context, "start", st.getId(), String.class);
                 }
+                if (e.done != null) {
+                    writeChild(w, context, "done", e.done, Boolean.class);
```
abayer: *glee* I can't remember off the top of my head why I care about having `done` persisted, but I know I do.

src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
```diff
@@ -114,29 +116,32 @@ void newStartNode(FlowStartNode n) throws IOException {
         execution.storage.storeNode(head, false);
     }
 
-    void setNewHead(FlowNode v) {
+    void setNewHead(@Nonnull FlowNode v) {
+        if (v == null) {
```
abayer: So my brain briefly melts at the combination of `@Nonnull` and `if (v == null)` - I'm assuming this is defense-in-depth, with the `@Nonnull` covering compile-time scenarios and the `if (v == null)` covering even edgier-case scenarios?

src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
```diff
@@ -114,29 +116,32 @@ void newStartNode(FlowStartNode n) throws IOException {
...
-                // in recovering from error and such situation, we sometimes need to grow the graph
-                // without running the program.
-                // TODO can CpsThreadGroup.notifyNewHead be used instead to notify both kinds of listeners?
-                execution.notifyListeners(Collections.singletonList(v), true);
-                execution.notifyListeners(Collections.singletonList(v), false);
-            }
         } catch (IOException e) {
             LOGGER.log(Level.FINE, "Failed to record new head: " + v, e);
         }
+        v.addAction(new TimingAction());
```
abayer: So the idea here is that we want to still do the `TimingAction`, `notifyNewHead`, etc, even if we can't store the node?

src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
svanoort: Yeah, it's protecting against some things that *shouldn't* happen but yet were still happening in the wild, even though Findbugs didn't find issues at compile-time (probably it wasn't digging deeply enough).

abayer: Yeah, I have no problem with that approach - probably worth adding a comment explaining that, though. It'd be awfully easy for a later maintainer (i.e., me, because I'm forgetful) to not know/remember that and just think "Oh, that's a pointless null check, lemme clear the dead code".

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: A trilean, acc. to Wikipedia.

pom.xml
```diff
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18-20180329.215807-4</version>
+            <version>2.18-20180406.172304-8</version>
```
jglick: please link to upstream PRs

pom.xml
```diff
@@ -69,7 +69,7 @@
         <workflow-support-plugin.version>2.17</workflow-support-plugin.version>
         <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
         <groovy-cps.version>1.24</groovy-cps.version>
-        <jenkins-test-harness.version>2.33</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2.37</jenkins-test-harness.version>
```
jglick: file a request to update the parent POM

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1207,18 +1205,24 @@ synchronized void onProgramEnd(Outcome outcome) {
...
+
+                String tempIotaStr = Integer.toString(this.iota.get());
+                FlowHead lastHead = heads.get(first.getId());
+                if (lastHead == null || lastHead.get() == null || !(lastHead.get().getId().equals(tempIotaStr))) {
+                    // Warning of problems with the final call to FlowHead.setNewHead
+                    LOGGER.log(Level.WARNING, "Invalid final head for execution "+this.owner+" with head: "+lastHead);
+                }
             }
+        } catch (Exception ex) {
+            done = Boolean.TRUE;
```
jglick: `true`

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -534,9 +534,12 @@ synchronized boolean switchToAsyncMode() {
                 @Override public void onSuccess(CpsThreadGroup result) {
                     try {
                         // TODO keep track of whether the program was saved anyway after saveState was called but before now, and do not bother resaving it in that case
```
jglick: BTW how does durability interact with `saveState`? Important.

svanoort: @jglick `CpsStepContext#saveState()` follows the same rules as overall Program durability - if your `DurabilityHint` has `isPersistWithEveryStep()` (i.e. `MAX_SURVIVABILITY` or `SURVIVABLE_NONATOMIC`) it saves normally.  

Otherwise (for `PERFORMANCE_OPTIMIZED`) it doesn't -- your program is not providing guaranteed persistence in the face of catastrophic system failures.  

Of course with a clean shutdown the pipeline will persist via the `@Terminator` as part of shutdown, so that's fine.

pom.xml
svanoort: @jglick  https://github.com/jenkinsci/plugin-pom/pull/101 - Done

PR #217, opened 2018-04-02
Simplifying snippetizer link URL construction

abayer:
Minor cleanup after comments on #209

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
@@ -75,32 +75,12 @@ public abstract class SnippetizerLink implements ExtensionPoint {
 
         StaplerRequest req = Stapler.getCurrentRequest();
         if (req == null) {
-            return u;
+            throw new IllegalStateException("Can't get display URL without Stapler context.");
         }
 
         Item i = req.findAncestorObject(Item.class);
 
-        StringBuilder toAppend = new StringBuilder();
-
-        toAppend.append(req.getContextPath());
-
-        if (!req.getContextPath().endsWith("/")) {
-            toAppend.append("/");
-        }
-
-        if (i == null) {
-            toAppend.append(u);
-        } else {
-            toAppend.append(i.getUrl());
-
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -75,32 +75,12 @@ public final String getDisplayUrl() {
...
-
-            if (!i.getUrl().endsWith("/")) {
-                toAppend.append("/");
-            }
-
-            toAppend.append(u);
-        }
-
-        return toAppend.toString();
+        return req.getContextPath() + '/' + (i == null ? "" : i.getUrl()) + u;
```
jglick: I assume you manually tested both cases here?

svanoort: Wait, why did we remove the check for if the context path has a '/' on it?  We had an actual stage view bug that resulted from a similar issue.

src/main/java/org/jenkinsci/plugins/workflow/cps/SnippetizerLink.java
```diff
@@ -75,32 +76,12 @@ public final String getDisplayUrl() {
...
-
-            if (!i.getUrl().endsWith("/")) {
-                toAppend.append("/");
-            }
-
-            toAppend.append(u);
-        }
-
-        return toAppend.toString();
+        return Paths.get( req.getContextPath(), i == null ? "" : i.getUrl(), u).toString();
```
bitwiseman: Using `Paths` to concatenate class to handle slashes - will remove or add as needed.

dwnusbaum: I am little wary of using `Path` for URLs, would it not result in file system-specific behavior? Maybe fine, I just haven't seen it used in this kind of context.

bitwiseman: And you would be correct. 😄

PR #218, opened 2018-04-09
[JENKINS-50665] - nonCPS.exec() wrapper + Support of argument parameter filtering

oleg-nenashev:
Sorry for integrating several changes in a single PR, I am just testing a fix for https://github.com/jenkinsci/pipeline-utility-steps-plugin/pull/46 in Pipeline Utility Steps. It will act as a downstream PR

- [x] - Run tests with 2.107.1
- [x] - [JENKINS-50665](https://issues.jenkins-ci.org/browse/JENKINS-50665) - introduce new `ArgumentsActionFilteringStepDescriptor ` interface, which allow tweaking persisted parameters if needed
- [x] - [JENKINS-50670](https://issues.jenkins-ci.org/browse/JENKINS-50670) - demonstrate the JEP-200 error propagation mistake I have noticed during the implementation

@reviewbybees @rsandell @abayer @svanoort @jglick


src/main/resources/org/jenkinsci/plugins/workflow/cps/steps/NonCPSGlobalVariable/NonCPSWrapperImpl.groovy
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.jenkinsci.plugins.workflow.cps.steps.NonCPSGlobalVariable
+
+import com.cloudbees.groovy.cps.NonCPS
+
+class NonCPSWrapperImpl implements Serializable {
+
+    @NonCPS
+    <V> V exec(Closure<V> body) {
+        body()
```
jglick: Does this actually work? I cannot imagine how it could. `@NonCPS` is applied _lexically_ to compilation of this method source. It has no effect on the CPS transform of the body closure. I think you only _think_ it works because your `printTimestamp` method returns `true` from `StepExecution.start` and thus is synchronous (would be more simply written using `SynchronousStepExecution`, and thus the CPS transformation status of the body is largely irrelevant. If you add, say, a `sleep 1` statement inside the closure I expect you will see it blow up, or rather (pending #211) quietly behave strangely.

oleg-nenashev: Will check it. in https://github.com/jenkinsci/pipeline-utility-steps-plugin/pull/46 a complex wrapper with multiple steps does not fail after such patch, but I agree it may be also not a case in real.

svanoort: Yeah, I have really mixed feelings on this. 

As a whole, the NonCPS mechanism is NOT good about advertising when it has been abused and failing appropriately -- I think @jglick might have some in-progress work that resolves that though?

oleg-nenashev:
@jglick I have removed NonCPS wrapper from the scope for now

jglick:
Does #220 not handle this?

svanoort:
@jglick Parts yes, parts no -- I think the ability @oleg-nenashev proposes to let steps customize filtering of what they persist is actually quite interesting and potentially useful completely aside from the original issue.  Maybe not *quite* the right implementation for general use, but I want to mull a bit to think about where this might be applied.

oleg-nenashev:
Anyway, I will rework https://github.com/jenkinsci/pipeline-utility-steps-plugin/pull/46 to NOT depend on this PR

jglick:
I think this has been abandoned?

dwnusbaum:
@oleg-nenashev I'm going to close this PR. If you want to keep working on it feel free to resolve the conflicts and reopen it.

PR #220, opened 2018-04-11
Bulletproof against [JENKINS-50752] by catching unserializable step arguments

svanoort:
... before they can cause catastrophic issues.  

Solution for [JENKINS-50752](https://issues.jenkins-ci.org/browse/JENKINS-50752) which is causing problems in the wild. 

Upstream is: https://github.com/jenkinsci/workflow-api-plugin/pull/68

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,8 +80,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <!-- Temp version until upstream PR released -->
-            <version>2.27-20180411.200653-14</version>
+            <version>2.27</version>
         </dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -80,7 +80,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.25</version>
+            <!-- Temp version until upstream PR released -->
```
svanoort: https://github.com/jenkinsci/workflow-api-plugin/pull/68

abayer: In a perfect world, I prefer a link to the PR in the POM, but obviously that's not the end of the world. =)

PR #222, opened 2018-04-19
Lazy load fix to get or null [JENKINS-50784]

svanoort:
Fixes [JENKINS-50784](https://issues.jenkins-ci.org/browse/JENKINS-50784) in conjunction with https://github.com/jenkinsci/workflow-job-plugin/pull/95.

See comments on the JIRA for explanation in detail - note that this is more or less what @jglick suggested.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18</version>
+            <!-- Temp SNAPSHOT from https://github.com/jenkinsci/workflow-job-plugin/pull/95 -->
+            <version>2.20-20180419.201134-1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -141,8 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <!-- Temp SNAPSHOT from https://github.com/jenkinsci/workflow-job-plugin/pull/95 -->
-            <version>2.20-20180419.201134-1</version>
+            <version>2.20</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
@@ -106,13 +106,29 @@ public class ReplayAction implements Action {
         return isEnabled() || isRebuildEnabled() ? "replay" : null;
     }
 
-    private @CheckForNull CpsFlowExecution getExecution() {
+    /** Poke for an execution without blocking - may be null if run is very fresh or has not lazy-loaded yet. */
+    private @CheckForNull CpsFlowExecution getExecutionLazy() {
         FlowExecutionOwner owner = ((FlowExecutionOwner.Executable) run).asFlowExecutionOwner();
         if (owner == null) {
             return null;
         }
         FlowExecution exec = owner.getOrNull();
         return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+    }
+
+    /** Fetches execution, blocking if needed while we wait for some of the loading process. */
+    private @CheckForNull CpsFlowExecution getExecutionBlocking() {
+        FlowExecutionOwner owner = ((FlowExecutionOwner.Executable) run).asFlowExecutionOwner();
+        if (owner == null) {
+            return null;
+        }
+        try {
+            FlowExecution exec = owner.get();
+            return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+        } catch (IOException ioe) {
+            LOGGER.log(Level.WARNING, "Error fetching execution for replay", ioe);
+        }
+        return null;
     }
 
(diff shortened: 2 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -135,27 +151,38 @@ private ReplayAction(Run run) {
             return false;
         }
 
-        CpsFlowExecution exec = getExecution();
-        if (exec == null) {
-            return false;
+        CpsFlowExecution exec = getExecutionLazy();
+        if (exec != null) {
+            return exec.isSandbox() || Jenkins.getActiveInstance().hasPermission(Jenkins.ADMINISTER); // We have to check for ADMIN because un-sandboxed code can execute arbitrary on-master code
```
rsandell: Don't you mean `Jenkins.RUN_SCRIPTS`?

src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -135,27 +151,38 @@ private ReplayAction(Run run) {
...
-        if (exec.isSandbox()) {
+    }
+
+    /** Runs the extra tests for replayability beyond {@link #isEnabled()} that require a blocking load of the execution. */
+    /* accessible to Jelly */ public boolean isReplayableSandboxTest() {
+        CpsFlowExecution exec = getExecutionBlocking();
+        if (exec != null) {
+            if (!exec.isSandbox()) {
+                // We have to check for ADMIN because un-sandboxed code can execute arbitrary on-master code
+                return Jenkins.getActiveInstance().hasPermission(Jenkins.ADMINISTER);
```
rsandell: `Jenkins.RUN_SCRIPTS`?

svanoort: You're right circa https://github.com/jenkinsci/scriptler-plugin/pull/32/files - fixed

svanoort:
Build failure is CI going down (sigh).

src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -200,7 +201,7 @@ public static void assertPermissionId() {
                 List<Permission> permissions = Run.PERMISSIONS.getPermissions();
                 assertThat(permissions, Matchers.hasItem(ReplayAction.REPLAY));
                 gmas.add(ReplayAction.REPLAY, "dev2");
-                gmas.add(Jenkins.READ, "dev3");
+                gmas.add(Jenkins.Rit gEAD, "dev3");
```
rsandell: Did the cat step on your keyboard? :)

svanoort: @rsandell Um.... 🐱

rsandell: Interesting that he can almost spell _git_ correctly :smirk_cat:

pom.xml
```diff
@@ -141,7 +141,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.18</version>
+            <!-- Temp SNAPSHOT from https://github.com/jenkinsci/workflow-job-plugin/pull/95 -->
+            <version>2.20-20180419.201134-1</version>
```
jglick: But it will also work without this update, right?

src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -196,6 +240,15 @@ private static boolean canReplay(WorkflowRun b, String user) {
         });
     }
 
+    private static boolean canReplayDeepTest(WorkflowRun b, String user) {
+        final ReplayAction a = b.getAction(ReplayAction.class);
+        return ACL.impersonate(User.get(user).impersonate(), new NotReallyRoleSensitiveCallable<Boolean,RuntimeException>() {
```
jglick: use `ACL.as`

pom.xml
svanoort: No, `getOrNull` is semi-hosed without this update.

jglick: IIUC with the released `workflow-job` the **Replay** link would be unconditionally available on a build whose execution was not yet loaded, and then when you click on it, you either get the replay form, or you get a rebuild form if you for example lacked `RUN_SCRIPTS`. Right? Or was there an NPE interfering?

src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayPipelineCommand.java
```diff
@@ -67,7 +67,7 @@
             // which currently has no protected method allowing getItemByFullName to be replaced.
             throw new AbortException("Not a Pipeline build");
         }
-        if (!action.isEnabled()) {
+        if (!action.isEnabled() || !action.isReplayableSandboxTest()) {
```
jglick: BTW I would suggest marking `isEnabled` restricted just to be sure no unknown party is innocently calling it and thinking it secure.

pom.xml
svanoort: Unconditionally available but then you'd never actually be able to do the replay because you'd always have null, so... oops?

PR #224, opened 2018-05-03
Fix and test for JENKINS-50888 by supporting serialization of execution before onLoad was called

svanoort:
[JENKINS-50888](https://issues.jenkins-ci.org/browse/JENKINS-50888)

We missed the case of someone immediately modifying the build and saving before we onLoad could get invoked.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1611,7 +1611,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             writeChild(w, context, "iota", e.iota.get(), Integer.class);
             synchronized (e) {
                 if (e.headsSerial != null && (e.heads == null || e.heads.isEmpty())) { // Persisting before onLoad has called to load up the real heads
-                    // Note: heads should not be empty
+                    // Note: unlike start nodes, heads should not be empty once onLoad loading has happened
                     for (Entry<Integer, String> entry : e.headsSerial.entrySet()) {
                         writeChild(w, context, "head", entry.getKey() + ":" + entry.getValue(), String.class);
                     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1610,11 +1610,24 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             }
             writeChild(w, context, "iota", e.iota.get(), Integer.class);
             synchronized (e) {
-                for (FlowHead h : e.heads.values()) {
-                    writeChild(w, context, "head", h.getId() + ":" + h.get().getId(), String.class);
+                if (e.headsSerial != null && (e.heads == null || e.heads.isEmpty())) { // Persisting before onLoad has called to load up the real heads
+                    for (Entry<Integer, String> entry : e.headsSerial.entrySet()) {
+                        writeChild(w, context, "head", entry.getKey() + ":" + entry.getValue(), String.class);
+                    }
+                } else {  // onLoad was invoked, we have real heads*/
+                    for (FlowHead h : e.heads.values()) {
+                        writeChild(w, context, "head", h.getId() + ":" + h.get().getId(), String.class);
+                    }
                 }
-                for (BlockStartNode st : e.startNodes) {
-                    writeChild(w, context, "start", st.getId(), String.class);
+
+                if (e.startNodesSerial != null && (e.startNodes == null)) {  // Start nodes have not been lazy-loaded yet
+                    for (String startId : e.startNodesSerial) {
+                        writeChild(w, context, "start", startId, String.class);
+                    }
+                } else {  // Start nodes were loaded, so we can convert back to serial form
+                    for (BlockStartNode st : e.startNodes) {
+                        writeChild(w, context, "start", st.getId(), String.class);
+                    }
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -337,4 +338,25 @@ public void inProgressButStartBlocksLost() throws Exception {
             assertCompletedCleanly(run);
         });
     }
+
+    @Issue("JENKINS-50888")  // Tried to modify build without lazy load being triggered
+    @Test public void modifyBeforeLazyLoad() {
+        story.thenWithHardShutdown(r -> {  // Normal build
```
jglick: Why would the hard shutdown matter if the build already completed?

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -337,4 +338,25 @@ public void inProgressButStartBlocksLost() throws Exception {
...
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("echo 'dosomething'", true));
+            r.buildAndAssertSuccess(p);
+        });
+        story.then(r -> {  // But wait, we try to modify the build without loading the execution
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            b.setDescription("Bob");
+            b.save();  // Will trigger an IOException potentially
```
jglick: You mean, before the fix it might have thrown `IOException`? Comment is confusing in this context. If you are just reading the code, you would wonder—wait, if this throws an exception, would the test not fail now?

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -337,4 +338,25 @@ public void inProgressButStartBlocksLost() throws Exception {
...
+        });
+        story.then( r-> {  // Verify that the FlowExecutionOwner can trigger lazy-load correctly
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            Assert.assertEquals("Bob", b.getDescription());
+            Assert.assertEquals("4", b.getExecution().getCurrentHeads().get(0).getId());
```
jglick: Asserting a particular flow node ID is very fragile. Please change the assertion to verify something intrinsically meaningful. Say, that it is a `FlowEndNode`, or that its parent is an `echo` step, or something.

(Did I not suggest in some old PR that there be some kind of test utility to assert the nodes in a build? Something in `workflow-api-*-tests.jar` would be ideal.)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1610,11 +1610,24 @@ public void marshal(Object source, HierarchicalStreamWriter w, MarshallingContex
             }
             writeChild(w, context, "iota", e.iota.get(), Integer.class);
             synchronized (e) {
-                for (FlowHead h : e.heads.values()) {
-                    writeChild(w, context, "head", h.getId() + ":" + h.get().getId(), String.class);
+                if (e.headsSerial != null && (e.heads == null || e.heads.isEmpty())) { // Persisting before onLoad has called to load up the real heads
+                    for (Entry<Integer, String> entry : e.headsSerial.entrySet()) {
+                        writeChild(w, context, "head", entry.getKey() + ":" + entry.getValue(), String.class);
+                    }
+                } else {  // onLoad was invoked, we have real heads*/
```
jglick: stray `*/`

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1610,11 +1610,24 @@ public void marshal(Object source, HierarchicalStreamWriter w, MarshallingContex
...
+                    for (FlowHead h : e.heads.values()) {
+                        writeChild(w, context, "head", h.getId() + ":" + h.get().getId(), String.class);
+                    }
                 }
-                for (BlockStartNode st : e.startNodes) {
-                    writeChild(w, context, "start", st.getId(), String.class);
+
+                if (e.startNodesSerial != null && (e.startNodes == null)) {  // Start nodes have not been lazy-loaded yet
```
jglick: Is there a reason the clause for `heads` checked `isEmpty` but the clause for `startNodes` does not?

svanoort: Heads should never be empty, start nodes can be if build is completed.

svanoort: (added a comment to explain)

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
svanoort: @jglick So... it's only fragile if we change how iotas are done, which at this point would break a lot of things (there was a time when we could have done that without much impact, but it was years ago).    But some of those sorts of structural tests can be done by combining filter Predicates with the FlowScanners pretty trivially. 

For a case like this though, we really *do* need to assert that the ID is correct because that ensures no sort of placeholder FlowNodes have been inserted.

PR #225, opened 2018-05-04
JENKINS-51123: Fix deadlock in CpsFlowExecution

laurentgo:
A deadlock can arise if multiple threads tries to access storage
simultaneously while one of them doing it from a synchronized
method. The cause is that timing operation also tries to get the
instance lock from inside the storage r/w lock.

Solution is to change the order of locking by having the timing
object wrapping the call to the r/w storage lock instead of the
opposite.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -477,7 +477,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     public FlowNodeStorage getStorage() {
         return storage;
     }
-    
+
     public String getScript() {
         return script;
     }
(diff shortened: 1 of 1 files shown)


svanoort:
@laurentgo Never mind, I can verify this is correct by inspecting the lock order, thanks for the fast fix and I am going to go ahead and merge+release.

florent-vial:
@laurentgo @svanoort i still think we need a test case here, i still have a deadlock issue with 2.52.

kutzi:
I'm got a deadlock with 2.52 as well. See my comment in JENKINS-51132 https://issues.jenkins-ci.org/browse/JENKINS-51132?focusedCommentId=337105&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-337105

svanoort:
@florent-vial @kutzi Please use the JIRA for comments like this, not Github PRs - the remaining deadlock was solved in 2.53.

kutzi:
I've commented it in JIRA as well - as I wrote in the comment here.

PR #226, opened 2018-05-07
Fix more deadlocks

svanoort:
See [deadlock file](https://issues.jenkins-ci.org/secure/attachment/42497/deadlock.txt) from [JENKINS-51132 comment](https://issues.jenkins-ci.org/browse/JENKINS-51132?focusedCommentId=337105&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-337105). 

This prevents deadlocks by ensuring we do not lock Execution before Run (usually by separating the synchronization scopes). 

Lock order should always be Run -> Execution when locks are nested, or locks should be separated (the approach generally used here).

Why aren't there testcases?   I have low confidence in our ability produce a testcase that replicates all these deadlocks.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -633,36 +633,32 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 }
             }
 
-            try {
-                programPromise = Futures.immediateFailedFuture(new IllegalStateException("Failed loading heads", failureReason));
-                LOGGER.log(Level.INFO, "Creating placeholder flownodes for execution: " + this);
-                if (this.owner != null) {
-                    try {
-                        owner.getListener().getLogger().println("Creating placeholder flownodes because failed loading originals.");
-                    } catch (Exception ex) {
-                        // It's okay to fail to log
-                    }
+            programPromise = Futures.immediateFailedFuture(new IllegalStateException("Failed loading heads", failureReason));
+            LOGGER.log(Level.INFO, "Creating placeholder flownodes for execution: " + this);
+            if (this.owner != null) {
+                try {
+                    owner.getListener().getLogger().println("Creating placeholder flownodes because failed loading originals.");
+                } catch (Exception ex) {
+                    // It's okay to fail to log
                 }
-
-                // Switch to fallback storage so we don't delete original node data
-                this.storageDir = (this.storageDir != null) ? this.storageDir + "-fallback" : "workflow-fallback";
-                this.storage = createStorage();  // Empty storage
-
-                // Clear out old start nodes and heads
-                this.startNodes = new Stack<BlockStartNode>();
-                FlowHead head = new FlowHead(this);
-                this.heads = new TreeMap<Integer, FlowHead>();
-                heads.put(head.getId(), head);
-                FlowStartNode start = new FlowStartNode(this, iotaStr());
-                head.newStartNode(start);
-
-                // Create end
-                FlowNode end = new FlowEndNode(this, iotaStr(), (FlowStartNode) startNodes.pop(), result, getCurrentHeads().toArray(new FlowNode[0]));
-                end.addAction(new ErrorAction(failureReason));
-                head.setNewHead(end);
-            } catch (Exception ex) {
-                throw ex;
             }
+
+            // Switch to fallback storage so we don't delete original node data
+            this.storageDir = (this.storageDir != null) ? this.storageDir + "-fallback" : "workflow-fallback";
+            this.storage = createStorage();  // Empty storage
+
+            // Clear out old start nodes and heads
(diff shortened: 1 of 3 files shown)


svanoort:
@abayer Yeah, we had methods in CpsFlowExecution that synchronize on first the `FlowExecution`, then call `Run#save()`, which synchronizes  on the `Run`   --- but in general the `WorkflowRun` will synchronize on itself and save itself (synchronizing on the Execution to lock it during save).   Ergo deadlockdom.

jglick:
> I have low confidence in our ability produce a testcase that replicates all these deadlocks.

In general it is pretty fruitless to write automated tests for deadlocks. Possible, just not productive. You are not demonstrating that the code is free of deadlocks, only that the behavior of a particular sequence of events has changed for the better.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -620,47 +620,52 @@ private synchronized String getHeadsAsString() {
      * Bypasses {@link #croak(Throwable)} and {@link #onProgramEnd(Outcome)} to guarantee a clean path.
      */
     @GuardedBy("this")
-    synchronized void createPlaceholderNodes(Throwable failureReason) throws Exception {
-        this.done = true;
```
jglick: Unreadable diff. [better](https://github.com/jenkinsci/workflow-cps-plugin/pull/226/files?w=1)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -620,47 +620,52 @@ private synchronized String getHeadsAsString() {
...
+                heads.put(head.getId(), head);
+                FlowStartNode start = new FlowStartNode(this, iotaStr());
+                head.newStartNode(start);
+
+                // Create end
+                FlowNode end = new FlowEndNode(this, iotaStr(), (FlowStartNode) startNodes.pop(), result, getCurrentHeads().toArray(new FlowNode[0]));
+                end.addAction(new ErrorAction(failureReason));
+                head.setNewHead(end);
+            } catch (Exception ex) {
+                throw ex;
```
jglick: Uh, what is this `try`-`catch`-`throw` accomplishing? Just delete it!

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -620,47 +620,52 @@ private synchronized String getHeadsAsString() {
...
+            }
+        }
+        try {
+            saveOwner();
         } catch (Exception ex) {
             throw ex;
```
jglick: ditto, makes little sense

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -353,6 +353,11 @@ private boolean run() {
                     threads.remove(t.id);
                     if (threads.isEmpty()) {
                         execution.onProgramEnd(o);
+                        try {
+                            this.execution.saveOwner();
+                        } catch (Exception ex) {
+                            LOGGER.log(Level.WARNING, "Error saving execution for "+this, ex);
```
jglick: Useless `toString()`. Try

```java
LOGGER.log(Level.WARNING, "Error saving execution for " + execution, ex);
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
```diff
@@ -18,6 +18,7 @@
 import org.codehaus.groovy.control.CompilationUnit;
 import org.codehaus.groovy.control.CompilerConfiguration;
 import org.codehaus.groovy.control.SourceUnit;
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
```
jglick: unused import

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1217,7 +1226,7 @@ public boolean isComplete() {
     }
 
     /**
-     * Record the end of the build.
+     * Record the end of the build.  Note: we should always follow this with a call to saveOwner to persist the result.
```
jglick: `@link`

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
svanoort: Main reason is to separate out the synchronization components, but also try to ensure the run save happens.

svanoort: Wait, no.  I'm going to blame jet lag and lack of sleep for this. You're right, nonsensical.

PR #228, opened 2018-05-17
Record internal calls to what look like Jenkins APIs made during the course of Pipeline builds

jglick:
Mainly useful for non-sandboxed scripts, though for now it captures events whether running in the sandbox or not. (And will report method calls even if they are `@Whitelisted`.)

Only lightly tested, and will miss anything happening inside a `@NonCPS` block, but might be a good way to get a rough sense of how crazy an installation’s trusted libraries are.

~Given a 2.150.x+ dep,~ we could also create a `Telemetry` extension to see what everyone is using.


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1603,6 +1618,9 @@ public void marshal(Object source, HierarchicalStreamWriter w, MarshallingContex
                 if (e.timings != null) {
                     writeChild(w, context, "timings", e.timings, Map.class);
                 }
+                if (e.internalCalls != null) {
+                    writeChild(w, context, "internalCalls", e.internalCalls, Set.class);
```
svanoort: Potentially clever solution to avoid use of synchronization: de/serialize to/from a trivial Set implementation, but use a thread-safe representation (such as ConcurrentHashMap-backed impls) in-memory.

svanoort:
>  I would have time to develop it further, but wanted to demonstrate that this is possible and maybe “someone” can pick it up and productize it.

@jglick That's pretty much the definition of a good WIP PR.   I doubt anybody is likely to pick this up and carry it through to a merge-ready feature unless you can at least show it achieves its main goal of capturing the Jenkins-internal API calls (vs. the much rare in-sandbox uses).  c.f. https://github.com/jenkinsci/workflow-cps-plugin/pull/228#discussion_r190425311

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -171,7 +171,7 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
 
         FunctionCallEnv caller = new FunctionCallEnv(null, onSuccess, null, null);
         if (currentThread.getExecution().isSandbox())
-            caller.setInvoker(new SandboxInvoker());
+            caller.setInvoker(new LoggingInvoker(new SandboxInvoker()));
```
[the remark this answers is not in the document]
jglick: Not sure offhand if this code path is used much, but yeah this should probably be amended to say

```java
} else {
    caller.setInvoker(new LoggingInvoker(new DefaultInvoker()));
}
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -455,6 +459,17 @@ synchronized void logTimings() {
         }
     }
 
+    /**
+     * Mark a call to an internal API made by this build.
+     * @param call a representation of the call site; for example, {@code hudson.model.Run.setDescription}
+     */
+    synchronized void recordInternalCall(String call) {
```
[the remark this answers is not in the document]
jglick: There should not be a lot of calls—a typical build will never call this.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Again, probably overkill for an infrequently-modified collection.

src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -0,0 +1,130 @@
...
+package org.jenkinsci.plugins.workflow.cps;
+
+import com.cloudbees.groovy.cps.impl.CallSiteBlock;
+import com.cloudbees.groovy.cps.sandbox.Invoker;
+import java.util.function.Supplier;
+
+/**
+ * Captures CPS-transformed events.
+ */
+final class LoggingInvoker implements Invoker {
```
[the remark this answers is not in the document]
jglick: > check for isInternal first, then worry about constructing strings to record

That is exactly what it already does.

src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -0,0 +1,130 @@
...
+        execution.recordInternalCall(call);
+    }
+
+    private static boolean isInternal(Class<?> clazz) {
+        // TODO more precise would be the logic in jenkins.security.ClassFilterImpl.isLocationWhitelisted
+        // (simply checking whether the class loader can “see”, say, jenkins/model/Jenkins.class
+        // would falsely mark third-party libs bundled in Jenkins plugins)
+        String name = clazz.getName();
+        // acc. to `find …/jenkinsci/*/src/main/java -type f -exec egrep -h '^package ' {} \; | sort | uniq` this is decent
+        return name.contains("jenkins") || name.contains("hudson") || name.contains("cloudbees");
```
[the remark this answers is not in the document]
jglick: From some very simple examples I did not see any false positives. Definitely agreed that we would want some test coverage.

As noted in the PR description, there are cases such as `RunWrapper.getNumber` that you _do_ want to record. A true false positive would be some call to a class inside `groovy-sandbox`, `groovy-cps`, or `workflow-cps` which is not actually done from the Pipeline script text but is just an artifact of some implementation hack. I would think that any such cases should not be going through `Invoker` to begin with, though.

jglick:
BTW current output for

```groovy
currentBuild.rawBuild.description = '…'
jenkins.model.Jenkins.instance.systemMessage = '…'
```

is:

```
Internal calls for make-internal-calls #10:
  hudson.model.Hudson.systemMessage
  org.jenkinsci.plugins.workflow.job.WorkflowRun.description
  org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper.rawBuild
```

Seems to be missing `jenkins.model.Jenkins.instance` (or `.getInstance`) for whatever reason.

jglick:
Needs to be integrated with JEP-304.

jglick:
@dwnusbaum would this still be wanted?

src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -0,0 +1,155 @@
...
+            assert false : "should never happen";
+            return;
+        }
+        execution.recordInternalCall(call);
+    }
+
+    private static boolean isInternal(Class<?> clazz) {
+        if (clazz == Safepoint.class || clazz == CpsClosure2.class) {
+            return false;
+        }
```
dwnusbaum, line 68: I think `CpsDefaultGroovyMethods` should be ignored here as well.

I ran this plugin's tests with a modification to fail every build where `getInternalCalls` is non-empty to look for false positives. Other than security-related tests that call `Jenkins.get`, the only failure was `CpsFlowDefinition2Test.sandboxInvokerUsed`, which also tracked `com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.$callClosureForMapEntry__groovy_lang_Closure__java_util_Map_Entry`  and `com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.$collectEntries__java_util_Map__java_util_Map__groovy_lang_Closure`.

src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -0,0 +1,159 @@
...
+        return delegate.methodCall(receiver, method, args);
+    }
+
+    @Override public Object constructorCall(Class lhs, Object[] args) throws Throwable {
+        maybeRecord(lhs, () -> lhs.getName() + ".<init>");
+        return delegate.constructorCall(lhs, args);
+    }
+
+    @Override public Object superCall(Class senderType, Object receiver, String method, Object[] args) throws Throwable {
+        maybeRecord(senderType, () -> senderType.getName() + ".<init>");
```
dwnusbaum, line 112: I might be missing something, but `<init>` here seems wrong. Not sure if you want to just track it as `senderType.method` or track it against `senderType.getSuperclass()` though.

jglick: Tried writing some test coverage for a Groovy class extending a Jenkins class but nothing gets reported at all, because we are not using complex logic analogous to `GroovyCallSiteSelector` to determine where a Java member is _defined_; we only check `receiver.class`. So `superCall` is probably useless anyway, though I can change it to report `method` in case it is ever used.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -447,6 +450,21 @@ void logTimings() {
         }
     }
 
+    /**
+     * Mark a call to an internal API made by this build.
+     * @param call a representation of the call site; for example, {@code hudson.model.Run.setDescription}
+     */
+    synchronized void recordInternalCall(@NonNull String call) {
+        if (internalCalls == null) {
+            internalCalls = new TreeSet<>();
```
dwnusbaum, line 459: Given that we have custom serialization logic here anyway, would it make sense to use `ConcurrentSkipListSet` as the in-memory type to avoid having to synchronize on `CpsFlowExecution` for all writes?

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -548,11 +566,15 @@ public void run() {
...
              */
             private Env createInitialEnv() {
-                return Envs.empty( isSandbox() ? new SandboxInvoker() : new DefaultInvoker());
+                return Envs.empty(createInvoker());
             }
         });
     }
 
+    Invoker createInvoker() {
+        return LoggingInvoker.create(isSandbox());
```
dwnusbaum, line 575: Maybe worth adding a system property to bypass `LoggingInvoker` in case there is an issue.

dwnusbaum (approved):
Seems fine in general. I am slightly worried about performance though. It is unfortunate that stuff in `@NonCPS` methods is not detected, since I'd guess that is where most of what we want to detect would happen. Perhaps we could add a heuristic check on the compilation side if this turns out to not help much in practice.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Override public Object methodPointer(Object lhs, String name) {
+        Class<?> clazz = classOf(lhs);
+        maybeRecord(clazz, () -> clazz.getName() + "." + name);
+        return delegate.methodPointer(lhs, name);
+    }
+
+    @Override public Object cast(Object value, Class<?> type, boolean ignoreAutoboxing, boolean coerce, boolean strict) throws Throwable {
+        Class<?> clazz = classOf(value);
+        maybeRecord(clazz, () -> clazz.getName() + " as " + type.getClass().getName());
```
dwnusbaum, line 156: I would probably check `type` instead of `clazz` here (or check both), or just drop this entirely. If `clazz` is a Jenkins type, then we should have already logged it earlier via some method call or property access, so I think this is redundant, but if you are casting some ordinary object to a Jenkins type then if the sandbox is inactive the construction of that object might not be intercepted otherwise. Of course just instantiating the object is probably not interesting, and whatever you do with it is likely to be intercepted separately, so maybe its fine to just not check anything here.

dwnusbaum:
> Sorry, what sort of check are you referring to?

I mean that if this approach turns out to not be useful in practice due to things like `@NonCPS` blocks, we could try a different approach, where we statically analyze the AST for mentions of Jenkins-related imports and classes. That would have its own limitations, but they would be distinct and so maybe together we would catch more than either approach would alone.

jglick:
> statically analyze the AST

Oh I see. Yes that could be considered, in case inspecting only CPS-transformed code turns out to be a significant limitation in practice.

jglick:
(I suppose it would suffice to only consider the bodies of methods marked `@NonCPS`, plus constructors.)

dwnusbaum:
> Perhaps, though I am guessing that most library developers do not even know this annotation exists.

Generally speaking, I agree, but for the kinds of libraries we are concerned about here I am not so sure.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
jglick: OK I can just skip it, probably a corner case.

dwnusbaum (approved):

PR #229, opened 2018-06-01
[JENKINS-51170] Enable the usage of the StepEnvironmentContributor extensionpoint

t-8ch:
Support the new StepEnvironmentContributor hook introduced in
jenkinsci/workflow-step-api-plugin#36.

This depends on jenkinsci/workflow-step-api-plugin#36.
The dependency version number used depends on the incrementalifyiation (JEP 305).
If requested I can perform this and submit another PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.28</version>
+        <version>3.32</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -68,7 +68,7 @@
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.1.0</git-plugin.version>
-        <workflow-support-plugin.version>2.21</workflow-support-plugin.version>
+        <workflow-support-plugin.version>3.2-rc711.a7efba98306d</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.25</groovy-cps.version>
         <structs-plugin.version>1.17</structs-plugin.version>
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.18</version>
+            <version>2.19-rc508.410020b15345</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -64,6 +64,10 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
     }
 
     @Override public EnvVars getEnvironment() throws IOException, InterruptedException {
+      return getEnvironment(getListener());
+    }
+
+    private TaskListener getListener() throws IOException {
         TaskListener listener;
         if (owner instanceof FlowExecutionOwner.Executable) {
             FlowExecutionOwner executionOwner = ((FlowExecutionOwner.Executable) owner).asFlowExecutionOwner();
@@ -75,6 +79,10 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
         } else {
             listener = new LogTaskListener(LOGGER, Level.INFO);
         }
+        return listener;
+    }
+
+    private EnvVars getEnvironment(TaskListener listener) throws IOException, InterruptedException {
         EnvVars e = owner.getEnvironment(listener);
         e.putAll(env);
         return e;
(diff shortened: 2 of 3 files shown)


jglick:
> The dependency version number used depends on the incrementalifyiation (JEP 305).
> If requested I can perform this

Yes should suffice to update the parent (currently to 3.12) and

```bash
mvn incrementals:incrementalify
```

Could be done in this PR so far as I am concerned.

jglick:
(Somehow the CI build seems to be running even without the `consume-incrementals` profile, not sure why. Might be an artifact of mirror setup on the CI builder.)

t-8ch:
I got an error an error with access restrictions (DoNotUse).
Will try again next week, just left for the weekend.

t-8ch:
The incrementalization is blocked on https://github.com/kohsuke/access-modifier/pull/12
(Or we work around by removing some access modifiers)

t-8ch:
The hooks work so far for steps. The logic in https://github.com/jenkinsci/workflow-support-plugin/pull/62 is actually sufficient. This PR only contains tests.

I am not sure about also calling the hooks for `EnvActionImpl`.
This would require to get the current interpreter state from groovy, getting the enclosing scope and somehow using this to fetch the parents scope `StepContext` from somewhere inside the CPS engine.

Any opinions if this part is required/wanted/superfluous/etc?

pom.xml
```diff
@@ -28,12 +28,12 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.4</version>
+        <version>3.12</version>
```
jglick: Recommend merging with #230 to avoid conflicts.

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -88,7 +96,12 @@ private EnvActionImpl() {
     @Override public String getProperty(String propertyName) {
         try {
             CpsThread t = CpsThread.current();
-            return EnvironmentExpander.getEffectiveEnvironment(getEnvironment(), t.getContextVariable(EnvVars.class), t.getContextVariable(EnvironmentExpander.class)).get(propertyName);
+            TaskListener listener = getListener();
+
+            return EnvironmentExpander.getEffectiveEnvironment(
+                    getEnvironment(getListener()), t.getContextVariable(EnvVars.class),
+                    t.getContextVariable(EnvironmentExpander.class), null, listener)
```
jglick: Note that if you follow my suggestion in the API PR, this call will then have identical behavior to what it did before—just avoiding a deprecation warning.

(I see no reason for `env.SOMETHING` to behave differently. There is no particular `StepContext` when you are in the middle of Groovy code. There are contexts for enclosing blocks, but the API is defined to allow setting an environment for a step, not for a chain of blocks.)

src/test/java/org/jenkinsci/plugins/workflow/DynamicEnvironmentExpanderTest.java
```diff
@@ -145,4 +152,64 @@
         }
     }
 
+    @Issue("JENKINS-51170")
+    @Test public void perStepEnvironment() {
+        story.addStep(new Statement() {
```
jglick: BTW we are on Java 8 here now, so you can use

```java
story.then(r -> {
    WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
    // …
});
```

src/test/java/org/jenkinsci/plugins/workflow/DynamicEnvironmentExpanderTest.java
```diff
@@ -145,4 +152,64 @@
...
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                // p.setDefinition(new CpsFlowDefinition("echo(/before VAR=$VAR/); " + StepEnvAdder.class.getCanonicalName() + ".value = 'after'; echo(/after VAR=$VAR/)", false));
```
jglick: Any reason why the commented-out code is still here? Left over from some debugging session?

src/test/java/org/jenkinsci/plugins/workflow/DynamicEnvironmentExpanderTest.java
```diff
@@ -145,4 +152,64 @@
...
+                StepContext context = getContext();
+                String message = this.var + "=" + context.get(EnvVars.class).get(var);
+                context.get(TaskListener.class).getLogger().println(message);
+                return null;
+            }
+        }
+
+        @TestExtension("perStepEnvironment") public static class DescriptorImpl extends StepDescriptor {
+            @Override public String getFunctionName() { return "printEnv"; }
+            @Override public Set<? extends Class<?>> getRequiredContext() { return Collections.emptySet(); }
```
jglick: actually

```java
return ImmutableSet.of(EnvVars.class, TaskListener.class);
```

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
t-8ch: @jglick Ok. Personally I only need the environment inside the steps. I assumed you wanted to keep the behaviour consistent. Let's leave it like that then.

t-8ch:
@jglick I implemented your comments. Thanks for the review!

t-8ch:
I see the following alternatives to aquire a listener:

1) `getExecution().getOwner().getListener()`
  * Is used just above
2) Remove the check that is throwing the exception (only for our internal call)
  * May have side-effects
3) Do the same as `EnvActionImpl.getListener()`
  * Looks to be similar to 1), but falls back to plain logging

Personally I think 3) seems to be the best.

t-8ch:
I implemented prototypes for all three scenarios and all of them pass the testsuites of `workflow-support` and `workflow-cps`.

dwnusbaum:
Thanks for looking into all of the options! Given that the line immediately above the new line in `workflow-support` uses `getExecution().getOwner().getListener()`, option 1 seems the best to me for consistency with that behavior. Option 2 seems a little worrying because of potential side effects like you noted, and it seems like if we used option 3, anything sent to the system logs would probably just end up being unhelpful log spam since it wouldn't be associated with the particular build it came from. So my preference would be 1 over 3 over 2.

t-8ch:
Will do. I won't get around until tomorrow though. If you want feel free to take a shot at it.

jglick:
> getExecution().getOwner().getListener()

Sounds good but better do it in `DefaultStepContext.getListener`: https://github.com/jenkinsci/workflow-support-plugin/pull/62#discussion_r248848737

> Do the same as `EnvActionImpl.getListener()`

No, do not do that.

t-8ch:
I went with 2 with a fallback to 1, as recommended by @jglick . The tests seem to be happy.

PR #230, opened 2018-06-06
Incrementalified

jglick:
Extracted from #229, since I also expect to need incremental versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,12 +28,12 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.4</version>
+        <version>3.13</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-cps</artifactId>
-    <version>2.54-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Groovy</name>
     <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+Groovy+Plugin</url>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -28,12 +28,12 @@
...
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.4</version>
+        <version>3.13</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-cps</artifactId>
-    <version>2.54-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
```
abayer: Because I haven't been paying enough attention - how does this work with `mvn release:prepare release:perform`?

jglick: [Guide](https://github.com/jenkinsci/incrementals-tools/blob/master/README.md#running-maven-releases)

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -535,7 +535,7 @@ public HttpResponse doGenerateSnippet(StaplerRequest req, @QueryParameter String
         return ExtensionList.lookup(SnippetizerLink.class);
     }
 
-    @Restricted(DoNotUse.class)
+    @Restricted(NoExternalUse.class)
```
svanoort: Seems unrelated?

svanoort:
You have the reviews, can you just merge+release the plugin pom version rather than using a special version?

src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
jglick: Mandated by updating the parent POM; works around https://github.com/kohsuke/access-modifier/pull/12.

jglick:
> can you just merge+release the plugin pom version

Will do.

PR #231, opened 2018-06-06
Do not use com.google.common.io.NullOutputStream

jglick:
Subsumes #230 and amends a mistake in #220: use of a beta API from the version of Guava currently bundled in Jenkins core, which has [subsequently been deleted](https://google.github.io/guava/releases/14.0/api/docs/com/google/common/io/NullOutputStream.html), causing mayhem if you try to run Pipeline-based tests with a newer Guava in the test classpath.

From [this search](https://github.com/search?q=user%3Ajenkinsci+%22com.google.common.io.NullOutputStream%22&type=Code) this seems to be the only usage in non-test code of this class except for `openshift-deployer` which has not been touched in three years.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
@@ -535,7 +535,7 @@ import org.kohsuke.stapler.StaplerRequest;
         return ExtensionList.lookup(SnippetizerLink.class);
     }
 
-    @Restricted(DoNotUse.class)
+    @Restricted(NoExternalUse.class)
     @Extension public static class PerJobAdder extends TransientActionFactory<Job> {
 
         @Override public Class<Job> type() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
@@ -25,19 +25,17 @@
 package org.jenkinsci.plugins.workflow.cps.actions;
 
 import com.google.common.collect.Maps;
-import com.google.common.io.NullOutputStream;
 import hudson.EnvVars;
 import jenkins.model.Jenkins;
+import org.apache.commons.io.output.NullOutputStream;
 import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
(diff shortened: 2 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -535,7 +535,7 @@ public HttpResponse doGenerateSnippet(StaplerRequest req, @QueryParameter String
         return ExtensionList.lookup(SnippetizerLink.class);
     }
 
-    @Restricted(DoNotUse.class)
+    @Restricted(NoExternalUse.class)
     @Extension public static class PerJobAdder extends TransientActionFactory<Job> {
```
svanoort: Why?

jglick: Explained in #230.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -276,8 +274,6 @@ Object sanitizeObjectAndRecordMutation(@CheckForNull Object o, @CheckForNull Env
         }
     }
 
-    static final OutputStream NULL_STRM = new NullOutputStream();
```
svanoort: Why did you remove this?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -25,19 +25,17 @@
 package org.jenkinsci.plugins.workflow.cps.actions;
 
 import com.google.common.collect.Maps;
-import com.google.common.io.NullOutputStream;
 import hudson.EnvVars;
 import jenkins.model.Jenkins;
+import org.apache.commons.io.output.NullOutputStream;
```
svanoort: This is fine

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: Because it is a flyweight object (no fields), so there is no need to cache it.

svanoort:
pending build, then merge-ready

jglick:
cf. https://github.com/jenkinsci/jenkins/pull/3481

PR #232, opened 2018-07-02
[JENKINS-52189] Ensure GraphListeners are notified of the FlowStartNode too

svanoort:
[JENKINS-52189](https://issues.jenkins-ci.org/browse/JENKINS-52189)

We might consider initializing the CpsThreadGroup before the FlowStartNode (in CpsFlowExecution) to make this more "correct" but I'm a bit concerned that might have unintended consequences.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
@@ -117,6 +117,19 @@ final class FlowHead implements Serializable {
             this.head = execution.startNodes.push(n);
         }
         execution.storage.storeNode(head, false);
+
+        CpsThreadGroup c = CpsThreadGroup.current();
+        if (c !=null) {
+            // if the manipulation is from within the program executing thread, then
+            // defer the notification till we get to a safe point.
+            c.notifyNewHead(head);
+        } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
```diff
@@ -117,6 +117,19 @@ void newStartNode(FlowStartNode n) throws IOException {
             this.head = execution.startNodes.push(n);
         }
         execution.storage.storeNode(head, false);
+
+        CpsThreadGroup c = CpsThreadGroup.current();
+        if (c !=null) {
```
svanoort: I'm leaving this in, in case we *do* get a  proper CpsThreadGroup down the road, so it does The Right Thing (tm).

t-8ch:
This will currently not work for `GraphListener`s attached using `FlowExecution.addListener()` in `FlowExecutionListener.onRunning()`.
The `FlowStartNode` is emitted before `FlowExecutionListener.onRunning()`.

It is obviously possible to use a global extension, but I was under the impression that `onRunning()` was an intended place to add new listeners.

jglick:
Note: might solve a minor display issue in logs that I noticed in JEP-210 (see #252), that we display the `FlowEndNode` but not the `FlowStartNode`. Depends on the timing of this event I suppose.

PR #234, opened 2018-07-11
[JENKINS-53358] Remove bogus persistence calls due to notifyListeners

svanoort:
[JENKINS-53358](https://issues.jenkins-ci.org/browse/JENKINS-53358)

These should not be invoked in performance-optimized durability mode -- if we're not trying to save with each step run, we shouldn't do it due to listeners firing.  This was an issue that degraded the IO improvements for performance-optimized mode and greatly increased IO calls.  The phrase "CD-660" may be meaningful to some here. 

Also fixes a buglet noted with the persistedClean flag handling.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.54</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.62</jenkins.version>
+        <jenkins.version>2.73.3</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.1.0</git-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -68,8 +68,8 @@
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.1.0</git-plugin.version>
-        <workflow-support-plugin.version>2.17</workflow-support-plugin.version>
-        <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
+        <workflow-support-plugin.version>2.20</workflow-support-plugin.version>
+        <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.24</groovy-cps.version>
         <jenkins-test-harness.version>2.37</jenkins-test-harness.version>
     </properties>
@@ -143,7 +143,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.23</version>
+            <!--<version>2.23</version> -->
+            <version>2.25-rc775.35b435d87ee7</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -143,7 +143,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.21</version>
+            <!--<version>2.23</version> -->
+            <version>2.25-rc775.35b435d87ee7</version>
```
svanoort: Only until I cut the release, then it will be converted to the final release version and this will be cleaned up

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1441,7 +1441,11 @@ void notifyListeners(List<FlowNode> nodes, boolean synchronous) {
                 }
             } finally {
                 if (synchronous) {
-                    bc.abort(); // hack to skip save—we are holding a lock
+                    bc.abort(); //// hack to skip save—we are holding a lock
+                } else if (!(getDurabilityHint().isPersistWithEveryStep()) && !(nodes.stream().anyMatch(x->x instanceof FlowEndNode))) {
+                    // We do not want to invoke gratuitous save calls if we aren't set to persist with every step
+                    // Except that we MUST trigger a persistence call at the very end of the execution
+                    bc.abort();
```
dwnusbaum: We aren't worried about whether this causes JENKINS-50199 to happen more often because even if it does, it should only matter when reloading configuration from disk, and that case is now handled by jenkinsci/workflow-job-plugin#104, right?

svanoort: @dwnusbaum Correct -- and if the cause of JENKINS-50199 is still triggered, then that's a bug somewhere else -- if we were relying on this to trigger persistence, that's a clear bug.

svanoort:
Sigh, looks like some more test flakiness to sort out (AFAICT the hangs are tied to copyLogs, so releasing and incorporating the log-handling rewrite may solve that).

svanoort:
@joseblas Please could you assist with investigating the test failure here, it seems to relate to https://issues.jenkins-ci.org/browse/JENKINS-34256 and some of the failures described here: https://issues.jenkins-ci.org/browse/JENKINS-50199?focusedCommentId=345572&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-345572  and perhaps https://issues.jenkins-ci.org/browse/JENKINS-46507

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1441,7 +1441,11 @@ void notifyListeners(List<FlowNode> nodes, boolean synchronous) {
                 }
             } finally {
                 if (synchronous) {
-                    bc.abort(); // hack to skip save—we are holding a lock
+                    bc.abort(); //// hack to skip save—we are holding a lock
```
jglick: Why the introduced `//`?

svanoort: Just to see if people are paying attention

PR #235, opened 2018-07-30
Add Chinese translation for pipeline replay

LinuxSuRen:


src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayCause/description.jelly
```diff
@@ -28,10 +28,10 @@ THE SOFTWARE.
     <j:set var="original" value="${it.original}"/>
     <j:choose>
         <j:when test="${original != null}">
-            Replayed <a href="${rootURL}/${original.url}" class="model-link">#${it.originalNumber}</a> (<a href="${rootURL}/${it.run.url}replay/diff">diff</a>)
+            ${%Replayed_by_build(rootURL/original.url,it.originalNumber)} (<a href="${rootURL}/${it.run.url}replay/diff">${%diff}</a>)
```
dwnusbaum: `rootURL/original.url` looks a little strange, do the expressions get coerced into strings and concatenated automatically? I would've expected `rootURL + "/" + original.url`. I'll double-check whether this works locally.

LinuxSuRen: I do another test again, and it's an issue. Already fixes.

src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/Messages.properties
```diff
@@ -1,2 +1,4 @@
 Replay.permission.description=Ability to perform a new Pipeline build with an edited script.
 ReplayCommand.shortDescription=Replay a Pipeline build with edited script taken from standard input
+Replay=Replay
```
dwnusbaum: It looks like these keys being capitalized is causing the build to fail. Maybe rename them to `ReplayAction.displayName` and `ReplayCause.shortDescription`?

dwnusbaum:
@LinuxSuRen thanks!

PR #236, opened 2018-08-11
Add keepLog to the list of writable properties

rodrigc:


rodrigc:
See:

* https://github.com/jenkinsci/workflow-support-plugin/pull/69
 * http://javadoc.jenkins.io/plugin/workflow-support/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.html

PR #239, opened 2018-08-29
[JENKINS-53333] Allow steps to be invoked using their full class name

dwnusbaum:
See [JENKINS-53333](https://issues.jenkins-ci.org/browse/JENKINS-53333).

This PR allows Pipeline steps to be specified using the full class name, in case multiple steps exist in the Jenkins instance with the same function name.

For example, this PR allows the following:

    'org.jenkinsci.plugins.workflow.steps.EchoStep'('Hello, world!')
    'org.jenkinsci.plugins.workflow.steps.EchoStep' 'Hello, world!'
    steps.'org.jenkinsci.plugins.workflow.steps.EchoStep'('Hello, world!')
    // all are equivalent to:
    echo 'Hello, world!'

The syntax is weird to say the least, but the following already works today (Verified in `DSLTest#quotedStep`, maybe just Groovy weirdness), so it is at least consistent on that front:

    'echo' 'Hello, world!' // Quotes around echo?!
    steps.'echo' 'Hello, world!'

@jglick suggested something like the following:

    steps['org.jenkinsci.plugins.workflow.steps.EchoStep']('Hello, world!')

That syntax definitely looks better, but seemed a little trickier to implement, since it looks like we'd need to override `GroovyObjectSupport#getProperty` and return a temporary object that hooks back into `DSL#invokeStep` when called. I am happy to go down that route if preferred, but since the current approach was so simple to implement and this doesn't seem like a common use case anyway, I wanted to get some feedback on whether the current syntax is good enough before proceeding.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -32,6 +32,7 @@ import groovy.lang.GroovyObject;
 import groovy.lang.GroovyObjectSupport;
 import groovy.lang.GroovyRuntimeException;
 import hudson.EnvVars;
+import hudson.Util;
 import hudson.model.Computer;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
@@ -53,7 +54,6 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
-import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -106,15 +106,10 @@ public class DSL extends GroovyObjectSupport implements Serializable {
      */
     private transient Map<String,StepDescriptor> stepClassNames;
     /**
-     * Map from ambiguous function names to fully qualified Step class names. Only contains
-     * ambiguous functions. Used to provide context in the warning message logged when invoking an
-     * an ambiguous step.
+     * Set of function names that are reused by distinct {@link StepDescriptor}s and for which
+     * we have not yet warned the user about the ambiguity.
      */
-    private transient Map<String,List<String>> ambiguousFunctions;
-    /**
-     * Set of ambiguous functions for which we have already logged a warning.
-     */
-    private transient Set<String> ambiguousFunctionsLogged;
+    private transient Set<String> unreportedAmbiguousFunctions;
 
     private static final Logger LOGGER = Logger.getLogger(DSL.class.getName());
 
@@ -165,34 +160,24 @@ public class DSL extends GroovyObjectSupport implements Serializable {
                     throw new GroovyRuntimeException(x);
                 }
             }
-            Set<String> ambiguousFunctionNames = new HashSet<>(0);
+            unreportedAmbiguousFunctions = new HashSet<>(0);
             for (StepDescriptor d : StepDescriptor.all()) {
                 String functionName = d.getFunctionName();
                 if (functions.containsKey(functionName)) {
-                    ambiguousFunctionNames.add(functionName);
+                    unreportedAmbiguousFunctions.add(functionName);
                 }
                 // The step descriptor with the highest value for Extension#ordinal() is used in the case of ambiguity.
                 functions.putIfAbsent(functionName, d);
                 stepClassNames.put(d.clazz.getName(), d);
             }
-            ambiguousFunctions = functions.keySet().stream()
-                    .filter(ambiguousFunctionNames::contains)
-                    .collect(Collectors.groupingBy(Function.identity(),
-                            Collectors.mapping(n -> functions.get(n).clazz.getName(), Collectors.toList())));
-            ambiguousFunctionsLogged = new HashSet<>(0);
         }
         final StepDescriptor sd = functions.getOrDefault(name, stepClassNames.get(name));
         if (sd != null) {
-            List<String> ambiguousClassNames = ambiguousFunctions.get(name);
-            if (ambiguousClassNames != null && ambiguousFunctionsLogged.add(name)) {
-                LOGGER.log(Level.WARNING, "Invoking ambiguous Pipeline Step ‘{0}’ ({1}) in {3}. " + 
-                    "The function name could refer to any of the following steps: {2}. " + 
-                    "To avoid ambiguity, you can invoke steps by class name. " +
-                    "For example: steps.''{1}''(... args ...)", new Object[] {
-                        sd.getFunctionName(), sd.clazz.getName(), ambiguousClassNames, handle
-                });
+            if (Util.isOverridden(DSL.class, getClass(), "invokeStep", StepDescriptor.class, Object.class) &&
+                    !Util.isOverridden(DSL.class, getClass(), "invokeStep", StepDescriptor.class, String.class, Object.class)) {
+                return invokeStep(sd, args);
             }
-            return invokeStep(sd,args);
+            return invokeStep(sd, name, args);
         }
         if (SymbolLookup.get().findDescriptor(Describable.class, name) != null) {
             return invokeDescribable(name,args);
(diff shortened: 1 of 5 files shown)


abayer:
I...hmm. I dunno. I don't love the syntax - my first instinct is to maybe have something like `runStep('com.example.SomeStep', ['args'])` but I'm not sure about that either. Lemme ponder this for a bit?

abayer:
Ok, I thought about it more, and I'm actually leaning more towards the approach @dwnusbaum has taken here. It's ugly, but it's also really really edge-casey - we never actually hit this problem at all until very recently, after all. I think the real question isn't how do we work around duplicate-named steps, it's how do we *document* the workaround. I'd lean towards logging a warning when a "duplicate" step gets invoked with a message saying "hey, this is a duplicate - you're running org.foo.SomeStep, but you might have meant to run org.bar.baz.SomeStep. If you wanted to do that, do 'org.bar.baz.SomeStep'() instead".

dwnusbaum:
The goal here is to provide a workaround in case plugin authors are not quick to fix the issue, but I do think it would make sense to add a warning so admins can be informed of the problem and the workaround, so I'll add that.

abayer:
@dwnusbaum Yeah, I'm pretty sure it's doable - tweaks to `DSL#invokeMethod` and `DSL.functions` to note when there are multiple `StepDescriptor`s with the same name should be easy enough. I'd definitely say pursue that approach.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -144,12 +165,33 @@ public Object invokeMethod(String name, Object args) {
...
+                    .collect(Collectors.groupingBy(Function.identity(),
+                            Collectors.mapping(n -> functions.get(n).clazz.getName(), Collectors.toList())));
+            ambiguousFunctionsLogged = new HashSet<>(0);
         }
-        final StepDescriptor sd = functions.get(name);
+        final StepDescriptor sd = functions.getOrDefault(name, stepClassNames.get(name));
         if (sd != null) {
+            List<String> ambiguousClassNames = ambiguousFunctions.get(name);
+            if (ambiguousClassNames != null && ambiguousFunctionsLogged.add(name)) {
+                LOGGER.log(Level.WARNING, "Invoking ambiguous Pipeline Step ‘{0}’ ({1}) in {3}. " +
```
dwnusbaum: I would prefer to send this to the log of the the running build instead, but to do so, we need the `StepContext` in `invokeStep`, but `invokeStep` has no way to know if the step was invoked using the ambiguous function name or the unambiguous class name. We could change the signature of `invokeStep` to include that information, but this class is public, and the method is private, so I'm not sure if that would be ok. My other thought was to store the name used to invoke the step in a field on the class so we could use it in `invokeStep`, but that would be completely broken if `invokeMethod` can be called concurrently. Thoughts?

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -144,12 +165,33 @@ public Object invokeMethod(String name, Object args) {
...
             }
+            Set<String> ambiguousFunctionNames = new HashSet<>(0);
             for (StepDescriptor d : StepDescriptor.all()) {
-                functions.put(d.getFunctionName(), d);
+                String functionName = d.getFunctionName();
+                if (functions.containsKey(functionName)) {
+                    ambiguousFunctionNames.add(functionName);
+                }
+                // TODO: Switch we switch to putIfAbsent so that the descriptor with the highest ordinal value is preferred for ambiguous functions?
+                functions.put(functionName, d);
```
dwnusbaum: Changing this so that the descriptor with the highest ordinal value is used in the case of ambiguity would be a breaking change for any pipelines that currently happen to be using the descriptor with the lowest ordinal value successfully. However, the only ambiguity I know of involves two descriptors whose ordinal is 0, and I am not sure if the loading order is even deterministic in that case. If not, then it seems like now would be as good a time as any to pin down the behavior.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: Maybe we could do something like 
```
CpsThread thread = CpsThread.current();
if (thread != null) {
  TaskListener listener = thread.getContextVariable(TaskListener.class)
  if (listener != null) {
    listener.getLogger().println("MESSAGE HERE")
  }
}
```

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: Yeah, I'm 100% fine with that.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
abayer: ha, I forgot semicolons.

dwnusbaum: I'll try that, thanks!

dwnusbaum: It looks like that doesn't quite work, because `thread.getContextVariable(TaskListener.class)` is null (and from looking through usages I think it could only be non-null inside of an async CpsBodyExecution). It also looks like [DefaultStepContext#get](https://github.com/jenkinsci/workflow-support-plugin/blob/8105510ba5617c19e90c8cf5ee467fb559833d8d/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java#L73) is the place that constructs TaskListeners from scratch, so I think we do need an actual context.

abayer: Well, boo. Then maybe change the signature for `invokeStep` but also add an overload with the existing signature, which would call the new signature with `null` for the step name?

dwnusbaum: Ok, I'll take that approach for now, thanks!

dwnusbaum: (done in 90e1e1b)

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -165,34 +160,24 @@ public Object invokeMethod(String name, Object args) {
...
-            List<String> ambiguousClassNames = ambiguousFunctions.get(name);
-            if (ambiguousClassNames != null && ambiguousFunctionsLogged.add(name)) {
-                LOGGER.log(Level.WARNING, "Invoking ambiguous Pipeline Step ‘{0}’ ({1}) in {3}. " + 
-                    "The function name could refer to any of the following steps: {2}. " + 
-                    "To avoid ambiguity, you can invoke steps by class name. " +
-                    "For example: steps.''{1}''(... args ...)", new Object[] {
-                        sd.getFunctionName(), sd.clazz.getName(), ambiguousClassNames, handle
-                });
+            if (Util.isOverridden(DSL.class, getClass(), "invokeStep", StepDescriptor.class, Object.class) &&
+                    !Util.isOverridden(DSL.class, getClass(), "invokeStep", StepDescriptor.class, String.class, Object.class)) {
```
abayer: ha, nice. Weird, probably in practice unneeded (I really don't think there's anything extending `DSL` out there anywhere), but why not? =)

abayer: Aaaah, makes sense.

dwnusbaum:
Thanks for the reviews! I just updated the SnippitizerTest `invokeStep` override and fixed an oversight in `reportAmbiguousStepInvocation` that logged an additional message when it didn't make sense. I plan to go ahead and merge and release today once the build passes, but if I get busy with other things then I will likely wait until Monday.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -144,13 +160,24 @@ public Object invokeMethod(String name, Object args) {
...
+                functions.putIfAbsent(functionName, d);
+                stepClassNames.put(d.clazz.getName(), d);
             }
         }
-        final StepDescriptor sd = functions.get(name);
+        final StepDescriptor sd = functions.getOrDefault(name, stepClassNames.get(name));
         if (sd != null) {
-            return invokeStep(sd,args);
+            if (Util.isOverridden(DSL.class, getClass(), "invokeStep", StepDescriptor.class, Object.class) &&
+                    !Util.isOverridden(DSL.class, getClass(), "invokeStep", StepDescriptor.class, String.class, Object.class)) {
```
jglick: Probably needless complexity. I know of no reason to subclass `DSL` except for the trick in `Snippetizer`.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -178,8 +205,19 @@ public Object invokeMethod(String name, Object args) {
 
     /**
      * When {@link #invokeMethod(String, Object)} is calling a {@link StepDescriptor}
+     * @deprecated Prefer {@link #invokeStep(StepDescriptor, String, Object)}
```
jglick: Noticed while resolving merge conflicts in #211 that this is still being called for metasteps like `core` or `wrap`, in which case the `functionName` is not very useful. Amending that to pass the symbol, though really tracking ambiguity there would require more work.

PR #240, opened 2018-09-05
[JENKINS-53240] Verify fix for script-security issue

abayer:
[JENKINS-53240](https://issues.jenkins-ci.org/browse/JENKINS-53420)

See https://github.com/jenkinsci/script-security-plugin/pull/223


abayer:
fwiw, I verified that a pile of tests failed when run against `script-security` 1.45, so the very fact that they passed for me locally with the SNAPSHOT is a good sign. =)

PR #243, opened 2018-09-12
use ABORTED icon color for interrupted steps

ArturHarasimiuk:
parallel() aborts branch execution when failFast == true and another
branch failed. In this case RED color is used in 'Pipeline Steps' view.
It makes more difficult to find real reason of failure because some of
red dots are misleading.

With this change we use BallColor.ABORTED when FlowInterruptedException
is the reason of step failure. Thanks to this user can easier find root
cause of build failure.

Signed-off-by: Artur Harasimiuk <artur.harasimiuk@intel.com>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepAtomNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepAtomNode.java
@@ -25,10 +25,12 @@
 package org.jenkinsci.plugins.workflow.cps.nodes;
 
 import hudson.model.Action;
+import hudson.model.BallColor;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.graph.AtomNode;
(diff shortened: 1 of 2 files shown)


svanoort:
Thanks for the contribution -- I have some mixed feelings about this because the FlowInterruptedException is used widely (and sometimes in unexpected ways) and this might result in misleading results for some cases.  My concern is a repeat of the fiasco with the UNSTABLE statuses. 

It probably also needs a related change in https://github.com/jenkinsci/pipeline-graph-analysis-plugin since that's what's used for the bulk of advanced visualizations. 

Would like a second opinion from someone else (@jglick @dwnusbaum ) before merging.

ArturHarasimiuk:
thanks!

I'm using the same logic as [com.cloudbees.workflow.rest.external.StatusExt.java#L53](https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/master/rest-api/src/main/java/com/cloudbees/workflow/rest/external/StatusExt.java#L53). Not sure if this helps.

ArturHarasimiuk:
moving to workflow-api per @jglick comment: https://github.com/jenkinsci/workflow-api-plugin/pull/78

PR #245, opened 2018-09-21
[JENKINS-53709] Pop context variables after CpsBodyExecution completes

dwnusbaum:
See [JENKINS-53079](https://issues.jenkins-ci.org/browse/JENKINS-53709).

Despite being annotated with `@PersistIn(NONE)`, `ContextVariableSet`s appear to have been persisted as part of `CpsThread` since at least 2014. Marking that field transient produces a ton of test failures, so I am assuming that the annotation was just incorrect and that it is actually ok and intentional that they be serialized as part of the program.

However, the context variables need to be cleared after a `CpsBodyExecution` completes, or else the variables will still be persisted along with the CpsThread even if it is executing unrelated code. In particular, this persistence causes issues when the context variables include an executor, as `ExecutorPickle` will attempt to rehydrate the executor upon deserialization even if the current pipeline code does not require the executor at all.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -348,7 +348,7 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-            synchronized (this) {
+            synchronized (CpsBodyExecution.this) {
                 thread.popContextVariables();
             }
             return Next.terminate(null);
@@ -367,7 +367,7 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onSuccess(sc, o);
             }
-            synchronized (this) {
+            synchronized (CpsBodyExecution.this) {
                 thread.popContextVariables();
             }
             return Next.terminate(null);
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -348,7 +348,7 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-
+            thread.popContextVariables();
             return Next.terminate(null);
         }
 
@@ -365,6 +365,7 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onSuccess(sc, o);
             }
+            thread.popContextVariables();
             return Next.terminate(null);
         }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -348,7 +348,9 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-            thread.popContextVariables();
+            synchronized (this) {
+                thread.popContextVariables();
+            }
             return Next.terminate(null);
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -135,6 +135,16 @@ public ContextVariableSet getContextVariables() {
         return contextVariables;
     }
 
+    /**
+     * Used to remove context variables after a {@link CpsBodyExecution} finishes executing so
+     * that those variables will no longer be persisted as part of the program. See JENKINS-53709.
+     */
+    void popContextVariables() {
+        if (contextVariables != null) {
+            contextVariables = contextVariables.getParent();
```
dwnusbaum: I think it makes sense to treat the variables like a stack and pop them to get the parent variables rather than just nulling them out, but I am not entirely sure.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -348,7 +348,7 @@ public Next receive(Object o) {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-
+            thread.popContextVariables();
```
dwnusbaum: I think the earlier call to `setOutcome` here and in SuccessAdapter keep us from needing to worry about this being called concurrently or more than once.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -97,7 +97,7 @@
     final FlowHead head;
 
     @Nullable
-    private final ContextVariableSet contextVariables;
+    private ContextVariableSet contextVariables;
```
dwnusbaum: Maybe this should be volatile? Synchronization is taken care of in `CpsBodyExecution`s handlers, but I think `getContextVariables` could be called from other threads so we need to make sure they see the latest value for reads. I need to spend a bit more time thinking about this.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -365,6 +367,9 @@ public Next receive(Object o) {
             for (BodyExecutionCallback c : callbacks) {
                 c.onSuccess(sc, o);
             }
+            synchronized (this) {
+                thread.popContextVariables();
```
dwnusbaum: I think the call to `popContextVariables` should be fast enough that there is no issue, but we could change it to the following if anyone prefers:

    CpsThread t;
    synchronized (this) {
        t = thread;
    }
    t.popContextVariables();

jglick:
> `ExecutorPickle` will attempt to rehydrate the executor upon deserialization

Another ill effect of the decision to use `Pickle`s rather than placing the burden of restart handling solely in `StepExecution.onResume`. Not sure whether it is too late to fix this design.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
jglick: > `getContextVariables` could be called from other threads

Whatever calls `StepContext.get(Class)`, yes. But all that should be done by the time the body completes anyway.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
jglick: I think there is just a different `CpsThread` in that case. If this were not true, there would already be a more visible bug (not just a leak). Possibly suffices to just

```java
contextVariables = null;
```

You could check by having some block-scoped step bind a context variable, then having another step run after its body completes check whether that variable is still accessible.

dwnusbaum: I did try something like the following before my change, and it failed because the `sh` step cannot find the context variables as expected. Is that what you mean?

    node {
        // Binds FilePath, Executor, TaskListener, context variables (I think)
    }
    sh('echo "foo"') // Context variables should be gone here

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -348,7 +348,9 @@ public Next receive(Object o) {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-
+            synchronized (CpsBodyExecution.this) {
+                thread.popContextVariables();
```
jglick: `thread` might be null here (see `!isLaunched()` clause above). I will fix that in #279.

PR #247, opened 2018-09-22
[DO NOT MERGE] Test the Java 11 supporting workflow-support in combination

svanoort:
Triggering CI against fixes to support Java 10/11 (this build is ensuring it does not cause regressions)

CC @oleg-nenashev since it fulfills a hackathon story


dwnusbaum:
workflow-support 3.0 was released a while ago, and we just bumped to 3.2 as part of #229, so I am going ahead and closing this.

PR #248, opened 2018-09-25
Migrate Chinese localization resource files

LinuxSuRen:
Migrate Chinese localization resource files into https://github.com/jenkinsci/localization-zh-cn-plugin


LinuxSuRen:
@dwnusbaum I will migrate all localization files finally. But I still have some issues need to resolve.

dwnusbaum:
@LinuxSuRen Ok, I just wasn't sure if it was an oversight or intentional. If you are ok with leaving the Messages.properties files here for now, then I am fine with merging this as-is.

LinuxSuRen:
OK, we take care of Messages.properties later. @dwnusbaum

PR #250, opened 2018-10-01
[JENKINS-40109] Make compilation errors serializable

dwnusbaum:
See [JENKINS-40109](https://issues.jenkins-ci.org/browse/JENKINS-40109).

This PR makes compilation errors thrown by `LoadStep` serializable, and exposes an exception that is reused to fix similar errors when loading shared libraries in jenkinsci/workflow-cps-global-lib-plugin#54

@reviewbybees


abayer:
fwiw, it wouldn't shock me if there are other places where plugins may want to use this - I feel like there are other non-coreish plugins that do loading/compilation within a Pipeline. Ok, yeah, there's definitely `pipeline-remote-loader` but I think that one's deprecated anyway. Still, another good reason to put the exception here. =)

dwnusbaum:
Downstream PR is up at jenkinsci/workflow-cps-global-lib-plugin/pull/54, so I'm removing the work-in-progress label.

PR #252, opened 2018-10-03
[JEP-210] Integration testing

jglick:
My own version of #251. Picks up https://github.com/jenkinsci/workflow-job-plugin/pull/27, #253, #254,  and https://github.com/jenkinsci/structs-plugin/pull/41.

PR #254, opened 2018-10-04
Making sure ArgumentsAction is attached in time for GraphListener.Synchronous notification

jglick:
Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/75. Extracted from #252 as an independently releasable fix necessary to avoid regressing `StepNodeTest` (specifically, proper display of metasteps in the classic UI console) after changes in JEP-210. Specifically, as of https://github.com/jenkinsci/workflow-job-plugin/pull/27#discussion_r222405174 `FlowNode.getDisplayFunctionName` is called from within a synchronous listener. (Formerly this was an asynchronous listener, which sounds wrong but then again `copyLogs` was asynchronous too so who cared?) And this method call pays attention to `ArgumentsAction` to display, for example, `archiveArtifacts` rather than `step` (the generic function name of `CoreStep`). So we need to ensure that at least this basic metadata about the `FlowNode` is actually present by the time listeners are told about it. Some other metadata gets added later, like `LabelAction`, but we tolerate that—the new console UI renders the label via a `ConsoleNote`, potentially much later, at which point the label is known.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -68,7 +68,6 @@ import org.jenkinsci.plugins.structs.describable.DescribableParameter;
 import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
 import static org.jenkinsci.plugins.workflow.cps.ThreadTaskResult.*;
 
-import org.jenkinsci.plugins.workflow.actions.TimingAction;
 import org.jenkinsci.plugins.workflow.cps.actions.ArgumentsActionImpl;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepAtomNode;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
@@ -78,7 +77,6 @@ import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.
 import org.jenkinsci.plugins.workflow.cps.steps.LoadStep;
 import org.jenkinsci.plugins.workflow.cps.steps.ParallelStep;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
-import org.jenkinsci.plugins.workflow.graph.AtomNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
 import org.jenkinsci.plugins.workflow.steps.MissingContextVariableException;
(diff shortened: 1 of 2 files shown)


svanoort:
For those who will be equally confused about the key missing piece: this depends on the upstream PR because we need to be able to get the Environment a bit early (before we have the FlowNode set up and everything).

jglick:
Note that the entire reason `ArgumentsActionImpl` needs an `EnvVars` to begin with is a hack. See [JENKINS-47101](https://issues.jenkins-ci.org/browse/JENKINS-47101) for discussion.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -247,21 +262,6 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
             d.checkContextAvailability(context);
             Thread.currentThread().setContextClassLoader(CpsVmExecutorService.ORIGINAL_CONTEXT_CLASS_LOADER.get());
```
svanoort: Things That Go Bump In The Night: changing the classloader used to run code -- which is a tongue-in-cheek way of saying "there may be a monster here, but I'm not 100% sure."

jglick: No, this call has test coverage IIRC. Anyway it is not related to this PR.

svanoort: Never mind, retracted, I missed something.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -229,14 +227,31 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
...
-            thread.head.setNewHead(an);
         } else {
             an = new StepStartNode(exec, d, thread.head.get());
-            thread.head.setNewHead(an);
         }
 
-        final CpsStepContext context = new CpsStepContext(d,thread,handle,an,ps.body);
+        // Ensure ArgumentsAction is attached before we notify even synchronous listeners:
+        final CpsStepContext context = new CpsStepContext(d, thread, handle, an, ps.body);
+        try {
```
svanoort: I *think* that you want the classloader switch to happen before the try block for safety, no?  Or is there a reason why we shouldn't continue running this logic under the `CpsVmExecutorService.ORIGINAL_CONTEXT_CLASS_LOADER.get()` anymore?

jglick: It never needed to be run with a special CCL to begin with. That is just where you happened to drop that new code. The CCL is for user code, mainly `StepExecution.start` implementations.

svanoort: The `ORIGINAL_CONTEXT_CLASSLOADER` is for non-Groovy code though IIRC -- and IIRC we are switching away from the GroovyClassLoader to avoid stuff being loaded there (and potential memory leaks or issues with classloaders not matching).

I *think* you still want this to be happening with the non-Groovy classloader.

jglick: > The `ORIGINAL_CONTEXT_CLASSLOADER` is for non-Groovy code though

For non-Groovy _plugin_ code, yes.

> we are switching away from the GroovyClassLoader to avoid stuff being loaded there (and potential memory leaks or issues with classloaders not matching)

We are switching to the CCL used for general plugin Java code while we run the general plugin Java code: `StepDescriptor.newInstance`, `Step.start`, and (especially) `StepExecution.start`. `ArgumentsActionImpl` has nothing to do with this.

jglick: FTR this system is from #50. #98 happened to drop `ArgumentsActionImpl` into the same `try` block as a convenience AFAICT—because it was checking for `instanceof ParallelStep`, which can only be done after instantiating the `Step`. This PR instead checks the `StepDescriptor`, which can be done earlier.

PR #256, opened 2018-10-07
Avoid deprecated api usage

darxriggs:
This is based on the analysis https://ci.jenkins.io/job/Infra/job/deprecated-usage-in-plugins/job/master/566/artifact/output/usage-by-plugin.html.


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1457,7 +1457,12 @@ void notifyListeners(List<FlowNode> nodes, boolean synchronous) {
             return ACL.SYSTEM;
         }
         try {
-            return User.get(user).impersonate();
+            User u = User.getById(user, false);
```
dwnusbaum: I have not looked into it in detail, but I think the test time out in `WorkflowTest#authentication` is because the previous code was equivalent to `User.getById(user, true)`. It's not clear to me whether `User.getById(user, false)` is correct and the test should be modified or if the code should be `User.getById(user, true)`.

darxriggs: Hi @jglick, I have seen that you added the `getAuthentication` method in 2014 via commit 822a09b67f4abe4da0092e2f557a676c9d9fde9d.
Could you shed some light on how the deprecated `User::get` should be replaced here?

Options
a) `User.getById(user, false)`
b) `User.getById(user, true)`
c) `User.getOrCreateByIdOrFullName(user)`

I am not that familiar with all the user management in Jenkins. I was expecting that the user already exists when this method is called, therefore I would not create one. But the test fails because no user is available after restarting Jenkins. Having a look at `User::getOrCreate` it shows that the user "someone" in the test is not saved to disk because `id` and `fullName` are equal. Therefore I am not sure how to proceed.

jglick: I think `User.getById(user, true)` (the direct replacement here of `User.get(user)` would be appropriate. Jenkins APIs are confusing. Even though a user may be defined in, say, LDAP, meaning there is a valid `Authentication` with a given ID, that does not necessarily mean a `User` object exists. Generally speaking, a `User` object is created on demand from an ID, or if there _happened_ to be some reason to persist their `config.xml`—a customized display name, some `UserProperty`s, etc.

darxriggs:
I rebased on master, fixed the reported issue and remove some more deprecated API usages.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Yes you want `getById`, since this is specifically an ID.

darxriggs:
Thank you.

PR #257, opened 2018-10-11
[JENKINS-54032] Harden ArgumentsAction further by protecting against bizarre inputs

svanoort:
Implements improvement [JENKINS-54032](https://issues.jenkins-ci.org/browse/JENKINS-54032) -- protecting us from retaining in-memory (and potentially serialized on disk) some of the crazy things that steps can take as inputs. 

Also incidentally protects against possibility of passing in `Describable` types that include security-sensitive information, by un-instantiating to get their argument info and applying masking.


src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -103,6 +108,22 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
...
     }
 
+    /** Restrict stored arguments to a reasonable subset of types so we don't retain totally arbitrary objects
+     *  in memory. Generally we aim to allow storing anything that maps correctly to an {@link UninstantiatedDescribable}
+     *  But we may allow a few extras doesn't address, if they're safe to store.
+     */
+    boolean isStoreableType(Object ob) {
+        if (ob == null) {
+            return true;
+        } else if (ob instanceof CharSequence || ob instanceof Number || ob instanceof Boolean
```
svanoort: I'm really not 100% on which types are legal here, so I'm opting to include a few extra that seem to be mentioned within the Structs plugin -- and some things that we know are safe and useful to store/visualize in some fashion.

svanoort: Note that we're putting this in most-likely order after the nullcheck (most steps take Strings, booleans, or numbers).

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -103,6 +108,22 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
...
+                || ob instanceof Map || ob instanceof List || ob instanceof UninstantiatedDescribable
+                || ob instanceof URL || ob instanceof Result || ob instanceof Exception) {
+            return true;
+        }
+        Class c = ob.getClass();
+        return c.isPrimitive() || c.isEnum() || c.isArray();
```
rsandell: So you are allowing `EvilObject[]` ?

svanoort: @rsandell IIRC, it gets coerced to a List internally (see https://github.com/jenkinsci/structs-plugin/blob/8d4446fe64ce79fa476345d49cdbb100307af76c/plugin/src/main/java/org/jenkinsci/plugins/structs/describable/DescribableModel.java#L262) but... y'know, the way DescribableModels work is kind of black magic, so better safe than sorry and I'll add Array sanitization.

svanoort: aand, done @rsandell

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -103,6 +108,22 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
...
+    /** Restrict stored arguments to a reasonable subset of types so we don't retain totally arbitrary objects
+     *  in memory. Generally we aim to allow storing anything that maps correctly to an {@link UninstantiatedDescribable}
+     *  But we may allow a few extras doesn't address, if they're safe to store.
+     */
+    boolean isStoreableType(Object ob) {
+        if (ob == null) {
+            return true;
+        } else if (ob instanceof CharSequence || ob instanceof Number || ob instanceof Boolean
+                || ob instanceof Map || ob instanceof List || ob instanceof UninstantiatedDescribable
+                || ob instanceof URL || ob instanceof Result || ob instanceof Exception) {
```
dwnusbaum: Any reason to use `Exception` instead of `Throwable`?

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
dwnusbaum: How common do we think enums/arrays are? Maybe it would make sense to move `c.isPrimitive() || c.isEnum() || c.isArray()` up before the `instanceof` checks, although I would be surprised if it made a difference in practice.

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -103,6 +108,22 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
                 : true;
     }
 
+    /** Restrict stored arguments to a reasonable subset of types so we don't retain totally arbitrary objects
+     *  in memory. Generally we aim to allow storing anything that maps correctly to an {@link UninstantiatedDescribable}
+     *  But we may allow a few extras doesn't address, if they're safe to store.
+     */
+    boolean isStoreableType(Object ob) {
```
jglick: s/storeable/storable/

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -236,24 +275,39 @@ Object sanitizeListAndRecordMutation(@Nonnull List objects, @CheckForNull EnvVar
...
+            return NotStoredReason.UNSERIALIZABLE;
+        }
+
         Object modded = tempVal;
         if (modded instanceof Map) {
             // Recursive sanitization, oh my!
             modded = sanitizeMapAndRecordMutation((Map)modded, vars);
         } else if (modded instanceof List) {
             modded = sanitizeListAndRecordMutation((List) modded, vars);
+        } else if (modded != null && modded.getClass().isArray()) {
```
jglick: Not sure why we are dealing with arrays here—`structs` deals with `List`s, or at least normalizes to them, but perhaps you are still working with unprocessed direct input from `StepDescriptor.instantiate`?

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -223,6 +214,25 @@ public void testRecursiveSanitizationOfContent() {
...
     }
 
+    @Test
+    public void testArraySanitization() {
+        EnvVars env = new EnvVars();
+        String secretUsername = "IAmA";
+        env.put("USERVARIABLE", secretUsername); // assume secretuser is a bound credential
+
+        HashMap<String, Object> args = new HashMap<>();
+        args.put("ints", new int[]{1,2,3});
```
jglick: Not a legal argument type as per `structs`.

svanoort:
test failure appears to be an unrelated test flake

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: Not particularly, I can switch it

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: How about "adorableType"?  And we just say all the other types are too ugly to be adorable. 
 :trollface:

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -103,6 +108,22 @@ public static boolean isStringSafe(@CheckForNull String input, @CheckForNull Env
...
 
+    /** Restrict stored arguments to a reasonable subset of types so we don't retain totally arbitrary objects
+     *  in memory. Generally we aim to allow storing anything that maps correctly to an {@link UninstantiatedDescribable}
+     *  But we may allow a few extras doesn't address, if they're safe to store.
+     */
+    boolean isStoreableType(Object ob) {
+        if (ob == null) {
+            return true;
+        } else if (ob instanceof CharSequence || ob instanceof Number || ob instanceof Boolean
+                || ob instanceof Map || ob instanceof List || ob instanceof UninstantiatedDescribable
```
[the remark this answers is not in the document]
svanoort: Huh?  Map is used to represent an exploded class.  Anyway, we would need to pass that through anyway as legal since we're using map to hold the exploded arguments (or using `#toMap` on `UninstantiatedDescribable`s).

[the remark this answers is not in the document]
jglick: I should have been clearer. `Map` is used to _represent_ `UninstantiatedDescribable` but it is not honored as a supported data type for input per se. For example, you cannot declare a step (or any `Describable`) parameter to be of type `Map<String, Integer>`.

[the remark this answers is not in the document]
jglick: IOW you _can_ decide to record such anomalous arguments, provided it seems safe to do so, but you _need not_ record anything outside of the officially supported data types.

`ChoiceParameterDefinition` and its usage of `List<String>` is a special mess. See https://github.com/jenkinsci/jenkins/pull/3014. The proper way to solve this would have been to declare `choices` to be of type `List<Choice>` where

```java
public static class Choice extends AbstractDescribableImpl<Choice> {
    public final String value;
    @DataBoundConstructor public Choice(String value) {this.value = value;}
    // optionally:
    @DataBoundSetter public String displayName;
    @Symbol("choice")
    @Extension public static class DescriptorImpl extends Descriptor<Choice> {}
}
```

which would allow use of normal `config.xml` idioms and also allow Pipeline usage like

```groovy
properties([parameters([choice(name: 'mode', choices: [choice('full'), choice('partial')])])])
```

without any special casing except a `newInstance` override to handle compatibility with scripts passing the old syntax

```groovy
choices: 'full\npartial'
```

src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
svanoort: Yes, my assumption is that it *might* be possible -- or in any case, that it's probably safer to deal with it than not to.

svanoort:
@dwnusbaum The most common types for arguments are (roughly in order) null, String, Boolean, and Number (usually an int or log) or Enum.  The other checks are mostly for catchall purposes.  

We could consider moving the Enum check earlier as a micro-optimization (and maybe do some other re-ordering of this) if this method comes up as a bottleneck in the future.

PR #261, opened 2018-11-07
Parent update

jglick:
Testing https://github.com/jenkinsci/plugin-pom/pull/130.


jglick:
Failing on Windows like https://github.com/jenkinsci/jenkins/pull/3720. Will look for a smaller plugin to test against.

PR #266, opened 2018-11-23
[JENKINS-54890] skip error when listener invoke we do not want to break the build if …

olamy:
…a listener is broken

Signed-off-by: olivier lamy <olamy@apache.org>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1434,7 +1434,12 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 for (FlowNode node : nodes) {
                     for (GraphListener listener : toRun) {
                         if (listener instanceof GraphListener.Synchronous == synchronous) {
-                            listener.onNewHead(node);
+                            try {
+                                listener.onNewHead(node);
+                            } catch ( Exception e ) {
+                                LOGGER.fine( "skip error with listener " + listener.getClass().getName()
+                                                 + ": " + e.getMessage() );
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/GraphListenerTest.java
```diff
@@ -0,0 +1,50 @@
...
+            + "}";
+        WorkflowJob j = r.createProject(WorkflowJob.class, "listener");
+        j.setDefinition(new CpsFlowDefinition(script, true));
+        r.buildAndAssertSuccess(j);
+    }
+
+    @TestExtension
+    public static class TestGraphListener implements GraphListener, Serializable
+    {
+        private Random random = new Random();
```
dwnusbaum: nit: Seems better to avoid randomness in the test, maybe by throwing an exception in the listener unconditionally and asserting that the exception's message is printed in the logs.

olamy: @dwnusbaum  good idea

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1434,7 +1434,12 @@ void notifyListeners(List<FlowNode> nodes, boolean synchronous) {
                 for (FlowNode node : nodes) {
                     for (GraphListener listener : toRun) {
                         if (listener instanceof GraphListener.Synchronous == synchronous) {
-                            listener.onNewHead(node);
+                            try {
+                                listener.onNewHead(node);
+                            } catch ( Exception e ) {
+                                LOGGER.fine( "skip error with listener " + listener.getClass().getName()
```
svanoort: You really want to log at a WARNING level for this -- if the listener is failing it usually indicates something fatal going wrong with the Pipeline.

olamy: yup I agree and done.

dwnusbaum:
Closing and reopening to get a fresh build.

PR #271, opened 2018-12-09
[JENKINS-55091] Let global variables expose their logical type

rudolfwalter:
Improve the type information provided by Pipeline global variables. See [JENKINS-55091](https://issues.jenkins-ci.org/browse/JENKINS-55091).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/GlobalVariable.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/GlobalVariable.java
@@ -33,6 +33,7 @@ import hudson.util.Iterators.FlattenIterator;
 import jenkins.model.RunAction2;
 
 import javax.annotation.Nonnull;
+import java.lang.reflect.Type;
 import java.util.Iterator;
 import javax.annotation.CheckForNull;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/GlobalVariable.java
```diff
@@ -69,13 +70,26 @@
...
      * @throws Exception if there was any problem creating it (will be thrown up to the script)
      * @see CpsScript#getProperty
      */
     public abstract @Nonnull Object getValue(@Nonnull CpsScript script) throws Exception;
 
+    /**
+     * Gets the logical type of this global variable - that is, the type returned by {@link #getValue}.
+     * <p>The default implementation returns the return type of {@link #getValue}, as implemented in the subclass.</p>
+     */
+    public Type getType() {
```
jglick, line 85: This is normally done by making the class generic and adding a method returning `Class<T>`.

jglick (commented):
As in https://github.com/jenkinsci/workflow-step-api-plugin/pull/40#pullrequestreview-189529757 I am not sure what this accomplishes exactly. Seems to be a piece of a larger effort not linked here?

src/main/java/org/jenkinsci/plugins/workflow/cps/GlobalVariable.java
rudolfwalter: The idea, iirc after 3 years, was to provide proper type information for tooling; eg, an IDE plugin backed by a real Jenkins instance or Jenkins Test Harness. I chose `Type` over `Class<T>` because `Type` can fully represent types including generics etc.

That said, I've moved on from this, and it's unlikely that I'll work on it in the foreseeable future. Please feel free to close this and the other PR. Sorry for the mess/confusion.

jglick: > an IDE plugin backed by a real Jenkins instance

Like GDSL as in https://github.com/jenkinsci/workflow-step-api-plugin/pull/40#pullrequestreview-189529757?

jglick:
Closing. Someone who uses IDEA (?) may want to pick this up and continue.

PR #275, opened 2019-03-22
[JENKINS-56688] Recommend Map.getOrDefault() over Map.get()

daniel-carrington:
Groovy Map.get(Object, Object) tries to modify the Map.
Using getOrDefault() instead avoids modifying the Map.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/ParamsVariable/help.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/ParamsVariable/help.jelly
@@ -32,7 +32,7 @@ THE SOFTWARE.
     <p>
         or to supply a nontrivial default value:
     </p>
-    <pre><code>if (params.get('BOOLEAN_PARAM_NAME', true)) {doSomething()}</code></pre>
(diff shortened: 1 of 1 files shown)


daniel-carrington:
Hi @dwnusbaum, one thing that I noticed while I was using getOrDefault is that since it's a Java method and not a Groovy method, users will need to approve the `method java.util.Map getOrDefault java.lang.Object java.lang.Object` signature in order to use this. Does this also need to be added to the documentation?

dwnusbaum:
Ah, thanks for pointing that out! In that case, we should add that method to the script-security whitelist [here](https://github.com/jenkinsci/script-security-plugin/blob/7cdf0bdc4e420ac96390f857fc162a98feee7220/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist#L313) (in a pull request to that repository) so it doesn't need to be approved (fine because it is safe for use in the sandbox) before we merge this PR (It would be good to add other similar methods like `compute`, `computeIfAbsent`, etc. at the same time).

jglick:
Amends #191. [JENKINS-56688](https://issues.jenkins-ci.org/browse/JENKINS-56688)

daniel-carrington:
@dwnusbaum, is this ready to merge now that the script-security whitelist change has been merged? Or is there a need for functional tests using this idiom?

PR #278, opened 2019-03-28
Use DefaultStepContext simplification

jglick:
Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/94.

Bound to fail until upstream build completes; currently stuck waiting for a Windows executor, sigh.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.1.0</git-plugin.version>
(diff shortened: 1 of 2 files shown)


jglick:
Hmm, upstream deployed but looks like either the Azure mirror is caching the negative result, or there is a negative cache result in some local repository. Either way, may need to wait a while. [INFRA-1671](https://issues.jenkins-ci.org/browse/INFRA-1671)

dwnusbaum:
workflow-support 3.3. was just released with the downstream change.

PR #279, opened 2019-04-05
[JENKINS-41791] Leaked closures from parallel & load steps

jglick:
[JENKINS-41791](https://issues.jenkins-ci.org/browse/JENKINS-41791)

Have you ever wondered what, exactly, is inside `program.dat`? Well I have. Alas, it is in a binary format used by a relatively obscure serialization framework, and I have never managed to write a translator to a human-readable format. So this `threadDump/program.xml` instead lets you use XStream to see the same information on a running program. You will be **shocked** at what you find after running `parallel` if you do not have this fix. Amends #245 & #83.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -360,11 +360,6 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-            synchronized (CpsBodyExecution.this) {
-                if (thread != null) {
-                    thread.popContextVariables();
-                }
-            }
             return Next.terminate(null);
         }
 
@@ -380,9 +375,6 @@ class CpsBodyExecution extends BodyExecution {
             StepContext sc = new CpsBodySubContext(context, en);
             for (BodyExecutionCallback c : callbacks) {
                 c.onSuccess(sc, o);
-            }
-            synchronized (CpsBodyExecution.this) {
-                thread.popContextVariables();
             }
             return Next.terminate(null);
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -324,10 +324,10 @@ class CpsBodyExecution extends BodyExecution {
     }
 
     private void setOutcome(Outcome o) {
-        if (bodyToUnexport != null && thread != null) {
-            thread.group.unexport(bodyToUnexport);
-        }
         synchronized (this) {
+            if (bodyToUnexport != null && thread != null) {
+                thread.group.unexport(bodyToUnexport);
+            }
             if (outcome!=null)
                 throw new IllegalStateException("Outcome is already set");
             this.outcome = o;
@@ -360,8 +360,8 @@ class CpsBodyExecution extends BodyExecution {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-            if (thread != null) {
-                synchronized (CpsBodyExecution.this) {
+            synchronized (CpsBodyExecution.this) {
+                if (thread != null) {
                     thread.popContextVariables();
                 }
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -42,6 +42,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.*;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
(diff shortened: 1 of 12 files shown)


jglick:
Rrgh, Windows builders…

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -348,8 +360,10 @@ public Next receive(Object o) {
             for (BodyExecutionCallback c : callbacks) {
                 c.onFailure(sc, t);
             }
-            synchronized (CpsBodyExecution.this) {
-                thread.popContextVariables();
+            if (thread != null) {
```
jglick: https://github.com/jenkinsci/workflow-cps-plugin/pull/245#discussion_r272439034

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -376,7 +386,10 @@ private boolean run() {
             if (scripts != null) {
                 scripts.clear();
             }
-            closures.clear();
+            if (!closures.isEmpty()) {
+                LOGGER.log(WARNING, "Stale closures in {0}", execution);
```
jglick: https://github.com/jenkinsci/workflow-cps-plugin/pull/83#discussion_r272443819

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -206,6 +206,10 @@ public boolean start() throws Exception {
                     "semaphore 'wait'\n", true));
             WorkflowRun b = p.scheduleBuild2(0).waitForStart();
             SemaphoreStep.waitForStart("wait/1", b);
+            String xml = ((CpsFlowExecution) b.getExecution()).programPromise.get().asXml();
+            System.out.print(xml);
+            assertThat(xml, not(containsString(s.getWorkspaceFor(p).getRemote())));
+            assertThat(xml, not(containsString("ExecutorStepExecution")));
```
jglick: Change `true` to `false` in `ParallelStepExecution` and you will see all sorts of junk here.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -282,11 +282,11 @@ public CpsBodyInvoker newBodyInvoker() {
         if (body == null) {
             throw new IllegalStateException("There is no body to invoke");
         }
-        return newBodyInvoker(body);
+        return newBodyInvoker(body, false);
```
jglick: Note that this is the normal case—everything _except_ `parallel` and `load`. Just normal block-scoped steps like `node`, (new) `stage`, etc. They already unexported their closure when the step completes in `CpsStepContext.scheduleNextRun` (whether or not the body was ever invoked, or invoked multiply).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -315,6 +324,9 @@ public synchronized Object get(long timeout, TimeUnit unit) throws InterruptedEx
     }
 
     private void setOutcome(Outcome o) {
+        if (bodyToUnexport != null && thread != null) {
```
jglick: Seems the easiest place to do this—called from both the success and failure adapters.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java
```diff
@@ -54,7 +54,7 @@ public boolean start() throws Exception {
 
         // execute body as another thread that shares the same head as this thread
         // as the body can pause.
-        cps.newBodyInvoker(t.getGroup().export(script))
+        cps.newBodyInvoker(t.getGroup().export(script), true)
```
jglick: The reported issue was about `parallel`, but while looking for unmatched calls to `export` I found this one too.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -376,7 +391,10 @@ private boolean run() {
             if (scripts != null) {
                 scripts.clear();
             }
-            closures.clear();
+            if (!closures.isEmpty()) {
+                LOGGER.log(WARNING, "Stale closures in {0}: {1}", new Object[] {execution, closures.keySet()});
```
jglick: https://github.com/jenkinsci/workflow-cps-plugin/pull/83#discussion_r272443819

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -349,7 +361,9 @@ public Next receive(Object o) {
                 c.onFailure(sc, t);
             }
             synchronized (CpsBodyExecution.this) {
-                thread.popContextVariables();
+                if (thread != null) {
+                    thread.popContextVariables();
```
dwnusbaum: As far as I can tell, this call is no longer necessary after the other changes in this PR (in both the success and failure adapters). All of the regression tests I added for JENKINS-53709 pass even if the body of `popContextVariables` is commented out, and the serialized form is identical based on the output of `CpsBodyExecutionTest.popContextVarsOnBodyCompletion`. Perhaps #245 was an incomplete fix for this class of problem, and this PR completes the fix.

jglick: Perhaps. I will check it. Thanks for the tip.

PR #282, opened 2019-04-15
Avoid Groovy interpolation in samples

jglick:
For a variety of reasons it is poor form to use Groovy interpolation in the arguments to `sh` etc. steps. Better to pass in environment variables and expand those in the host shell.

Tested on both Linux and Windows 10 using a Maven installation located in a directory with a space in the name, to exercise the quoting logic a bit.

### DIFF
--- a/src/main/js/samples.js
+++ b/src/main/js/samples.js
@@ -67,10 +67,12 @@ samples.push({
         "   }\n" +
         "   stage('Build') {\n" +
         "      // Run the maven build\n" +
-        "      if (isUnix()) {\n" +
-        "         sh \"'${mvnHome}/bin/mvn' -Dmaven.test.failure.ignore clean package\"\n" +
-        "      } else {\n" +
-        "         bat(/\"${mvnHome}\\bin\\mvn\" -Dmaven.test.failure.ignore clean package/)\n" +
+        "      withEnv([\"MVN_HOME=$mvnHome\"]) {\n" +
+        "         if (isUnix()) {\n" +
+        "            sh '\"$MVN_HOME/bin/mvn\" -Dmaven.test.failure.ignore clean package'\n" +
+        "         } else {\n" +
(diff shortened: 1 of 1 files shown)


jglick:
> what some of those reasons are

[This](https://github.com/jenkinsci/credentials-binding-plugin/blob/37ce338945d63c119a2d9c4101f8e298e5a2d0ca/src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html#L28-L49) and [argument display](https://issues.jenkins-ci.org/browse/JENKINS-47101) and general quoting mayhem.

src/main/js/samples.js
```diff
@@ -67,10 +67,12 @@ samples.push({
...
         "      // Run the maven build\n" +
-        "      if (isUnix()) {\n" +
-        "         sh \"'${mvnHome}/bin/mvn' -Dmaven.test.failure.ignore clean package\"\n" +
-        "      } else {\n" +
-        "         bat(/\"${mvnHome}\\bin\\mvn\" -Dmaven.test.failure.ignore clean package/)\n" +
+        "      withEnv([\"MVN_HOME=$mvnHome\"]) {\n" +
+        "         if (isUnix()) {\n" +
+        "            sh '\"$MVN_HOME/bin/mvn\" -Dmaven.test.failure.ignore clean package'\n" +
+        "         } else {\n" +
+        "            bat(/\"%MVN_HOME%\\bin\\mvn\" -Dmaven.test.failure.ignore clean package/)\n" +
```
daniel-beck: Using "regex" string delimiters here seems weird. Why, other than that the old example did it? https://github.com/jenkinsci/workflow-cps-plugin/commit/b775c2aa77aa0c13390464f73e24dab9bf588c4f#diff-030263ebf957b3984b3dd6a035d5dfb2R73 doesn't explain.

jglick: Makes it easier to use `/` on Windows: no need for an escape.

PR #288, opened 2019-05-10
[JENKINS-56890] Make access to CpsThreadGroup.threads thread-safe

dwnusbaum:
See [JENKINS-56890](https://issues.jenkins-ci.org/browse/JENKINS-56890). Subsumes #287 (CC @wdforson).

Assuming that the issue has to do with reading from the map while it is being updated, this PR makes `CpsThreadGroup.thread` private and wraps it using `Collections.unmodifiableNavigableMap` to synchronize reads and writes (if the issue is instead a timing issue and the value is not yet in the map when the lookup occurs then this wouldn't help, but that seems less likely to me). All existing mutation occurs only on the CPS VM thread, so I think that as long as all iteration and read accesses that happen off of the CPS VM thread are synchronized we are safe. I was not able to reproduce the reported issue in a test, but here is the basis of what I was using to try to reproduce the issue:

```java
@Issue("JENKINS-56890")
@Test public void generalNonBlockingStepExecutionThreadSafety() {
    rr.then(r -> {
        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "demo");
        // I tried various combinations of hundreds of parallel/sequential blocks
        // with synchronous and asynchronous bodies but was unable to reproduce
        // the issue.
                "nonBlocking {\n" +
                "  echo 'foo'\n" +
                "}\n", true));
        WorkflowRun b = r.buildAndAssertSuccess(p);
        r.assertLogNotContains("cannot find current thread", b);
    });
}

/**
 * Step that uses {@link StepContext} in GeneralNonBlockingStepExecution.TailCall while not on the CPS VM thread
 * to demonstrate a concurrency issue in {@link CpsStepContext}.
 */
public static class NonBlockingStep extends Step {
    @DataBoundConstructor
    public NonBlockingStep() { }
    @Override
    public StepExecution start(StepContext context) throws Exception {
        return new Execution(context);
    }
    private static class Execution extends GeneralNonBlockingStepExecution {
        public Execution(StepContext context) {
            super(context);
        }
        @Override
        public boolean start() throws Exception {
            run(() -> {
                getContext().newBodyInvoker().withCallback(new Callback()).start();
            });
            return false;
        }
        private final class Callback extends TailCall {
            @Override
            protected void finished(StepContext context) throws Exception {
                Run<?, ?> run = context.get(Run.class);
                TaskListener listener = context.get(TaskListener.class);
                //listener.getLogger().println("Finished executing nonBlockingStep in " + run);
            }
        }
    }
    @TestExtension("generalNonBlockingStepExecutionThreadSafety")
    public static class DescriptorImpl extends StepDescriptor {
        @Override
        public String getFunctionName() {
            return "nonBlocking";
        }
        @Override
        public boolean takesImplicitBlockArgument() {
            return true;
        }
        @Override
        public Set<? extends Class<?>> getRequiredContext() {
            return ImmutableSet.of(Run.class, TaskListener.class);
        }
    }
}
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -208,7 +208,7 @@ class CpsBodyExecution extends BodyExecution {
                     List<StepExecution> executions = new ArrayList<>();
                     // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
                     Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
-                    for (CpsThread t : g.threads.values()) {
+                    for (CpsThread t : g.getThreads()) {
                         m.put(t.head, t);
                     }
                     for (CpsThread t : m.values()) {
@@ -260,7 +260,7 @@ class CpsBodyExecution extends BodyExecution {
                 public void onSuccess(CpsThreadGroup g) {
                     // Similar to getCurrentExecutions but we want the raw CpsThread, not a StepExecution; cf. CpsFlowExecution.interrupt
                     Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
-                    for (CpsThread t : thread.group.threads.values()) {
+                    for (CpsThread t : thread.group.getThreads()) {
                         m.put(t.head, t);
                     }
                     for (CpsThread t : Iterators.reverse(ImmutableList.copyOf(m.values()))) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
@@ -215,17 +215,12 @@ public class CpsStepContext extends DefaultStepContext { // TODO add XStream cla
 
     /**
      * Returns the thread that is executing this step.
-     * Needs to take {@link CpsThreadGroup} as a parameter to prove that the caller is in CpsVmThread.
      *
      * @return
      *      null if the thread has finished executing.
      */
     @CheckForNull CpsThread getThread(CpsThreadGroup g) {
-        CpsThread thread = g.threads.get(threadId);
-        if (thread == null) {
-            LOGGER.log(Level.FINE, "no thread " + threadId + " among " + g.threads.keySet(), new IllegalStateException());
-        }
-        return thread;
+        return g.getThread(threadId);
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
@@ -14,6 +14,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 
@@ -127,12 +128,8 @@ public final class CpsThreadDump {
 
     public static CpsThreadDump from(CpsThreadGroup g) {
         // all the threads that share the same head form a logically single thread
-        Map<FlowHead, List<CpsThread>> m = new LinkedHashMap<FlowHead,List<CpsThread>>();
-        for (CpsThread t : g.getThreads()) {
-            List<CpsThread> l = m.get(t.head);
-            if (l==null)    m.put(t.head, l = new ArrayList<CpsThread>());
-            l.add(t);
-        }
+        Map<FlowHead, List<CpsThread>> m = g.safelyApplyToThreads(threads -> threads.stream()
+                .collect(Collectors.groupingBy(thread -> thread.head, LinkedHashMap::new, Collectors.toList())));
 
         CpsThreadDump td = new CpsThreadDump();
         for (List<CpsThread> e : m.values())
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
@@ -128,7 +128,7 @@ public final class CpsThreadDump {
     public static CpsThreadDump from(CpsThreadGroup g) {
         // all the threads that share the same head form a logically single thread
         Map<FlowHead, List<CpsThread>> m = new LinkedHashMap<FlowHead,List<CpsThread>>();
-        for (CpsThread t : g.threads.values()) {
+        for (CpsThread t : g.getThreads()) {
             List<CpsThread> l = m.get(t.head);
             if (l==null)    m.put(t.head, l = new ArrayList<CpsThread>());
             l.add(t);
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
@@ -14,7 +14,6 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 
@@ -128,8 +127,12 @@ public final class CpsThreadDump {
 
     public static CpsThreadDump from(CpsThreadGroup g) {
         // all the threads that share the same head form a logically single thread
-        Map<FlowHead, List<CpsThread>> m = g.safelyApplyToThreads(threads -> threads.stream()
-                .collect(Collectors.groupingBy(thread -> thread.head, LinkedHashMap::new, Collectors.toList())));
+        Map<FlowHead, List<CpsThread>> m = new LinkedHashMap<FlowHead,List<CpsThread>>();
+        for (CpsThread t : g.getThreads()) {
+            List<CpsThread> l = m.get(t.head);
+            if (l==null)    m.put(t.head, l = new ArrayList<CpsThread>());
+            l.add(t);
+        }
 
         CpsThreadDump td = new CpsThreadDump();
         for (List<CpsThread> e : m.values())
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -62,13 +62,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
-import java.util.TreeMap;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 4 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -215,17 +215,12 @@ public String getDisplayName() {
 
     /**
      * Returns the thread that is executing this step.
-     * Needs to take {@link CpsThreadGroup} as a parameter to prove that the caller is in CpsVmThread.
```
dwnusbaum: Just having a `CpsThreadGroup` is not enough to prove that the caller is on the CPS VM thread as `CpsStepContext.getThreadGroupSynchronously` appears to work fine from any thread.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -200,8 +204,29 @@ private void assertVmThread() {
...
+            LOGGER.log(Level.FINE, "no thread " + id + " among " + threads.keySet(), new IllegalStateException());
+        }
+        return thread;
+    }
+
+    /**
+     * Returns an unmodifiable view of all threads in the thread group.
+     */
+    @CpsVmThreadOnly
+    public Collection<CpsThread> getThreads() {
```
dwnusbaum: I introduced this method to be able to make `threads` private, and to make it easy to assert that all code iterating over threads is doing so in the CPS VM thread and so doesn't need any additional synchronization.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
```diff
@@ -128,7 +128,7 @@ public static CpsThreadDump from(Throwable t) {
     public static CpsThreadDump from(CpsThreadGroup g) {
         // all the threads that share the same head form a logically single thread
         Map<FlowHead, List<CpsThread>> m = new LinkedHashMap<FlowHead,List<CpsThread>>();
-        for (CpsThread t : g.threads.values()) {
+        for (CpsThread t : g.getThreads()) {
```
dwnusbaum: Looks like this is called off of the CPS VM thread in some cases, see `CpsThreadDumpTest.simple`: 

```
java.lang.AssertionError
at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.assertVmThread(CpsThreadGroup.java:204)
at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.getThreads(CpsThreadGroup.java:228)
at org.jenkinsci.plugins.workflow.cps.CpsThreadDump.from(CpsThreadDump.java:131)
at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.getThreadDump(CpsThreadGroup.java:466)
at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.getThreadDump(CpsFlowExecution.java:1056)
at org.jenkinsci.plugins.workflow.cps.CpsThreadDumpAction.threadDumpSynchronous(CpsThreadDumpAction.java:60)
at org.jenkinsci.plugins.workflow.cps.CpsThreadDumpTest.simple(CpsThreadDumpTest.java:51)
```
I'll investigate more on Monday. I guess we'll need to add some synchronization around this call for safety in case of concurrent modification and make `CpsThreadGroup.getThreads` not `@CpsVmThreadOnly`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
wdforson: I'd go with `Iterable` by default (i.e. unless some specialization in `Collection` is needed -- this usually being `Collection::size`).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -200,8 +204,29 @@ private void assertVmThread() {
...
+     * Normally called from the CPS VM thread, but may be called from other threads via {@link CpsStepContext#doGet}.
+     *
+     * @return
+     *      null if the thread has finished executing.
+     */
     public CpsThread getThread(int id) {
-        return threads.get(id);
+        CpsThread thread = threads.get(id);
+        if (thread == null && LOGGER.isLoggable(Level.FINE)) {
+            LOGGER.log(Level.FINE, "no thread " + id + " among " + threads.keySet(), new IllegalStateException());
```
wdforson: this is perhaps a little nitpicky, but...while this code is being touched, may as well disambiguate the error logging. that is, in the current situation, there's no way to distinguish between the cases where `id` is present in the map, but is mapped to a `null` value, and the case where `id` is simply absent.

two approaches come to mind:

a) use `Map::getOrDefault` with some kinda dummy default (...which would require _defining_ some dummy instance of `CpsThread`, which is likely overkill for this use case)
b) log `threads.entrySet()` rather than `threads.keySet()`
c) change the value type from `CpsThread` to `Optional<CpsThread>` :trollface: 

(also, I'd stress again that I am totally ignorant about this codebase, so if the possibility of a `null` value being inserted into the threads map is sufficiently remote as to be n/a, then never mind)

dwnusbaum: Good point, but I think since we are using a standard `TreeMap` with natural ordering of keys, `null` is never permitted as a value and trying to insert `null` would throw an exception, see [TreeMap.put docs](https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html#put-K-V-).

wdforson: ah, true. since `threads` is constructed without a comparator, it would use natural ordering, and thus disallow null.

in other words, _with that context_, the current logging is not ambiguous. if, as shown in some [sample output in the ticket](https://issues.jenkins-ci.org/browse/JENKINS-56890?focusedCommentId=364471&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-364471), `threads.get($id)` returns `null`, yet `$id` is present in the key set that is subsequently logged, then you know that there is a concurrency bug.

wdforson: that said, I would personally still log `threads.entrySet()`, unless there is some good reason not to (which seems quite unlikely).

dwnusbaum: > I would personally still log threads.entrySet()

We could, but `CpsThread.toString` returns output like `Thread #id @hashcode`, so it probably wouldn't be particularly useful since the Map's keys are the ids.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
dwnusbaum: I went with `Iterable` in cb6b2b2 since all existing uses are just for-each loops.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
wdforson: my reasoning in favor of `entrySet()` is that it would eliminate the potential _appearance_ of ambiguity vis-a-vis the possibility of a `null` value for anyone looking at the logging output without context.

for example, here is an excerpt from the comment I linked above:

> In both cases, the thread IDs being looked up do exist in the `threads` TreeSet, but their value is null? But I don't see anywhere where a null value gets inserted into that TreeSet. Could this be a race condition, accessing that value as it's being added/removed by another thread?

if `threads.entrySet()` were logged, this confusion wouldn't happen (even if the `toString` implementation of `CpsThread` is super minimal).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
jglick: The test deliberately bypasses the UI (see note in `threadDumpSynchronous`) so that is not meaningful. At any rate, we do want the UI to display _something_ even if the CPS VM thread is currently blocked on something. (See also [JENKINS-31839](https://issues.jenkins-ci.org/browse/JENKINS-31839).)

jglick:
> I was not able to reproduce the reported issue in a test

Might have to run the step a bunch of times in a loop or something, who knows.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
jglick: Better to just use annotations/assertions to verify that null values could never be added.

dwnusbaum:
> Might have to run the step a bunch of times in a loop or something, who knows.

Yeah probably. I tried combinations of loops, parallels, and sequences of synchronous and asynchronous steps in relation to the step using `GeneralizedNonBlockingExecution`, but was only able to reproduce [JENKINS-50474](https://issues.jenkins-ci.org/browse/JENKINS-50474?focusedCommentId=366503&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-366503).

dwnusbaum:
@jglick Any thoughts/preference on using a `ConcurrentSkipListMap` instead of a synchronized `TreeMap`? That would allow us to avoid explicit synchronization, and I think a weakly-consistent snapshot iterator is what we want in the case of `CpsThreadDump` anyway.

jglick:
> Any thoughts/preference

Sounds OK.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
dwnusbaum: Now that we are using a `ConcurrentSkipListMap` we shouldn't have any issues even if the VM thread is blocked on a read/write for some reason.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -208,7 +208,7 @@ private Continuable createContinuable(CpsThread currentThread, CpsCallableInvoca
                     List<StepExecution> executions = new ArrayList<>();
                     // cf. trick in CpsFlowExecution.getCurrentExecutions(true)
                     Map<FlowHead, CpsThread> m = new LinkedHashMap<>();
-                    for (CpsThread t : g.threads.values()) {
+                    for (CpsThread t : g.getThreads()) {
```
jglick: Note that this is now no behavioral change, you are just encapsulating the field.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -215,17 +215,12 @@ public String getDisplayName() {
...
      * @return
      *      null if the thread has finished executing.
      */
     @CheckForNull CpsThread getThread(CpsThreadGroup g) {
-        CpsThread thread = g.threads.get(threadId);
-        if (thread == null) {
-            LOGGER.log(Level.FINE, "no thread " + threadId + " among " + g.threads.keySet(), new IllegalStateException());
-        }
-        return thread;
+        return g.getThread(threadId);
```
jglick: Could just inline this method now I guess. No reason for it to exist as an instance method.

dwnusbaum: Yeah, I looked into removing the method earlier, but ran into a [usage in `CpsBodyInvoker`](https://github.com/jenkinsci/workflow-cps-plugin/blob/e5e5f8054ff485685e0a5b36db65d549604ed956/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyInvoker.java#L151) that would require making `CpsStepContext.threadId` accessible and so it seemed like a wash.

jglick: Ah, sorry, missed that this was encapsulating a `private` variable in this class too.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
dwnusbaum: Yeah, to me it still seems valuable to encapsulate the field and only provide a read only view of the values since that is what was being used everywhere anyway, but I don't feel too strongly about it. I am happy to revert these changes to keep the diff minimal if you prefer.

jglick: The added encapsulation seems wise.

PR #293, opened 2019-06-06
Add pipeline source: HTTP

julienduchesne:
Allows fetching a Jenkinsfile from a given HTTP URL
Basic auth credentials can be set
Usecases:
- SCM Host APIs provide full files (faster than checkout)
- Amazon S3 (or other cloud file storage providers)

image

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright 2019 CloudBees, Inc.
+ * Copyright 2019 Coveo Solutions Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -32,6 +32,7 @@ import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.URIRequirementBuilder;
 import hudson.Extension;
 import hudson.FilePath;
+import hudson.Util;
 import hudson.model.*;
 import hudson.scm.SCM;
 import hudson.scm.SCMDescriptor;
@@ -43,10 +44,7 @@ import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.flow.*;
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.Stapler;
-import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.*;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -64,17 +62,12 @@ import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.
 @PersistIn(JOB) public class CpsHttpFlowDefinition extends FlowDefinition {
 
     private final String scriptUrl;
-    private final String credentialsId;
     private final int retryCount;
+    private String credentialsId;
 
-    @DataBoundConstructor public CpsHttpFlowDefinition(String scriptUrl, String credentialsId, int retryCount) {
+    @DataBoundConstructor public CpsHttpFlowDefinition(String scriptUrl, int retryCount) {
         this.scriptUrl = scriptUrl.trim();
-        this.credentialsId = credentialsId;
         this.retryCount = retryCount;
-    }
-
-    public String getCredentialsId() {
-        return credentialsId;
     }
 
     public String getScriptUrl() {
@@ -83,6 +76,14 @@ import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.
 
     public int getRetryCount() {
         return retryCount;
+    }
+
+    @DataBoundSetter public void setCredentialsId(String credentialsId) {
+        this.credentialsId = Util.fixEmpty(credentialsId);
+    }
+
+    public String getCredentialsId() {
+        return credentialsId;
     }
 
     @Override public CpsFlowExecution create(FlowExecutionOwner owner, TaskListener listener, List<? extends Action> actions)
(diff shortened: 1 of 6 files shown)


julienduchesne:
Our personal use case is that we pull the same repo from bitbucket hundreds of times on the same Jenkins instance. I think we are starting to hit a limit with all those concurrent checkouts. We want to switch to using the Bitbucket API instead

julienduchesne:
@jenkinsci/code-reviewers. What do you think?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2019 CloudBees, Inc.
```
oleg-nenashev: Needs to be changed if you do not work for CloudBees

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
```diff
@@ -0,0 +1,158 @@
...
+
+        String expandedScriptUrl = build.getEnvironment(listener).expand(scriptUrl);
+        URL url = new URL(expandedScriptUrl);
+        listener.getLogger().println("Fetching pipeline from " + expandedScriptUrl);
+        int count = 0;
+        int maxTries = retryCount + 1;
+
+        while (true) {
+            try {
+                HttpURLConnection con = (HttpURLConnection) url.openConnection();
```
oleg-nenashev: https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin/blob/master/src/main/java/io/jenkins/plugins/httpclient/RobustHTTPClient.java could be used for better stability

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
julienduchesne: I just write my name or nothing at all?

oleg-nenashev: Your name, your company name, or whatever "Plugin contributors" if you prefer.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinitionTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class CpsHttpFlowDefinitionTest {
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Test public void testRunJenkinsHomePageAsPipeline() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        CpsHttpFlowDefinition def = new CpsHttpFlowDefinition(r.jenkins.getRootUrl(), "", 3);
+        p.setDefinition(def);
+        WorkflowRun b = r.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
+        r.assertLogContains("unexpected token", b);
```
jglick: Should be at least one _positive_ test! You can use the `userContent` system for example.

src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition/help-scriptUrl.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    Full URL to your script
```
jglick:
```suggestion
    Full URL to your script.
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
```diff
@@ -0,0 +1,158 @@
...
+
+@PersistIn(JOB) public class CpsHttpFlowDefinition extends FlowDefinition {
+
+    private final String scriptUrl;
+    private final String credentialsId;
+    private final int retryCount;
+
+    @DataBoundConstructor public CpsHttpFlowDefinition(String scriptUrl, String credentialsId, int retryCount) {
+        this.scriptUrl = scriptUrl.trim();
+        this.credentialsId = credentialsId;
```
jglick:
```suggestion
        this.credentialsId = Util.fixEmpty(credentialsId);
```

and later you can simply check for null.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
```diff
@@ -0,0 +1,158 @@
...
+
+import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.JOB;
+
+@PersistIn(JOB) public class CpsHttpFlowDefinition extends FlowDefinition {
+
+    private final String scriptUrl;
+    private final String credentialsId;
+    private final int retryCount;
+
+    @DataBoundConstructor public CpsHttpFlowDefinition(String scriptUrl, String credentialsId, int retryCount) {
```
jglick: Generally speaking, optional parameters should be moved to a `@DataBoundSetter`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
jglick: And easier authentication probably.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
```diff
@@ -0,0 +1,158 @@
...
+        int count = 0;
+        int maxTries = retryCount + 1;
+
+        while (true) {
+            try {
+                HttpURLConnection con = (HttpURLConnection) url.openConnection();
+                con.setRequestMethod("GET");
+
+                if (!StringUtils.isBlank(credentialsId)) {
+                    UsernamePasswordCredentials credentials = (UsernamePasswordCredentials) CredentialsMatchers.firstOrNull(CredentialsProvider.lookupCredentials(UsernamePasswordCredentials.class, Jenkins.getInstance(), ACL.SYSTEM, Collections.emptyList()), CredentialsMatchers.withId(credentialsId));
```
jglick: IIRC there is some additional API that should be called to record that this build is using these credentials.

jglick:
> we are starting to hit a limit with all those concurrent checkouts. We want to switch to using the Bitbucket API instead

So that would be better handled by just selecting the `CpsScmFileSystem.lightweight` option, assuming the BitBucket SCM plugin implements `SCMFileSystem` (if not, fix it to do so).

jglick:
or [this way](https://github.com/AmadeusITGroup/workflow-cps-global-lib-http-plugin#directly-in-the-jenkinsfile)

julienduchesne:
> Is this a sufficiently common need that it deserves to be in `workflow-cps`? Most people would use the SCM script source. Could be done as a separate plugin with an independent maintainer I think.

I think it's generic enough to be there but that is my opinion. If you won't accept, it will be a plugin of course

> > we are starting to hit a limit with all those concurrent checkouts. We want to switch to using the Bitbucket API instead
> 
> So that would be better handled by just selecting the `CpsScmFileSystem.lightweight` option, assuming the BitBucket SCM plugin implements `SCMFileSystem` (if not, fix it to do so).

We are using interpolation in our branch name which the lightweight doesn't support. Anyhow, HTTP is much much faster.


> or [this way](https://github.com/AmadeusITGroup/workflow-cps-global-lib-http-plugin#directly-in-the-jenkinsfile)

Libraries cannot be used as a pipeline entrypoint. You need an inital Jenkinsfile to pull a library

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsHttpFlowDefinition.java
julienduchesne: I haven't found any usage of this anywhere. Do you have any examples?

jglick: [query](https://github.com/search?q=user%3Ajenkinsci+RobustHTTPClient&type=Code)

julienduchesne:
Alright. The changes have been made. So, what do you guys think about whether or not this belongs in this repo?

daniel-beck:
Can easily be done as a separate plugin, that's how I did it with https://plugins.jenkins.io/inline-pipeline

julienduchesne:
Of course, I'm just trying to understand at what point something should be integrated or be made as a separate plugin. Does it make sense to create a new plugin to host a single 100-line code file when that code is not any less generic than the SCM implementation?

Also, any pointers on how to add a plugin to the `jenkinsci` org in github? If I split it, I'd like it to be an "official" plugin

julienduchesne:
Alright. I migrated the plugin here: https://github.com/coveo/workflow-cps-http-plugin

dwnusbaum:
> Also, any pointers on how to add a plugin to the jenkinsci org in github?

@julienduchesne see https://jenkins.io/doc/developer/publishing/requesting-hosting/. Note also that someone recently created https://github.com/jenkinsci/workflow-cps-global-lib-http-plugin, maybe it makes sense for these features to live in the same plugin.

PR #296, opened 2019-06-14
Remove gmaven and groovy test

bitwiseman:
Experimental WIP - JENKINS-58028


bitwiseman:
@dwnusbaum 
Fair point.  I didn't analyze this beyond drop in replacement of existing functionality.

bitwiseman:
@dwnusbaum  Done.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -0,0 +1,43 @@
...
+    public void simplestPossibleTest() throws Exception {
+        CpsFlowDefinition flow = new CpsFlowDefinition(
+                "def sqrt(int x) { return Math.sqrt(x); }\nfor (int i=0; i<10; i++)\nsqrt(i);",
+                false);
+
+        createExecution(flow);
+        exec.start();
+        exec.waitForSuspension();
+
+        assert exec.isComplete();
```
basil: Per [Error Prone's `UseCorrectAssertInTests` check](https://errorprone.info/bugpattern/UseCorrectAssertInTests):

> Java assert statements may or may not be evaluated depending on runtime flags to the JVM invocation. When used in tests, this means that the test assertions may not be checked, and a test may pass when it should actually fail. To avoid this, use one of the assertion libraries that are always enabled, such as JUnit’s `org.junit.Assert` or Google’s Truth library.

In this case, I'd suggest adding this import …

```java
import static org.junit.Assert.assertTrue;
```

… and then changing this line to:

```java
assertTrue(exec.isComplete());
```

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -0,0 +1,43 @@
...
+                "throw new Throwable('This is a fire drill, not a real fire');",
+                false);
+
+        // get this going...
+        createExecution(flow);
+        exec.start();
+
+        // it should stop at watch and suspend.
+        exec.waitForSuspension();
+        assert exec.isComplete();
```
basil: For the reasons explained above, I'd suggest changing this to:

```java
assertTrue(exec.isComplete());
```

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -0,0 +1,43 @@
...
+        Throwable t = exec.getCauseOfFailure();
+        assert t.getClass().equals(Throwable.class);
```
basil: For the reasons explained above, In this case, I'd suggest adding this import …

```java
import static org.junit.Assert.assertEquals;
```

… and then changing these two lines to:

```java
assertEquals(Throwable.class, t.getClass());
assertEquals("This is a fire drill, not a real fire", t.getMessage());
```

bitwiseman: @basil 
Thanks for this feedback.  FYI, you can suggest changes (#6 on https://help.github.com/en/articles/reviewing-proposed-changes-in-a-pull-request) to get your intent across quickly and make it easier for other to take your change.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -0,0 +1,46 @@
...
+ * @author Kohsuke Kawaguchi
+ */
+public class CpsFlowDefinitionTest extends AbstractCpsFlowTest {
+    /**
+     * CpsFlowDefinition's simplest possible test.
+     */
+    @Test
+    public void simplestPossibleTest() throws Exception {
+        CpsFlowDefinition flow = new CpsFlowDefinition(
+                "def sqrt(int x) { return Math.sqrt(x); }\nfor (int i=0; i<10; i++)\nsqrt(i);",
```
dwnusbaum: nit: I personally find it a bit more readable to go ahead and split the string if the scripts requires a newline, something like this:
```suggestion
                "def sqrt(int x) { return Math.sqrt(x); }\n" + 
                "for (int i=0; i<10; i++) { sqrt(i); }",
```

PR #297, opened 2019-06-17
Reproduce and fix JENKINS-56682

dwnusbaum:
See [JENKINS-56682](https://issues.jenkins-ci.org/browse/JENKINS-56682), https://github.com/jenkinsci/script-security-plugin/pull/259, and https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/71.

The PR to script-security does not affect Pipeline, because workflow-cps never calls `GroovySandbox.runScript`. `CpsGroovyShell` is already using a copy of `CleanGroovyClassloader`, which seems to be necessary for the fix as discussed in https://github.com/jenkinsci/script-security-plugin/pull/259#discussion_r290874101, so we just need to make sure that we add a `ClassLoaderWhitelist` for the shell's class loader when parsing scripts. `SandboxContinuable` also interacts with `GroovySandbox`, but it already specifies a whitelist that includes a class loader whitelist for `shell.getClassLoader()`, so we don't need to change it at all.

Marking as work-in-progress since it looks like this isn't enough to make the test in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/71 pass.


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -590,4 +590,36 @@ public void blockConstructorInvocationAtRuntime() throws Exception {
...
+    @Issue("JENKINS-56682")
+    @Test
+    public void scriptInitializersAtFieldSyntax() throws Exception {
+        WorkflowJob p = jenkins.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(
+                "import groovy.transform.Field\n" +
+                "@Field static int foo = 1\n" +
+                "@Field int bar = foo + 1\n" +
+                "@Field int baz = bar + 1\n" +
+                "echo \"baz is ${baz}\"", true));
```
jglick: Arguably more readable:

```suggestion
                "echo(/baz is $baz/)", true));
```

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -590,4 +590,36 @@ public void blockConstructorInvocationAtRuntime() throws Exception {
...
+    public void scriptInitializersClassSyntax() throws Exception {
+        WorkflowJob p = jenkins.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(
+                "import org.codehaus.groovy.runtime.InvokerHelper\n" +
+                "class MyScript extends org.jenkinsci.plugins.workflow.cps.CpsScript {\n" +
+                "  { MyScript.foo++ }\n" + // The instance initializer seems to be context sensitive, if placed below the field it is treated as a closure...
+                "  static { MyScript.foo++ }\n" +
+                "  static int foo = 0\n" +
+                "  def run() {\n" +
+                "    echo \"MyScript.foo is ${MyScript.foo}\"\n " +
```
jglick:
```suggestion
                "    echo(/MyScript.foo is $MyScript.foo/)\n " +
```

dwnusbaum: Done in 4b3b3e4.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
dwnusbaum: Done in 4b3b3e4.

dwnusbaum:
Ok, I think https://github.com/jenkinsci/workflow-cps-plugin/pull/297/commits/0938337e4782bf98084d1d2d2e65e9ff003c64a7 should fix the issue in workflow-cps-global-lib. The problem was that we need to whitelist the trusted class loader as well.

I am a little surprised that all tests passed when using `Whitelist.all()` instead of `CpsWhitelist` in `CpsGroovyShell.doParse`, since `CpsWhitelist` isn't marked with `@Extension` I don't think it would available as part of `Whitelist.all()`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
```diff
@@ -127,7 +127,15 @@ public Script parse(GroovyCodeSource codeSource) throws CompilationFailedExcepti
     }
 
     private Script doParse(GroovyCodeSource codeSource) throws CompilationFailedException {
-        try (GroovySandbox.Scope scope = new GroovySandbox().enter()) {
+        GroovySandbox sandbox = new GroovySandbox();
+        if (execution != null) {
+            sandbox.withWhitelist(new GroovyClassLoaderWhitelist(CpsWhitelist.get(),
```
dwnusbaum: See https://github.com/jenkinsci/workflow-cps-plugin/pull/297#issuecomment-502818187 for discussion of the switch from the implicit `Whitelist.all()` to `CpsWhitelist.get()`. TL;DR: I think it doesn't really matter, the things that would be affected are broken either way.

dwnusbaum: Ok, so parsing happens after the CPS transformation, but doesn't run in the CPS VM thread, so calling any CPS-transformed method during script parsing as in this example will fail: 

```
    @Issue("JENKINS-56682")
    @Test
    public void scriptInitializersClassSyntaxWithMethodCall() throws Exception {
        WorkflowJob p = jenkins.jenkins.createProject(WorkflowJob.class, "p");
                "class MyScript extends org.jenkinsci.plugins.workflow.cps.CpsScript {\n" +
                "  static { bar() }\n" +
                "  static int foo = 0\n" +
                "  static def bar() { MyScript.foo = MyScript.foo + 1 }\n" +
                "  def run() {\n" +
                "    echo(/MyScript.foo is ${MyScript.foo}/)\n " +
                "  }\n" +
                "}\n", true));
        WorkflowRun b = jenkins.buildAndAssertSuccess(p);
        jenkins.assertLogContains("MyScript.foo is 1", b);
    }
```
No matter the choice of whitelist, this fails with:

> CpsCallableInvocation{methodName=bar, call=com.cloudbees.groovy.cps.impl.CpsFunction@20907d04, receiver=class MyScript, arguments=[]}

I am going to switch back to `Whitelist.all()` since I don't think there is any case where `CpsWhitelist.get()` makes sense during script parsing.

dwnusbaum: Done in b3aebc7.

jglick: Anyway based on https://github.com/jenkinsci/workflow-cps-plugin/blob/7eab5dc0c39289a372a05904d65596c4192c4005/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsWhitelist.java#L44-L48 I suspect that use of `CpsWhitelist` here would have introduced vulnerabilities.

PR #298, opened 2019-06-20
Add Java 11 build

bitwiseman:
@jenkinsci/java11-support 
Let's see what happens.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.cps;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
 import hudson.Extension;
 import hudson.FilePath;
@@ -35,7 +36,6 @@ import hudson.model.Job;
 import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.model.Run;
-import hudson.model.Run.RunnerAbortedException;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItem;
 import hudson.scm.SCM;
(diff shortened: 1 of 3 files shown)


bitwiseman:
Aw, I can't do this because I'm untrusted.  Still only building java8.

dwnusbaum:
> Aw, I can't do this because I'm untrusted. Still only building java8.

You have [release permission](https://github.com/jenkins-infra/repository-permissions-updater/blob/acf08242fbff011301ac6e8a17217999df7690f2/permissions/plugin-workflow-cps.yml#L14), just need to track someone down with voice on IRC to get GitHub permission as well.

bitwiseman:
Hm, I wonder why I didn't see those locally.

bitwiseman:
@dwnusbaum Success!

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -92,6 +92,11 @@ public boolean isLightweight() {
         this.lightweight = lightweight;
     }
 
+    @SuppressFBWarnings(
+            value = {"NP_LOAD_OF_KNOWN_NULL_VALUE", "RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE",
+                    "RCN_REDUNDANT_NULLCHECK_OF_NULL_VALUE", "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE"},
+            justification = "false positives for try-resource in java 11"
```
dwnusbaum: Ah, this one is super annoying.

jglick: This looks familiar; do you have a link to the SpotBugs issue offhand?

dwnusbaum: https://github.com/spotbugs/spotbugs/issues/756

PR #299, opened 2019-06-26
[JENKINS-58084] Call StepListeners before executing steps

abayer:
[JENKINS-58084](https://issues.jenkins-ci.org/browse/JENKINS-58084)

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/96

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.36-rc912.28a8040f3a6d</version>
+            <version>2.36-rc914.89ff1dcb2f3c</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.36-rc914.89ff1dcb2f3c</version>
+            <version>2.36</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.73</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.138.4</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -264,6 +266,12 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
 
             // Persist the node - block start and end nodes do their own persistence.
             CpsFlowExecution.maybeAutoPersistNode(an);
+
+            // Call any registered StepListeners.
+            for (StepListener sl : ExtensionList.lookup(StepListener.class)) {
+                sl.newStep(s, context);
```
dwnusbaum: If a `StepListener` extension is broken for some reason, then just by throwing an exception it could stop all steps from executing. We catch `Throwable` around calls to `GraphListener` and log failures, but if we intend for the listeners to throw exceptions to stop the step, that wouldn't work here. 

Maybe it would be better to have `StepListener` implementations call `context.onFailure` when they want to abort a step, rather than throwing an exception directly out of `StepListener.newStep`, so that we could distinguish "this listener is broken" from "this listener wants to stop the step from executing". You'd need to tweak the code here to look at `CpsStepContext.isCompleted` before calling `s.start(context)` in that case, and I'm not sure whether we'd want to run all listeners or not once one failed the step.

Maybe also worth documenting upstream that you should not call `StepContext.onSuccess` or `StepContext.newBodyInvoker` (or `StepContext.onFailure` with the current code) from a listener, or wrapping the context here to make those methods throw exceptions?

One other minor note is that `Extension.ordinal` seems somewhat relevant for this extension, since any listener that just wants to collect data/metrics about steps (if that is a valid use case) but probably wants consistency in whether they run before or after extensions that might fail steps. If `CpsStepContext.onFailure` were used, then maybe this wouldn't matter if we always ran all listeners, although then we'd have the possibility that more than one listener might try to fail a step which would cause some "already completed" warnings in `CpsStepContext.completed`.

abayer: Oh, duh, `context.onFailure` is the right way to do it. I'll rework this.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -269,9 +271,23 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
...
             CpsFlowExecution.maybeAutoPersistNode(an);
-            StepExecution e = s.start(context);
-            thread.setStep(e);
-            sync = e.start();
+
+            // Call any registered StepListeners.
+            for (StepListener sl : ExtensionList.lookup(StepListener.class)) {
+                try {
+                    sl.notifyOfNewStep(s, context);
+                } catch (Exception e) {
```
dwnusbaum: Seems like a good idea to be extra careful like with `GraphListener` methods and catch `Throwable`.

```suggestion
                } catch (Throwable e) {
```

pom.xml
```diff
@@ -84,7 +84,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.32</version>
+            <version>2.36-rc914.89ff1dcb2f3c</version>
```
dwnusbaum:
```suggestion
            <version>2.36</version>
```

PR #303, opened 2019-07-15
[JENKINS-30572] Support ParallelStep in gdsl.groovy and fix parameter types for some steps

dwnusbaum:
See [JENKINS-30572](https://issues.jenkins-ci.org/browse/JENKINS-30572).

Fixes syntax highlighting for `parallel` when using `gdsl.groovy` with IntelliJ (tested manually). Does not change the fact that the snippet generator throws an exception for `parallel` or that doc generation for `parallel` fails [here](https://jenkins.io/doc/pipeline/steps/workflow-cps/). Also fixes the GDSL for parameters of type `List`, which were being reported as type `Map`.

I did some experimentation, and you can get rid of the special-casing in gdsl.groovy, get the snippet generator to return an empty step rather than throwing an error, and make doc generation work correctly by adding `@DataBoundConstructor` to `ParallelStep`'s constructor, adding public getters for `closures` and `failFast`, and then using [this branch of structs](https://github.com/dwnusbaum/structs-plugin/compare/master...dwnusbaum:JENKINS-28510) that adds support for parameters of type Map. Since `Closure` can't be data bound, I don't think it really makes sense to do those things (and the branch of structs isn't very useful independently without support for Map parameters in core/Stapler), but it seems worth mentioning. If you only add `@DataBoundConstructor` and the getters, the snippet generator works but doc generation fails because of an exception in `ParameterType.java` in structs: `UnsupportedOperationException: do not know how to categorize attributes of type java.util.Map<java.lang.String, groovy.lang.Closure>`.

Here is a diff of the resulting GDSL before and after this change: 

```diff
3a4,5
> method(name: 'parallel', type: 'Object', params: ['closures':'java.util.Map'], doc: 'Execute in parallel')
> method(name: 'parallel', type: 'Object', namedParams: [parameter(name: 'closures', type: 'java.util.Map'), parameter(name: 'failFast', type: 'boolean'), ], doc: 'Execute in parallel')
5c7
< method(name: 'build', type: 'Object', namedParams: [parameter(name: 'job', type: 'java.lang.String'), parameter(name: 'parameters', type: 'Map'), parameter(name: 'propagate', type: 'boolean'), parameter(name: 'quietPeriod', type: 'java.lang.Integer'), parameter(name: 'wait', type: 'boolean'), ], doc: 'Build a job')
---
> method(name: 'build', type: 'Object', namedParams: [parameter(name: 'job', type: 'java.lang.String'), parameter(name: 'parameters', type: 'java.util.List'), parameter(name: 'propagate', type: 'boolean'), parameter(name: 'quietPeriod', type: 'java.lang.Integer'), parameter(name: 'wait', type: 'boolean'), ], doc: 'Build a job')
9c11
< method(name: 'input', type: 'Object', namedParams: [parameter(name: 'message', type: 'java.lang.String'), parameter(name: 'id', type: 'java.lang.String'), parameter(name: 'ok', type: 'java.lang.String'), parameter(name: 'parameters', type: 'Map'), parameter(name: 'submitter', type: 'java.lang.String'), parameter(name: 'submitterParameter', type: 'java.lang.String'), ], doc: 'Wait for interactive input')
---
> method(name: 'input', type: 'Object', namedParams: [parameter(name: 'message', type: 'java.lang.String'), parameter(name: 'id', type: 'java.lang.String'), parameter(name: 'ok', type: 'java.lang.String'), parameter(name: 'parameters', type: 'java.util.List'), parameter(name: 'submitter', type: 'java.lang.String'), parameter(name: 'submitterParameter', type: 'java.lang.String'), ], doc: 'Wait for interactive input')
17c19
< method(name: 'properties', type: 'Object', params: [properties:'Map'], doc: 'Set job properties')
---
> method(name: 'properties', type: 'Object', params: [properties:'java.util.List'], doc: 'Set job properties')
19c21
< method(name: 'resolveScm', type: 'Object', namedParams: [parameter(name: 'source', type: 'Map'), parameter(name: 'targets', type: 'Map'), parameter(name: 'ignoreErrors', type: 'boolean'), ], doc: 'Resolves an SCM from an SCM Source and a list of candidate target branch names')
---
> method(name: 'resolveScm', type: 'Object', namedParams: [parameter(name: 'source', type: 'Map'), parameter(name: 'targets', type: 'java.util.List'), parameter(name: 'ignoreErrors', type: 'boolean'), ], doc: 'Resolves an SCM from an SCM Source and a list of candidate target branch names')
31,32c33,34
< method(name: 'withCredentials', type: 'Object', params: [bindings:Map, body:'Closure'], doc: 'Bind credentials to variables')
< method(name: 'withEnv', type: 'Object', params: [overrides:Map, body:'Closure'], doc: 'Set environment variables')
---
> method(name: 'withCredentials', type: 'Object', params: [bindings:java.util.List, body:'Closure'], doc: 'Bind credentials to variables')
> method(name: 'withEnv', type: 'Object', params: [overrides:java.util.List, body:'Closure'], doc: 'Set environment variables')
79c81
< method(name: 'dockerFingerprintFrom', type: 'Object', namedParams: [parameter(name: 'dockerfile', type: 'java.lang.String'), parameter(name: 'image', type: 'java.lang.String'), parameter(name: 'buildArgs', type: 'Map'), parameter(name: 'toolName', type: 'java.lang.String'), ], doc: 'Record trace of a Docker image used in FROM')
---
> method(name: 'dockerFingerprintFrom', type: 'Object', namedParams: [parameter(name: 'dockerfile', type: 'java.lang.String'), parameter(name: 'image', type: 'java.lang.String'), parameter(name: 'buildArgs', type: 'java.util.Map'), parameter(name: 'toolName', type: 'java.lang.String'), ], doc: 'Record trace of a Docker image used in FROM')
81c83
< method(name: 'unarchive', type: 'Object', namedParams: [parameter(name: 'mapping', type: 'Map'), ], doc: 'Copy archived artifacts into the workspace')
---
> method(name: 'unarchive', type: 'Object', namedParams: [parameter(name: 'mapping', type: 'java.util.Map'), ], doc: 'Copy archived artifacts into the workspace')
91d92
< // class org.jenkinsci.plugins.workflow.cps.steps.ParallelStep: There's no @DataBoundConstructor on any constructor of class org.jenkinsci.plugins.workflow.cps.steps.ParallelStep
\ No newline at end of file
```

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/gdsl.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/gdsl.groovy
@@ -6,6 +6,7 @@ import hudson.Functions
 import org.jenkinsci.plugins.structs.describable.DescribableModel
 import org.jenkinsci.plugins.workflow.cps.GlobalVariable
 import org.jenkinsci.plugins.workflow.cps.Snippetizer
+import org.jenkinsci.plugins.workflow.cps.steps.ParallelStep
 import org.jenkinsci.plugins.workflow.steps.Step
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor
 
@@ -18,6 +19,8 @@ Snippetizer snippetizer = my;
 Map<StepDescriptor,DescribableModel> steps = new LinkedHashMap<StepDescriptor, DescribableModel>();
 
 def errorSteps = [:]
+def nodeContext = []
+def scriptContext = []
 
 [false, true].each { isAdvanced ->
     snippetizer.getQuasiDescriptors(isAdvanced).each { d ->
@@ -25,6 +28,12 @@ def errorSteps = [:]
             return // TODO JENKINS-37215
         }
         StepDescriptor step = Jenkins.get().getDescriptor(d.real.clazz)
+        if (step instanceof ParallelStep.DescriptorImpl) {
+            // ParallelStep does not support data binding, so we just hard-code the correct entries.
+            scriptContext.add("method(name: 'parallel', type: 'Object', params: ['closures':'java.util.Map'], doc: 'Execute in parallel')")
+            scriptContext.add("method(name: 'parallel', type: 'Object', namedParams: [parameter(name: 'closures', type: 'java.util.Map'), parameter(name: 'failFast', type: 'boolean'), ], doc: 'Execute in parallel')")
+            return
+        }
         DescribableModel schema
         try {
             schema = new DescribableModel(d.real.clazz)
@@ -38,9 +47,6 @@ def errorSteps = [:]
 
     }
 }
-
-def nodeContext = []
-def scriptContext = []
 
 steps.each { StepDescriptor step, DescribableModel model ->
     def params = [:], opts = [:]
@@ -128,13 +134,17 @@ ${errorSteps.collect { k, v ->
 }.join("\n")}
 """)
 
-// Render anything other than a primitive, Closure, or a java.* class as a Map.
+// Render anything other than a primitive, Closure, or a java.* class or interface as a Map.
 def objectTypeToMap(String type) {
-    if (type != null && type.contains(".") && !(type.startsWith("java"))) {
-        return "Map"
-    } else {
-        return type
+    if (type != null && type.contains(".")) {
+        if (type.startsWith("interface ")) {
+            type = type.substring("interface ".length());
+        }
+        if (!type.startsWith("java")) {
+            return "Map"
+        }
     }
+    return type
 }
 
 def fetchActualClass(Type type) {


dwnusbaum:
> I would suggest checking whether you can use CustomDescribableModel here.

I don't think `CustomDescribableModel` is useful here. The exception is thrown from `Descriptor.newInstance(StaplerRequest, JSONObject)` in the case of the snippet generator, and `DescribableModel.of` for doc generation. Right now `CustomDescribableModel` only matters for `DescribableModel.instantiate` and `DescribableModel.uninstantiate2`: 

Exception from the snippet generator:

```
org.kohsuke.stapler.NoStaplerConstructorException: There's no @DataBoundConstructor on any constructor of class org.jenkinsci.plugins.workflow.cps.steps.ParallelStep
    at org.kohsuke.stapler.ClassDescriptor.loadConstructorParamNames (ClassDescriptor.java:265)
    at org.kohsuke.stapler.RequestImpl.instantiate (RequestImpl.java:764)
    at org.kohsuke.stapler.RequestImpl.access$200 (RequestImpl.java:83)
    at org.kohsuke.stapler.RequestImpl$TypePair.convertJSON (RequestImpl.java:678)
    at org.kohsuke.stapler.RequestImpl.bindJSON (RequestImpl.java:478)
    at org.kohsuke.stapler.RequestImpl.bindJSON (RequestImpl.java:474)
    at hudson.model.Descriptor.newInstance (Descriptor.java:592)
    at org.jenkinsci.plugins.workflow.cps.Snippetizer.doGenerateSnippet (Snippetizer.java:489)
```

Exception from doc generation:
```
org.kohsuke.stapler.NoStaplerConstructorException: There's no @DataBoundConstructor on any constructor of class org.jenkinsci.plugins.workflow.cps.steps.ParallelStep
  at org.kohsuke.stapler.ClassDescriptor.loadConstructorParamNames(ClassDescriptor.java:265)
  at org.jenkinsci.plugins.structs.describable.DescribableModel.<init>(DescribableModel.java:144)
```

I think the snippet generator could be fixed with a `Descriptor.newInstance` override, but I'm not sure about doc generation. Maybe possible to fix with new methods on `CustomDescribableModel` that allow the constructor and parameter types to be computed dynamically, although maybe better to just special case `ParallelStep` in the doc generator unless there are other use cases for something like that.

PR #304, opened 2019-07-15
Remove tests for BuildTriggerStep

dwnusbaum:
I am migrating these tests to pipeline-build-step via https://github.com/jenkinsci/pipeline-build-step-plugin/pull/27.


src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -229,7 +229,7 @@
     @Test public void doGdsl() throws Exception {
         JenkinsRule.WebClient wc = r.createWebClient();
         String gdsl = wc.goTo(Snippetizer.ACTION_URL + "/gdsl", "text/plain").getWebResponse().getContentAsString();
-        assertThat("Description is included as doc", gdsl, containsString("Build a job"));
+        assertThat("Description is included as doc", gdsl, containsString("Shell Script"));
```
jglick: In retrospect I should have been a little less lazy and used `@TestExtension`.

PR #305, opened 2019-07-17
[JENKINS-58501] Add whitelist for exotic cases throwing CPS mismatch warning logs 

steven-terrana:
As described in [JENKINS-58501](https://issues.jenkins-ci.org/browse/JENKINS-58501), the [Templating Engine Plugin](https://plugins.jenkins.io/templating-engine) utilizes metaprogramming techniques to control program flow. 

The introduction of logging CPS mismatches has flooded the console log with false positive warnings that come about from how the plugin executes pipeline code. 

Rather than try to handle every edge case that may throw a false positive for the CPS mismatch warning logs, I think it would be a good approach to add a ``AbstractMismatchWhitelist`` extension that allows plugins to register their own edge cases that should prevent the log from being printed. 

Tests have been added that confirm the warning log is printed for cases not caught by a registered extension of the whitelist, and that mismatches that are caught prevent the log from being printed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/AbstractMismatchWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/AbstractMismatchWhitelist.java
@@ -1,0 +1,16 @@
+
+package org.jenkinsci.plugins.workflow.cps;
+
+import hudson.ExtensionPoint;
+import hudson.ExtensionList;
+import jenkins.model.Jenkins; 
+
+public abstract class AbstractMismatchWhitelist implements ExtensionPoint {
+    public Boolean ignoreCpsMismatch(String expectedReceiver, String expectedMethodName, String actualReceiver, String actualMethodName){
+        return false; 
+    }
+
+    public static ExtensionList<AbstractMismatchWhitelist> all(){
+        return Jenkins.getActiveInstance().getExtensionList(AbstractMismatchWhitelist.class);
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -2,6 +2,7 @@ package org.jenkinsci.plugins.workflow.cps;
 
 import com.cloudbees.groovy.cps.impl.CpsCallableInvocation;
 import hudson.Main;
+import hudson.ExtensionList;
 import hudson.model.Computer;
 import hudson.remoting.SingleLaneExecutorService;
 import hudson.security.ACL;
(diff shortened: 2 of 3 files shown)


dwnusbaum:
Seems kind of heavyweight to have a whitelist for this. If there is no good way to fix the underlying issue, would it be enough to just suppress all warnings where `expectedReceiver` is an instance of `MetaClass` (and maybe a few other classes with similar uses)?

The simplest example noted in the ticket might be fixable by tweaking the code in groovy-cps to recognize reflective patterns like `MetaClass.invokeMethod(object, method)` and change what is recorded as the expected receiver and method name, but that seems messy.

jglick:
Amends #211.

steven-terrana:
@dwnusbaum - i felt the same way, but unfortunately for our plugin we've implemented a proxy layer between executing a step pipeline step and pipeline code being executed, so the false positives that get thrown for us are a little more varied than just adding MetaClass and ExpandoMetaClass. 

I would be happy to attach a build log that shows all of the warnings that are getting thrown, but my feeling was that an abstract whitelist doesn't change functionality for the majority but gives plugins that have specific false positives to be able to address this.

jglick:
Would deserve a downstream PR consuming the incremental version and demonstrating effectiveness.

I agree that introducing an API for this sounds like overkill. Better to patch `groovy-cps` to not produce the false warnings to begin with. As a starting point, I would suggest a PR to `workflow-cps-plugin` that includes an `@Ignored` test that demonstrates an improper warning in a _minimal, self-contained_ script.

jglick:
> we've implemented a proxy layer between executing a step pipeline step and pipeline code being executed

Consider whether https://github.com/jenkinsci/workflow-api-plugin/pull/96 would provide a simpler solution without Groovy tricks.

steven-terrana:
I can do that.  

The majority of our cases are when we're dynamically executing methods defined on scripts that have been parsed where the actual receiver is of the pattern "script15633701193482130252211" and groovy.lang.ExpandoMetaClass.invokeMethod

steven-terrana:
@jglick - i'll certainly take a look into the workflow-api-plugin to see if we can simplify some of the underlying implementation instead of using groovy tricks, but for the short term we are where we are. 

I will create @Ignore'd unit tests demonstrating the use cases that are throwing false positives for us and see if we can patch these in groovy-cps.

steven-terrana:
I have opened #306 with unit tests showing the above mentioned.

jglick:
Given https://github.com/cloudbees/groovy-cps/pull/99, do we still think we need this API?

steven-terrana:
cloudbees/groovy-cps#99 will solve the use cases that caused us a problem.   

i think it would be okay to hold off on this. I would see value in it if there were many such scenarios, but i think the templating engine plugin might have just gotten some bad luck on this one - so the API probably wouldn't provide wide reaching value to most folks and introduce another external API and tech debt.

jglick:
@dwnusbaum WDYT?

dwnusbaum:
> WDYT? 

Yeah I would close this if cloudbees/groovy-cps#99 fixes the main issues (although since this commit was present in #306 and then reverted in that PR, GitHub has marked this as merged as well, so there is nothing to close here, maybe would have made more sense for me to squash 306 when merging)

PR #306, opened 2019-07-18
[JENKINS-58501] add ignored tests for metaprogramming CPS mismatch logs

steven-terrana:
per @jglick's advice in #305 , opening a PR with ignored unit tests for the case of CPS mismatch logs being emitted for: 

**MetaClassImpl.invokeMethod**: 
~~~
expected to call groovy.lang.MetaClassImpl.invokeMethod but wound up catching org.jenkinsci.plugins.workflow.cps.CpsClosure2.call
~~~

**ExpandoMetaClass.invokeMethod**:
~~~
expected to call groovy.lang.ExpandoMetaClass.invokeMethod but wound up catching org.jenkinsci.plugins.workflow.cps.CpsClosure2.call
~~~

3. **methodMissing**:
~~~
expected to call Example.doSomething but wound up catching Example.methodMissing
~~~


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -123,4 +124,52 @@
...
+        CpsVmExecutorService.FAIL_ON_MISMATCH = false;
+        try {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            errors.checkSucceeds(() -> {
+                p.setDefinition(new CpsFlowDefinition(
+                    "import org.codehaus.groovy.runtime.InvokerHelper \n" + 
+                    "c = { println 'doing a thing' } \n" +
+                    "InvokerHelper.getMetaClass(c).invokeMethod(c, 'call', null)", false));
+                WorkflowRun b = r.buildAndAssertSuccess(p);
+                r.assertLogNotContains("MetaClassImpl", b);
```
dwnusbaum: nit: nice to do something like this to make it extra clear what is happening without even having to run the test: 

```suggestion
                r.assertLogNotContains(CpsVmExecutorService.mismatchMessage("MetaClassImpl", "invokeMethod", "CpsClosure2", "call"), b); // Or whatever the right values are.
```

steven-terrana:
Thanks!

trying to lend a helping hand.  opened a PR to groovy-cps that makes the unit tests added in this PR pass. 

https://github.com/cloudbees/groovy-cps/pull/99

PR #307, opened 2019-07-18
[JENKINS-33217] Pass TaskListener to DescribableModel

car-roll:
Passing `FlowExecution`'s `TaskListener` down to `DescribableModel.instantiate` to allow pipeline visibility of stray parameter warnings.

These changes rely on the following `structs-plugin` changes: https://github.com/jenkinsci/structs-plugin/pull/49

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.29</groovy-cps.version>
-        <structs-plugin.version>1.20-rc294.782175d40f81</structs-plugin.version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/49 -->
+        <structs-plugin.version>1.20-rc297.69da1904839b</structs-plugin.version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/49 -->
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.29</groovy-cps.version>
-        <structs-plugin.version>1.20-rc297.69da1904839b</structs-plugin.version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/49 -->
+        <structs-plugin.version>1.20</structs-plugin.version>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.29</groovy-cps.version>
-        <structs-plugin.version>1.18</structs-plugin.version>
+        <structs-plugin.version>1.20-rc293.e3faccddadc2</structs-plugin.version>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.29</groovy-cps.version>
-        <structs-plugin.version>1.20-rc293.e3faccddadc2</structs-plugin.version>
+        <structs-plugin.version>1.20-rc294.782175d40f81</structs-plugin.version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/49 -->
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
     <dependencies>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -264,7 +264,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
                 s = d.newInstance(ps.namedArgs);
             } else {
                 DescribableModel<? extends Step> stepModel = DescribableModel.of(d.clazz);
-                s = stepModel.instantiate(ps.namedArgs, handle.getListener());
+                s = stepModel.instantiate(ps.namedArgs, context.get(TaskListener.class));
             }
 
             // Persist the node - block start and end nodes do their own persistence.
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -260,7 +260,12 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
             }
             d.checkContextAvailability(context);
             Thread.currentThread().setContextClassLoader(CpsVmExecutorService.ORIGINAL_CONTEXT_CLASS_LOADER.get());
-            s = d.newInstance(ps.namedArgs);
+            if (Util.isOverridden(StepDescriptor.class, d.getClass(), "newInstance", Map.class)) {
+                s = d.newInstance(ps.namedArgs);
+            } else {
+                DescribableModel<?> stepModel = DescribableModel.of(d.clazz);
```
dwnusbaum: nit: I _think_ the type parameters would be ok if you did the following, and then you wouldn't need the explicit cast to `Step` on the next line.
```suggestion
                DescribableModel<? extends Step> stepModel = DescribableModel.of(d.clazz);
```

pom.xml
```diff
@@ -72,7 +72,7 @@
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <groovy-cps.version>1.29</groovy-cps.version>
-        <structs-plugin.version>1.18</structs-plugin.version>
+        <structs-plugin.version>1.20-rc293.e3faccddadc2</structs-plugin.version>
```
jglick:
```suggestion
        <structs-plugin.version>1.20-rc293.e3faccddadc2</structs-plugin.version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/49 -->
```

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -260,7 +260,12 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
             }
             d.checkContextAvailability(context);
             Thread.currentThread().setContextClassLoader(CpsVmExecutorService.ORIGINAL_CONTEXT_CLASS_LOADER.get());
-            s = d.newInstance(ps.namedArgs);
+            if (Util.isOverridden(StepDescriptor.class, d.getClass(), "newInstance", Map.class)) {
```
jglick: …which is deprecated, so we do not care too much about those cases.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -260,7 +260,12 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
...
+                s = d.newInstance(ps.namedArgs);
+            } else {
+                DescribableModel<?> stepModel = DescribableModel.of(d.clazz);
+                s = (Step) stepModel.instantiate(ps.namedArgs, handle.getListener());
```
jglick: The cast is unnecessary if you do not throw away the type restriction `? extends Step` which `of` correctly produced:

```java
s = DescribableModel.of(d.clazz).instantiate(ps.namedArgs, handle.getListener());
```

jglick:
This needs an integration test BTW.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
dwnusbaum: Deprecated yes, but in particular this case will always used by `ParallelStep`, and what `ParallelStep` is doing cannot currently be replicated with `CustomDescribableModel`, so it is still somewhat important.

On the other hand overrides like [this](https://github.com/jenkinsci/pipeline-utility-steps-plugin/blob/0b912dac3870b880d354dcbe0d21cfff4b08ff93/src/main/java/org/jenkinsci/plugins/pipeline/utility/steps/AbstractFileOrTextStepDescriptorImpl.java#L24) could definitely be replaced with `CustomDescribableModel`.

jglick: > overrides like [this](https://github.com/jenkinsci/pipeline-utility-steps-plugin/blob/0b912dac3870b880d354dcbe0d21cfff4b08ff93/src/main/java/org/jenkinsci/plugins/pipeline/utility/steps/AbstractFileOrTextStepDescriptorImpl.java#L24) could definitely be replaced with `CustomDescribableModel`

Better to delete the override and throw `AbortException` from `StepExecution.start`.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -260,7 +260,12 @@ protected Object invokeStep(StepDescriptor d, String name, Object args) {
             }
             d.checkContextAvailability(context);
             Thread.currentThread().setContextClassLoader(CpsVmExecutorService.ORIGINAL_CONTEXT_CLASS_LOADER.get());
-            s = d.newInstance(ps.namedArgs);
+            if (Util.isOverridden(StepDescriptor.class, d.getClass(), "newInstance", Map.class)) {
+                s = d.newInstance(ps.namedArgs);
+            } else {
+                DescribableModel<? extends Step> stepModel = DescribableModel.of(d.clazz);
+                s = stepModel.instantiate(ps.namedArgs, handle.getListener());
```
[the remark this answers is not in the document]
car-roll: Yes, `handle.getListener()` is the overall build listener. Will change.

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -33,6 +33,7 @@
 import java.util.Set;
 import static org.hamcrest.Matchers.containsString;
 
+import org.hamcrest.Matchers;
```
basil: This import appears to be unused as of 8af1b4aa7501e3fcca039ffb5c25861eba0aa86a and can now be deleted.

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -49,6 +50,8 @@
 import org.jenkinsci.plugins.workflow.steps.SynchronousStepExecution;
 import org.jenkinsci.plugins.workflow.testMetaStep.AmbiguousEchoLowerStep;
 import org.jenkinsci.plugins.workflow.testMetaStep.AmbiguousEchoUpperStep;
+
+import static org.hamcrest.Matchers.instanceOf;
```
basil: This import appears to be unused as of 8af1b4a and can now be deleted.

dwnusbaum:
Build failure is due to an issue with `buildPipeline`, should be fixed by https://github.com/jenkins-infra/pipeline-library/pull/96.

dwnusbaum:
> [ERROR] error An unexpected error occurred: "ENOSPC: no space left on device, mkdir '/home/jenkins/.cache/yarn/v1/npm-jenkins-js-modules-1.3.0-d1872a0917a7b4bcae216a598ea7076f874c49be'".

Trying one more time for a passing build. If that doesn't work I think I think I will merge anyway since the current build passed on 5/6 branches. Maybe also good to get rid of the branches that build on Java 8 with Jenkins 2.164.1 at the same time since they aren't particularly useful and their presence increase the possibility of failures due to flaky infra.

PR #309, opened 2019-07-30
Add more tests for CPS mismatch warning false positives

dwnusbaum:
Adds two more tests for false positives, one for [JENKINS-58407](https://issues.jenkins-ci.org/browse/JENKINS-58407) and one for https://groups.google.com/d/msg/jenkinsci-users/2m5JGYJfYRs/t1hRV-rhFAAJ.

I don't really have any good ideas of how we could fix these cases without also introducing a bunch of false negatives. I think any serious fix would require recreating a good amount of Groovy's method dispatch logic so that we could figure out what `object.x()` is doing behind the scenes (For example, in the first new test case it ends up being something like `(object.x)()`, and in the second case it is something like `(object.get('x'))()`).

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
@@ -167,6 +167,24 @@ public class CpsVmExecutorServiceTest {
                 r.assertLogNotContains("methodMissing", b);
                 return null;
             });
+            errors.checkSucceeds(() -> {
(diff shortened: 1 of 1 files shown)


jglick:
And update the wiki to explain that you can work around these issues by simply interjecting `.call` into the call site.

dwnusbaum:
> And update the wiki to explain that you can work around these issues by simply interjecting .call into the call site.

Done, please review https://wiki.jenkins.io/display/JENKINS/Pipeline+CPS+method+mismatches#PipelineCPSmethodmismatches-FalsePositives when you get a chance.

PR #310, opened 2019-07-31
Add regression tests for SECURITY-1465

dwnusbaum:
Adding some regression tests for SECURITY-1465. Maybe some are unnecessary, but I'd like to have at least `blockLhsInMethodPointerExpression`, `blockRhsInMethodPointerExpression`, and `blockCastingUnsafeUserDefinedImplementationsOfCollection` in case changes to groovy-cps affect those tests.


dwnusbaum:
> win2012-1aefc0 was marked offline: Connection was broken: java.nio.channels.ClosedChannelException

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -634,4 +634,72 @@ public void scriptInitializerCallsCpsTransformedMethod() throws Exception {
         jenkins.assertLogContains("CpsCallableInvocation{methodName=bar,", b);
     }
 
+    @Issue("SECURITY-1465")
+    @Test public void blockLhsInMethodPointerExpression() throws Exception {
+        WorkflowJob p = jenkins.jenkins.createProject(WorkflowJob.class, "p");
```
jglick: This suite is a really good candidate for `@ClassRule` BTW. Maybe pay that tax now, so you are decreasing rather than increasing test time? (And stop using `AbstractCpsFlowTest`. Its utilities are not used here AFAIK, and it ought to be deprecated or deleted.)

dwnusbaum: Fair enough, I'll look into using `@ClassRule` here.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -32,55 +32,29 @@
...
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.LoggerRule;
 
-public class CpsFlowDefinition2Test extends AbstractCpsFlowTest {
+public class CpsFlowDefinition2Test {
 
     @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @ClassRule public static JenkinsRule jenkins = new JenkinsRule();
```
dwnusbaum: Cuts the test time from ~65 seconds to ~35 seconds on my machine.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
dwnusbaum: Done in 40b09db and 5127785.

src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
```diff
@@ -33,31 +33,31 @@
...
     @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @ClassRule public static JenkinsRule jenkins = new JenkinsRule();
 
-    WorkflowJob p;
+    private static WorkflowJob p;
 
-    @Before public void setUp() throws Exception {
-        super.setUp();
+    @BeforeClass public static void setUp() throws Exception {
         p = jenkins.jenkins.createProject(WorkflowJob.class, "demo");
```
jglick: Note that this will reuse the same project each time. Probably fine, but it might be better to delete this method and add

```java
WorkflowJob p = jenkins.createProject(WorkflowJob.class);
```

to the start of every test method. Or just revert the change to the field and to the signature of this method and merely delete the `name` argument.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -1,47 +0,0 @@
...
-
-import org.junit.Test;
-
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.assertEquals;
-
-/**
- * @author Kohsuke Kawaguchi
- */
-public class CpsFlowDefinitionTest extends AbstractCpsFlowTest {
```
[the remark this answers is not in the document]
jglick: Yes these tests date to the first days of Workflow and are not very useful now.

Worth checking whether `FlowExecution.getCauseOfFailure` is covered anywhere else, though I am not even sure if anyone calls it anyway.

src/test/java/org/jenkinsci/plugins/workflow/WorkflowJobNonRestartingTest.java
dwnusbaum: Good point, I switched to creating a project in every test method in ea3cd8a.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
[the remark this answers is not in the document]
dwnusbaum: I went ahead and added coverage for `FlowExecution.getCauseOfFailure` to a generic sandbox test in ea3cd8a since it didn't look like it was used anywhere else.

PR #311, opened 2019-07-31
[JENKINS-58501] Update to groovy-cps 1.30

dwnusbaum:
See [JENKINS-58501](https://issues.jenkins-ci.org/browse/JENKINS-58501).

Picks up https://github.com/cloudbees/groovy-cps/pull/99 from @steven-terrana to suppress the mismatch warnings for some kinds of metaprogramming that confuse the mismatch detection code.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <groovy-cps.version>1.29</groovy-cps.version>
+        <groovy-cps.version>1.30</groovy-cps.version>
         <structs-plugin.version>1.18</structs-plugin.version>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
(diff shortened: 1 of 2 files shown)


jglick:
Should this back out #305?

dwnusbaum:
@jglick see https://github.com/jenkinsci/workflow-cps-plugin/pull/305#issuecomment-514667427. I didn't notice that #306 included #305 but then reverted it, and did not squash-merge, so GitHub thinks #305 was merged, which is not technically wrong, but it was immediately reverted by the next commit in #306.

jglick:
So the upshot is that the next release will not include the unnecessary new API? Fine.

dwnusbaum:
> So the upshot is that the next release will not include the unnecessary new API?

Yes, the new API will not be available in the next release and was never available from `HEAD` of `master` at any point, it just _looks_ like it was merged because the [commit](4e4e3ed) in #305 was also in #306 (along with the [commit](e7319b7) to revert #305), so when #306 got merged GitHub automatically marked #305 as merged (technically right but very confusing). Would not have been so confusing if #306 was based on `master` instead of on #305 or if I had squash-merged #306 so that the addition and reversion were cancelled out before they hit `master`.

PR #313, opened 2019-08-06
[JENKINS-58817]Creating a startup argument

alexanderrtaylor:
Created a startup argument to be able to configure how long it takes for remote requests to timeout. By default it was 5 minutes but it may be useful to control that


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -167,7 +173,7 @@ public StepExecution getStep() {
         final CpsThread old = CURRENT.get();
         CURRENT.set(this);
 
-        try (Timeout timeout = Timeout.limit(5, TimeUnit.MINUTES)) {
+        try (Timeout timeout = Timeout.limit(REMOTE_TIMEOUT, TimeUnit.MINUTES)) {
```
dwnusbaum: IIUC, this is the timeout for executing a single chunk of CPS-transformed code in the CPS VM, which normally shouldn't take too long because each call to a continuation (or maybe only some of them) should come back here and reset the timer. I think if you are hitting this timeout you are doing something wrong like calling `Thread.sleep` directly in your Pipeline. Do you have an example case of where this would be useful?

alexanderrtaylor: @dwnusbaum There was a situation where a remote call (as a function) was being made from an Operations Center to a master as a groovy script and the script took a long time to execute gathering data. The return did eventually come back it just took a while and it was seen as a single chunk which meant it terminated after 5 minutes.

The biggest issue is that there was not really a good way to `batch` the groovy script to return periodically since that would add additional remoting threads that had to be out there and did not provide any sort of actual benefit to the code

dwnusbaum: What I mean though is that a Pipeline should not be making remoting calls directly in the first place. That is something that should be done in a step so it can happen on a background thread rather than holding up this thread.

There is related discussion in https://github.com/jenkinsci/workflow-cps-plugin/pull/102, which added the timeout. In particular, see https://github.com/jenkinsci/workflow-cps-plugin/pull/102#issuecomment-279398217.

dwnusbaum:
Given the reasons above, I am going to close this. If you are running into this timeout, you are probably trying to do something in a Pipeline that it was not designed to do. Consider moving whatever code is blocking the CPS VM thread for more than 5 minutes into an asynchronous `Step` that executes on a background thread instead.

PR #314, opened 2019-08-13
[JENKINS-58643] Ignore mismatch warnings when receiver is a plugin

RobiNino:
This patch was proposed by @dwnusbaum in [JENKINS-58643](https://issues.jenkins-ci.org/browse/JENKINS-58643), and it successfully resolves the issue described there.

When handling CPS mismatch, warnings are ignored whenever the receiver is a plugin.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -11,6 +11,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
+import jenkins.model.Jenkins;
 import jenkins.util.InterceptingExecutorService;
 
 /**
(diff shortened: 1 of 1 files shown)


dwnusbaum:
@RobiNino Thanks for the PR! It would be great to have some kind of regression test if possible. I think you might be able to create a test that uses `@TestExtension` with a `GlobalVariable` that does something bad and then use it in a test and assert that no warnings are logged because it came from a `GlobalVariable` defined in a Plugin rather than CPS-transformed Groovy code.

Were you able to check if this patch also masked errors in `pipeline-model-definition` or `docker-workflow` as discussed [here](https://issues.jenkins-ci.org/browse/JENKINS-58643?focusedCommentId=372513&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-372513)?

dwnusbaum:
Closing and reopening for a fresh build.

abayer:
I'm re-running the build.

robertwedd:
@jglick @dwnusbaum when can we expect review to be done so that the Fix can come in?

dwnusbaum:
@robertwedd I looked at the PR and asked some questions: https://github.com/jenkinsci/workflow-cps-plugin/pull/314#issuecomment-520950751. Personally I'd like to see a test for this and some kind of investigation into whether it affects things like Declarative before I approve it.

RobiNino:
@dwnusbaum Currently working on something else, will get back to work on the PR as soon as I'm done.
From a quick look on the test, do you mean implementing a `@TestExtension` class that extends `GlobalVariable`, then use the class in a `CpsFlowDefinition`? If not in a `CpsFlowDefinition`, in what environment should I call it?
Thanks

abayer:
Any update here? I'd like to be able to merge it, but it does still need the tests @dwnusbaum talks about.

eyalbe4:
@dwnusbaum,
Anything else blocking this merge?
Thanks

robertwedd:
@dwnusbaum @abayer do we know if this merge fixes the issue described here https://issues.jenkins-ci.org/browse/JENKINS-58643 if so does the customer need to just update the workflow-cps-plugin i also found information here which says something else needs to be changed https://issues.jenkins-ci.org/browse/JENKINS-58407

PR #315, opened 2019-08-13
 JENKINS-58620 Test for the GroovyShell.evaluate mismatch ignore

sparshev:
See [JENKINS-58620](https://issues.jenkins-ci.org/browse/JENKINS-58620).

CpsGroovyShell test required never version of groovy-cps plugin
Please do not merge until cloudbees/groovy-cps#101 will be accepted and new groovy-cps will be released.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <groovy-cps.version>1.30</groovy-cps.version>
+        <groovy-cps.version>1.31</groovy-cps.version>
         <structs-plugin.version>1.20</structs-plugin.version>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
@@ -160,6 +160,24 @@ public class CpsVmExecutorServiceTest {
         }
     }
 
+    @Issue("JENKINS-58620")
+    @Test public void wrongCatcherGroovyShellEvaluate() throws Exception {
+        boolean origFailOnMismatch = CpsVmExecutorService.FAIL_ON_MISMATCH;
+        CpsVmExecutorService.FAIL_ON_MISMATCH = false;
+        try {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            errors.checkSucceeds(() -> {
+                p.setDefinition(new CpsFlowDefinition("org.jenkinsci.plugins.workflow.cps.CpsThread.current().getExecution().getShell().evaluate('echo \"Goofy\"', 'X.groovy')", false));
+                WorkflowRun b = r.buildAndAssertSuccess(p);
+                r.assertLogContains("Goofy", b);
+                r.assertLogNotContains("expected to call", b);
+                return null;
+            });
+        } finally {
+            CpsVmExecutorService.FAIL_ON_MISMATCH = origFailOnMismatch;
+        }
+    }
+
     @Issue("JENKINS-58407")
     @Ignore
     @Test public void mismatchFalsePositives() throws Exception {


pom.xml
```diff
@@ -71,7 +71,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <groovy-cps.version>1.30</groovy-cps.version>
+        <groovy-cps.version>1.31</groovy-cps.version>
```
jglick: You can use a deployed timestamped snapshot (pending JEP-305 configuration for this library).

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -160,6 +160,24 @@
         }
     }
 
+    @Issue({"JENKINS-58620"})
+    @Test public void wrongCatcherGroovyShellEvaluate() throws Exception {
```
jglick: This can just be added to `wrongCatcher` I think. (Or the class could be reworked a bit to use `JenkinsRule` as a `ClassRule`.)

sparshev: I don't really like to put everything in one place and unfortunately have not enough knowledge of the logic that is tested - so can't properly rewrite the class for optimization...

pom.xml
sparshev: Ok, could you please direct me - probably I have no access to find the right snapshot for groovy-cps, do you know where to look?

jglick: Pending JEP-305 support for this library (promised by @dwnusbaum at some point), you would have to `mvn deploy` upstream manually, which presumes that you are a plugin developer and thus authenticated to Artifactory.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
jglick: Should probably be put on hold until a plugin maintainer can refactor this class accordingly.

pom.xml
sparshev: Right, so since groovy-cps#101 is a feature branch and I'm not a plugin developer without access to Artifactory - it will never be available for this pull request. We're stuck here.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
sparshev: You mean we have to wait for some not planned action just to merge the unit tests? Looks not promising at all. If you really want - I can move this unit test to `wrongCatcher` if it will make any better to be close to merge. But I really think it's not necessary for the current case.

jglick: Well we need to wait until upstream is `deploy`ed anyway.

pom.xml
sparshev: Ok, now 1.31 release of the groovy-cps is available - and working well with the change.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
sparshev: Sure. I updated a little bit - removed curved brackets since they don't needed in `@Issue`.

jglick: Assuming #320 is merged, you should be able to add a new test without adding significantly to execution time.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -160,6 +160,24 @@
...
     }
 
+    @Issue("JENKINS-58620")
+    @Test public void wrongCatcherGroovyShellEvaluate() throws Exception {
+        boolean origFailOnMismatch = CpsVmExecutorService.FAIL_ON_MISMATCH;
+        CpsVmExecutorService.FAIL_ON_MISMATCH = false;
+        try {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            errors.checkSucceeds(() -> {
+                p.setDefinition(new CpsFlowDefinition("org.jenkinsci.plugins.workflow.cps.CpsThread.current().getExecution().getShell().evaluate('echo \"Goofy\"', 'X.groovy')", false));
```
jglick: I suspect just

```suggestion
                p.setDefinition(new CpsFlowDefinition("evaluate('echo(/Goofy/)', 'X.groovy')", false));
```

would suffice to reproduce.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
sparshev: Ok, thank you - will check and rewrite accordingly.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
sparshev: No, this logic will not test the issue. I tried to use some compact commands - but only the one I used here is showing the notice.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
sparshev: I will prepare a change when #320 will be merged - because most probably it will conflict with my changes.

sparshev:
Actually I thought #320 will be merged first...

abayer:
Oh, crap, sorry - @jglick, do you need any help with #320?

sparshev:
@abayer you can revert my changes, merge #320 and I will change my PR to fit the master.

PR #316, opened 2019-08-19
[JENKINS-58407] Verify that `foo.someField()` doesn't trigger mismatch

abayer:
[JENKINS-58407](https://issues.jenkins-ci.org/browse/JENKINS-58407)

See https://github.com/cloudbees/groovy-cps/pull/102.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
@@ -161,8 +161,7 @@ public class CpsVmExecutorServiceTest {
     }
 
     @Issue("JENKINS-58407")
-    @Ignore
-    @Test public void mismatchFalsePositives() throws Exception {
+    @Test public void closureInvokeShouldBeTreatedAsClosureCall() throws Exception {
         boolean origFailOnMismatch = CpsVmExecutorService.FAIL_ON_MISMATCH;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -161,8 +161,7 @@
     }
 
     @Issue("JENKINS-58407")
-    @Ignore
-    @Test public void mismatchFalsePositives() throws Exception {
+    @Test public void closureInvokeShouldBeTreatedAsClosureCall() throws Exception {
```
jglick: Minor merge conflict with #320.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorServiceTest.java
```diff
@@ -179,7 +178,7 @@
             });
             errors.checkSucceeds(() -> {
                 p.setDefinition(new CpsFlowDefinition(
-                    "def cs = [ action: {-> sleep(1) } ]\n" +
+                    "def cs = [ action: {-> echo 'test' } ]\n" +
```
jglick: I do not think you want to do that. `echo` is synchronous, which is why these tests use `sleep`. Anyway edited in #320.

dwnusbaum:
#315 updated to groovy-cps 1.31, and #320 unignored the tests (and rearranged them), so I am closing this.

PR #317, opened 2019-08-27
[JENKINS-59083] Do not try to open the listener for a completed build

dwnusbaum:
See https://github.com/jenkinsci/workflow-job-plugin/pull/140.


jglick:
Downstream passes with this so I think it can be merged.

dwnusbaum:
I was thinking about adding a test case here, but if the downstream PR looks ok then I don't think it's worth the effort.

PR #319, opened 2019-08-30
Add `loadFromClasspath()` pipeline step

awittha:
This adds a `loadFromClasspath(...)` pipeline step. It's just like the `load(...)` step, except it loads from the Jenkins master's classpath instead of from an agent's disk.

Usage
=====

```groovy
loadFromClasspath(
    classloaderSource: this,
    path: 'com/example/Pipeline.groovy' )
```

loadFromClasspath Title

loadFromClasspath description

Motivation
=====

I have Groovy pipeline scripts provided to my Jenkins masters' classpaths, and I need a way to run them from other pipeline scripts.

Per a conversation with @jglick about the interface changes to `CpsStepContext#newBodyInvoker` that "nobody should be using methods like this from external code," I'd like to see the capability I need become _internal_ to `workflow-cps`.

I Have Questions
=====

I have _not_ written tests for this step yet, because I don't fully understand the semantics of the new `unexport` option on `CpsStepContext#newBodyInvoker`. I'd like to understand that better (and be able to update the accompanying documentation and comments) before moving forward.

Please see the specific line-comments for the uncertainties I still have.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStep.java
@@ -1,0 +1,106 @@
+package org.jenkinsci.plugins.workflow.cps.steps;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.jenkinsci.plugins.workflow.cps.CpsStepContext;
+import org.jenkinsci.plugins.workflow.cps.CpsThreadGroup;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
+
+import hudson.Extension;
+import hudson.model.TaskListener;
+
+/**
+ * Load the contents of some classpath resource and attempt to evaluate them as a Groovy Pipeline script in the current pipeline execution.
+ * 
+ * @author Austin Witt
+ * 
+ * @since 2.74
+ *
+ */
+public class LoadFromClasspathStep extends Step {
+
+	/**
+	 * The classpath of a Groovy Pipeline script
+	 */
+	protected String path;
+
+	/**
+	 * An object to use to get the classloader to use.
+	 * <br><br>
+	 * Since this will be targeting a Groovy script file (not bytecode) resource on the classpath,
+	 * a GroovyClassLoader must be used to look for the class to ensure that the Groovy script text is located and compiled.
+	 * <br><br>
+	 * Using a regular Java ClassLoader will only look for bytecode. 
+	 * This may fail - but if it does succeed, the resulting script's source code will not be able to be CPS-transformed
+	 * (since the source code was not loaded) and so it won't be able to be used by pipeline scripts without breaking the whole pipeline execution.
+	 * <br><br>
+	 * An easy way to find a GroovyClassLoader is to provide the current Groovy script (<code>this</code>) as the object.
+	 */
+	protected Object classloaderSource;
+	
+	/**
+	 * Whether properties from the loaded script should be unexported when serializing the pipeline.
+	 * Setting this to <code>true</code> may affect property availability after pipeline restarts.
+	 * Setting this to <code>false</code> may make some pipelines non-resumable.
+	 * 
+	 * @see CpsStepContext#newBodyInvoker(org.jenkinsci.plugins.workflow.cps.BodyReference, boolean)
+	 * @see CpsThreadGroup#unexport(org.jenkinsci.plugins.workflow.cps.BodyReference)
+	 */
+	protected boolean unexport = DescriptorImpl.DEFAULT_UNEXPORT;
+
+	@DataBoundConstructor
+	public LoadFromClasspathStep(String path) { this.path = path; }
+
+	public String getPath() { return path; }
+
+	public Object getClassloaderSource() { return classloaderSource; }
+
+	@DataBoundSetter
+	public void setClassloaderSource(Object classloaderSource) { this.classloaderSource = classloaderSource; }
+	
+	public boolean isUnexport() { return unexport; }
+	
+	@DataBoundSetter
+	public void setUnexport( boolean unexport ) { this.unexport = unexport; }
+
+	@Override
+	public StepExecution start(StepContext _context) throws Exception {
+		return new LoadFromClasspathStepExecution( _context, path, classloaderSource, unexport );
+	}
+
+	@Extension
+	public static class DescriptorImpl extends StepDescriptor {
+		
+		public static final boolean DEFAULT_UNEXPORT = false;
+
+		@Override
+		public Set<? extends Class<?>> getRequiredContext() {
+			return new HashSet<>(Arrays.asList(
+				TaskListener.class
+			));
+		}
+
+		@Override
+		public String getFunctionName() {
+			return "loadFromClasspath";
+		}
+
+		@Override
+		public String getDisplayName() {
+			return "Evaluate the contents of a classpath resource as a Groovy Pipeline script";
+		}
+
+		@Override
+		public boolean isAdvanced() {
+			return true;
+		}
+	}
+
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStepExecution.java
@@ -1,0 +1,134 @@
+package org.jenkinsci.plugins.workflow.cps.steps;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Scanner;
+
+import org.apache.tools.ant.taskdefs.Classloader;
+import org.codehaus.groovy.control.MultipleCompilationErrorsException;
+import org.jenkinsci.plugins.workflow.cps.CpsCompilationErrorsException;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
+import org.jenkinsci.plugins.workflow.cps.CpsStepContext;
+import org.jenkinsci.plugins.workflow.cps.CpsThread;
+import org.jenkinsci.plugins.workflow.cps.replay.ReplayAction;
+import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
+
+import groovy.lang.Script;
+import hudson.model.TaskListener;
+
+/**
+ * Execution implementation for {@link LoadFromClasspathStep}.
+ *
+ * @author Austin Witt
+ * 
+ * @since 2.74
+ *
+ */
+public class LoadFromClasspathStepExecution extends StepExecution {
+	
+	private static final long serialVersionUID = 1L;
+
+	protected String path;
+
+	protected Object classloaderSource;
+	
+	protected boolean unexport;
+	
+	protected transient TaskListener listener;
+
+	public LoadFromClasspathStepExecution(
+		StepContext _context, 
+		String _path, 
+		Object _classloaderSource,
+		boolean _unexport)
+			throws
+				IOException,
+				InterruptedException
+	{
+		super( _context );
+
+		path = _path;
+		classloaderSource = _classloaderSource;
+		listener = _context.get( TaskListener.class );
+	}
+
+	@Override
+	public boolean start() throws Exception {
+		
+		CpsStepContext cps = (CpsStepContext) getContext();
+		CpsThread t = CpsThread.current();
+		CpsFlowExecution execution = t.getExecution();
+
+		String text = get_classpath_resource_contents( classloaderSource != null ? classloaderSource : this, path );
+		String clazz = execution.getNextScriptName( path );
+		String newText = ReplayAction.replace( execution, clazz );
+		if( newText != null ) {
+			listener.getLogger().println( "Replacing Groovy text with edited version" );
+			text = newText;
+		}
+
+		Script script;
+		try {
+			script = execution.getShell().parse(text);
+		} catch (MultipleCompilationErrorsException e) {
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStepExecution.java
```diff
@@ -0,0 +1,134 @@
...
+		try {
+			script = execution.getShell().parse(text);
+		} catch (MultipleCompilationErrorsException e) {
+			// Convert to a serializable exception, see JENKINS-40109.
+			throw new CpsCompilationErrorsException(e);
+		}
+
+		// execute body as another thread that shares the same head as this thread
+		// as the body can pause.
+		cps.newBodyInvoker(t.getGroup().export( script ), unexport )
```
awittha: Based on the contents of `LoadStepExecution`, I guessed that `unexport: true` was the value I should use here.

However, I found the following issue:

# `@groovy.transform.Field` not available after pipeline restart

`Jenkinsfile`
```groovy
@groovy.transform.Field
def FOO=null

FOO="bar"

loadFromClasspath("com/example/Pipeline.groovy")
```

`Pipeline.groovy`
```
echo "FOO=${FOO}"
checkpoint()
echo "FOO="${FOO}"
```

This pipeline runs, but after resuming at the `checkpoint()`, property `FOO` is undefined. Switching to `unexport: false` fixed this.

# Something Else I didn't Pin Down

There was another issue which may have just been that `GlobalVariables` that got loaded during a `loadFromClasspath()`'d script weren't available after pipeline restart, either. However, after determining that `unexport: false` fixed the first issue, I checked and it also fixed the other issue I was seeing. I'm postponing further voodoo troubleshooting until I have a better understanding of `unexport`.

src/main/resources/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStep/config.jelly
```diff
@@ -0,0 +1,20 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form">
+
+	<f:entry field="path" title="${%Classpath Resource Path}" description="The classpath location of the Groovy Pipeline Script to load">
+	    <f:textbox/>
+	</f:entry>
+	
+	<f:entry field="classloaderSource" title="${%Classpath Hint Object}" description="An object to grab a ClassLoader from to use to load the script">
+	    <f:textbox default="null"/>
```
awittha: I would like the name the developer types into this box to be used in the snippet generator exactly as-is, _without_ surrounding it with quotation marks, so that `this` or `someVariableThatWillExistInTheirPipeline` can be used.

I haven't been able to figure this out - is it possible with the stapler-powered snippet generator? If so, what do I need to do?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStep.java
```diff
@@ -0,0 +1,106 @@
...
+	protected Object classloaderSource;
+	
+	/**
+	 * Whether properties from the loaded script should be unexported when serializing the pipeline.
+	 * Setting this to <code>true</code> may affect property availability after pipeline restarts.
+	 * Setting this to <code>false</code> may make some pipelines non-resumable.
+	 * 
+	 * @see CpsStepContext#newBodyInvoker(org.jenkinsci.plugins.workflow.cps.BodyReference, boolean)
+	 * @see CpsThreadGroup#unexport(org.jenkinsci.plugins.workflow.cps.BodyReference)
+	 */
```
awittha: I did the best I could for this JavaDoc with my current understanding of `unexport`. Please let me know if any of it is wrong, misleading, or suboptimal.

jglick:
I do not see a need for such a step. You can already do something like this. See `docker-workflow` etc. But we do not encourage this style of extension. Exposing this sort of feature as a general step also invites security exploits that we would rather not deal with.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadFromClasspathStepExecution.java
```diff
@@ -0,0 +1,134 @@
...
+	
+	private static final long serialVersionUID = 1L;
+
+	protected String path;
+
+	protected Object classloaderSource;
+	
+	protected boolean unexport;
+	
+	protected transient TaskListener listener;
```
awittha: If this is not `transient`, this fails to serialize:

```
an exception which occurred:
	in field org.jenkinsci.plugins.workflow.log.BufferedBuildListener$Replacement.ros
	in object org.jenkinsci.plugins.workflow.log.BufferedBuildListener$Replacement@4a3299ab
	in field org.jenkinsci.plugins.workflow.log.TaskListenerDecorator$CloseableTaskListener.closeDelegate
	in object org.jenkinsci.plugins.workflow.log.TaskListenerDecorator$CloseableTaskListener@33aa7b11
	in field org.jenkinsci.plugins.workflow.cps.steps.LoadFromClasspathStepExecution.listener
	in object org.jenkinsci.plugins.workflow.cps.steps.LoadFromClasspathStepExecution@3428f778
	in field org.jenkinsci.plugins.workflow.cps.CpsThread.step
	in object org.jenkinsci.plugins.workflow.cps.CpsThread@56b269e8
	in field org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.threads
	in object org.jenkinsci.plugins.workflow.cps.CpsThreadGroup@4bfe79d1
	in object org.jenkinsci.plugins.workflow.cps.CpsThreadGroup@4bfe79d1
Caused: java.io.NotSerializableException: The calling thread Thread[Running CpsFlowExecution[Owner[devtools/test-project-please-ignore/master/10:devtools/test-project-please-ignore/master #10]],5,main] has no associated channel. The current object hudson.remoting.RemoteOutputStream@19de90ba is interface org.jenkinsci.remoting.SerializableOnlyOverRemoting, but it is likely being serialized/deserialized without the channel
	at org.jenkinsci.remoting.SerializableOnlyOverRemoting.getChannelForSerialization(SerializableOnlyOverRemoting.java:67)
	at hudson.remoting.RemoteOutputStream.writeObject(RemoteOutputStream.java:83)
```

If this _is_ `transient`, SpotBugs complains: 

```
[2019-08-30T16:34:29.442Z] [ERROR] The field org.jenkinsci.plugins.workflow.cps.steps.LoadFromClasspathStepExecution.listener is transient but isn't set by deserialization [org.jenkinsci.plugins.workflow.cps.steps.LoadFromClasspathStepExecution] In LoadFromClasspathStepExecution.java SE_TRANSIENT_FIELD_NOT_RESTORED
```

I'm okay agreeing with SpotBugs' assessment, but... the [`LoadStepExecution`](https://github.com/jenkinsci/workflow-cps-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java#L30) has a `transient TaskListener` that doesn't show up in the SpotBugs warnings.... and as far as I can tell, doesn't explicitly re-set the `listener` when it's deserialized.

Have I missed something here? What's the right thing to do with this listener?

awittha:
When you refer to `docker-workflow`, are you referring to https://github.com/jenkinsci/docker-workflow-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/docker/workflow/Docker.groovy ?

awittha:
There are some important differences between this `loadFromClasspath()` and the `GlobalVariable` extension point such that I require `loadFromClasspath()` for some loading and `GlobalVariable` for others.

First, the GlobalVariable load is not parameterizable, as it's delegated to by a Jenkins pipeline property lookup. This means that GlobalVariable-loaded groovy is in a 1:1 relationship between groovy source file and named, usable pipeline step. (Originally, this is what I did - with a parent class and a new child implementation for every pipeline script I needed to load. A lot of copy/paste of boilerplate!)

Second, GlobalVariable-based script loading didn't interface nicely with build replays - this `loadFromClasspath()` does. When the semantics of the loaded code is "pipeline script" (e.g. a prepared `Jenkinsfile`) and _not_ "some object that is merely the vehicle by which you do things" (as in `docker-workflow`s `docker` variable), properly behaving during replays is a consideration.

I require a way to load Pipeline code from the classpath, such that it can interface with build replays. Ideally, the boilierplate required by each individual occurence of this use would be minimal. This `loadFromClasspath()` implementation seemed to be the optimal solution.

I currently implement this in my own plugin, which has been working fine for years. However, if it's preferred that "external" code shouldn't interact with `CpsStepContext#newBodyInvoker`, then I ought to move the implementation to `workflow-cps`. 

Alternatively, if there's another, preferred mechanism to load pipeline scripts from the classpath such that they interact with build replays correctly, let me know and I can see about migrating over to that - I'd certainly prefer to maintain less code, rather than more!

dwnusbaum:
> Alternatively, if there's another, preferred mechanism to load pipeline scripts from the classpath such that they interact with build replays correctly, let me know and I can see about migrating over to that

I think the preference is that extensions do not load scripts like this at all, i.e. avoid do anything that even resembles what docker-workflow does with the `docker` `GlobalVariable`, and instead adapt their use cases to work either as a shared library or a normal `Step` if possible. 

Do you have any examples of a script you are loading using a step like this so I can try to understand your use case?

awittha:
@dwnusbaum I do not have examples that I can share on the public internet at the moment. However, the primary scripts loaded this way are Pipeline definitions of two classes:

1. such as one would find in a software project's `Jenkinsfile` - e.g. a delivery pipeline
2. such as one would find in the "pipeline script" WebUI box - e.g. a "task" definition that is consumed by "building with parameters."

They're on the classpath and _not_ a shared library because we deliver shared/re-usable pipeline code via plugins, not shared libraries. This was an intentional decision driven by business requirements after experimenting with both mechanisms.

Key things that the Shared Library mechanism doesn't enable that we can do from plugins include:
- dependency specification & resolution (the pipeline scripts will always run because every step they use will have its plugin installed when the script lands on the Jenkins master)
- as a result of above, _composability_ - multiple groups can work on inter-dependent components on their own schedule and in their own repos, and consumers don't have to meticulously hand-sculpt their Jenkins master's configuration & plugin list.
- whitelisting (plugins can ship an appropriate whitelist for their pipeline steps, whereas shared libraries must be run fully trusted or go through manual approval for every signature encountered on every branch of execution)
- snippet generator (plugins can populate the snippet generator with usage information in an intuitive fashion, shared libraries cannot)
- implicit convention for software lifecycle: Plugins build and publish with Maven, and they all do it the same way - and are consumed via a finalized, packaged artifact. Shared libraries not only carry no implication of a build lifecycle (meaning everyone rolls their own slightly-different one from scratch), but are consumed directly out of source-control refs; `master` may be consumed just as easily as a tag.
- consistency - plugins can only be installed by users with that master-wide permission, and they affect the whole master. Every build that uses something from a plugin will use the _same_ thing - whereas "I'm using shared library X" may actually refer to radically different codebases across builds, depending on folder or `Jenkinsfile` configuration.

jglick:
> the GlobalVariable load is not parameterizable

Not clear what this means. Like any Groovy object, it can define methods with arbitrary parameters.

> GlobalVariable-based script loading didn't interface nicely with build replays

What specifically were you looking for here? That after a Jenkins restart across which a plugin bundling a DSL-like `GlobalVariable` was updated, a replay of a build preceding the restart would use the previous rather than the current DSL?

> Key things that the Shared Library mechanism doesn't enable that we can do from plugins include:

Probably I cannot convince you to revisit this decision, but just in case:

> 1. **Dependency specification & resolution**: the pipeline scripts will always run because every step they use will have its plugin installed when the script lands on the Jenkins master

Not really sure what this means.

> 2. as a result of above, **Composability** - multiple groups can work on inter-dependent components on their own schedule and in their own repos, and consumers don't have to meticulously hand-sculpt their Jenkins master's configuration & plugin list.

This either. If anything, libraries should make this much simpler, as no particular master configuration or plugin list is required.

> 3. also as a result of the above, pipelines can **use JVM classes that aren't Pipeline steps** (e.g. POJO or DTO objects from 3rd-party libraries) - because those libraries can be a dependency of the plugin. Shared libraries cannot do this on their own.

Actually they can (using `@Grab`), though we really discourage this. Any serious computation is better done in external processes.

> 4. **Whitelisting**: plugins can ship an appropriate whitelist for their pipeline steps, whereas shared libraries must be run fully trusted or go through manual approval for every signature encountered on every branch of execution

Which I would consider an advantage: best to use sandboxed scripts with no special whitelist.

> 5. **Snippet generator:** plugins can populate the snippet generator with usage information in an intuitive fashion, shared libraries cannot

Indeed, though the step proposed here does not improve the situation.

> 6. **Implicit convention for software lifecycle:** Plugins build and publish with Maven, and they all do it the same way - and are consumed via a finalized, packaged artifact. Shared libraries not only carry no implication of a build lifecycle (meaning everyone rolls their own slightly-different one from scratch), but are consumed directly out of source-control refs; `master` may be consumed just as easily as a tag.

The default implementations of `LibraryResolver` indeed are based on SCM checkouts, for convenience. It should be straightforward to provide alternate implementations, for example based on a Maven resolver.

> 7. **Consistency:**  plugins can only be installed by users with that master-wide permission, and they affect the whole master. Every build that uses something from a plugin will use the _same_ thing - whereas "I'm using shared library X" may actually refer to radically different codebases across builds, depending on folder or `Jenkinsfile` configuration.

If your library definition allows its default version to overridden and your resolver decides to _offer_ this sort of version resolution, sure, which is an (important) feature. You can choose to lock down the contents instead.

> I currently implement this in my own plugin, which has been working fine for years.

Then I would suggest maintaining the status quo.

> if it's preferred that "external" code shouldn't interact with `CpsStepContext#newBodyInvoker`

It is certainly preferred, but it is of course up to you what proprietary code running on site calls. Just do not expect this to be a stable or documented API; it is part of the internal implementation of Pipeline. The supported plugin extension interface for Pipeline is `Step`, plus the semi-deprecated `GlobalVariable`.

awittha:
@jglick, to be clear, the decision to use plugins over libraries has many reasons, and is presented as the reason why I use `loadFromClasspath(...)` as implemented in this PR. There justification for that choice was fleshed out to show that there was a reason for getting into the situation that led to using `loadFromClasspath(...)`, not as direct justifications of `loadFromClasspath(...)` itself.

If the recommendation is now

> Then I would suggest maintaining the status quo _[given the additional context since-provided]_.

instead of "external code really shouldn't call [`CpsStepContext#newBodyInvoker(...)`]," I'm OK with that - I just wanted to be sure I was consuming Jenkins code & participating in Jenkins OSS in the most-optimal way.

In that case, I don't need this PR and don't need to continue taking up folk's time trying to justify it (but I'd be happy to continue to elaborate on the things I've said that weren't clear, if you're interested).

Please let me know!

PR #323, opened 2019-09-16
[JENKINS-59398] Flatten UserIdCause into ReplayCause

jsoref:
This is a proposal for https://issues.jenkins-ci.org/browse/JENKINS-59398

The other approach would be teaching blueocean to do the same work.


jglick:
Related to #398?

PR #325, opened 2019-09-18
[JENKINS-49073] Parallel step should propagate the worst result when not using failFast

basil:
### Problem

See [JENKINS-49073 (comment)](https://issues.jenkins-ci.org/browse/JENKINS-49073?focusedCommentId=375702&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-375702).

### Solution

Inspired by [jenkinsci/pipeline-build-step-plugin#24 (comment)](https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24#issuecomment-532671888). Add a new `propagateWorst` property to the `parallel` step that implements the desired behavior. When this is in effect, the "worst" branch propagates its build status, as defined by the ordering Jesse suggested.

### Implementation

To make this change easier to implement, I first did two commits that simply refactor the existing code. These should be reviewed first to ensure that I didn't break anything.

1. The first commit simplifies `originalFailure` from a `SimpleEntry<String,Throwable>` to just a `Throwable`. The key was never used, so this is just a simplification.
2. The second commit changes the saved on-disk state from the first failure with all subsequent failures chained onto it via suppressed exceptions to a `LinkedHashSet` of failures in insertion order. This allows us to delay choosing the top-level exception until later in the execution in `checkAllDone`, which is needed in order to implement the new `propagateWorst` functionality.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -170,7 +170,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pipeline-build-step</artifactId>
-            <version>2.10-rc366.8d263648aa27</version> <!-- TODO: Pending release of https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24 -->
+            <version>2.10</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <revision>2.75</revision>
+        <revision>2.76</revision>
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.138.4</jenkins.version>
         <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
     <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Groovy</name>
-    <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+Groovy+Plugin</url>
+    <url>https://github.com/jenkinsci/workflow-cps-plugin</url>
     <licenses>
         <license>
             <name>MIT License</name>
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.62</version>
+            <version>1.63</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <groovy-cps.version>1.30</groovy-cps.version>
+        <groovy-cps.version>1.31</groovy-cps.version>
         <structs-plugin.version>1.20</structs-plugin.version>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -169,6 +169,12 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-build-step</artifactId>
+            <version>2.10-rc366.8d263648aa27</version> <!-- TODO: Pending release of https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24 -->
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pipeline-input-step</artifactId>
             <version>2.5</version>
             <scope>test</scope>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
@@ -76,7 +76,12 @@ public class ParallelStep extends Step {
         private final boolean failFast;
         /** Have we called stop on the StepExecution? */
         private boolean stopSent = false;
-        /** If we fail fast, we need to record the first failure. Use a linked hash set to maintain insertion order. */
+        /**
+         * If we fail fast, we need to record the first failure.
+         *
+         * <p>We use a set because we may be encountering the same abort being delivered across
+         * branches. We use a linked hash set to maintain insertion order.
+         */
         private final LinkedHashSet<Throwable> failures = new LinkedHashSet<>();
 
         /**
@@ -128,10 +133,6 @@ public class ParallelStep extends Step {
                 } catch (IOException | InterruptedException x) {
                     LOGGER.log(Level.WARNING, null, x);
                 }
-                /*
-                 * Use a set because we may be encountering the same abort being delivered across
-                 * branches.
-                 */
                 handler.failures.add(t);
                 checkAllDone(true);
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
@@ -100,7 +100,7 @@ public class ParallelStep extends Step {
          * Collect the results of sub-workflows as they complete.
          * The key set is fully populated from the beginning.
          */
-        private final Map<String,Outcome> outcomes = new HashMap<String, Outcome>();
+        private final Map<String,Outcome> outcomes = new HashMap<>();
 
         ResultHandler(
                 StepContext context,
@@ -159,7 +159,7 @@ public class ParallelStep extends Step {
             }
 
             private void checkAllDone(boolean stepFailed) {
-                Map<String,Object> success = new HashMap<String, Object>();
+                Map<String,Object> success = new HashMap<>();
                 for (Entry<String,Outcome> e : handler.outcomes.entrySet()) {
                     Outcome o = e.getValue();
 
@@ -284,7 +284,7 @@ public class ParallelStep extends Step {
         public Step newInstance(Map<String,Object> arguments) {
             boolean failFast = false;
             boolean propagateWorst = false;
-            Map<String,Closure<?>> closures = new LinkedHashMap<String, Closure<?>>();
+            Map<String,Closure<?>> closures = new LinkedHashMap<>();
             for (Entry<String,Object> e : arguments.entrySet()) {
                 if ((e.getValue() instanceof Closure)) {
                     closures.put(e.getKey(), (Closure<?>)e.getValue());
@@ -312,7 +312,7 @@ public class ParallelStep extends Step {
 
         @Override public Map<String,Object> defineArguments(Step step) throws UnsupportedOperationException {
             ParallelStep ps = (ParallelStep) step;
-            Map<String,Object> retVal = new TreeMap<String,Object>(ps.closures);
+            Map<String,Object> retVal = new TreeMap<>(ps.closures);
             if (ps.failFast) {
                 retVal.put(FAIL_FAST_FLAG, Boolean.TRUE);
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
@@ -16,7 +16,6 @@ import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
 import java.io.Serializable;
-import java.util.AbstractMap.SimpleEntry;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -72,7 +71,7 @@ public class ParallelStep extends Step {
         /** Have we called stop on the StepExecution? */
         private boolean stopSent = false;
         /** if we failFast we need to record the first failure */
-        private SimpleEntry<String,Throwable> originalFailure = null;
+        private Throwable originalFailure = null;
 
         /**
          * Collect the results of sub-workflows as they complete.
@@ -124,9 +123,9 @@ public class ParallelStep extends Step {
                     LOGGER.log(Level.WARNING, null, x);
                 }
                 if (handler.originalFailure == null) {
-                    handler.originalFailure = new SimpleEntry<String, Throwable>(name, t);
+                    handler.originalFailure = t;
                 } else {
-                    Throwable originalT = handler.originalFailure.getValue();
+                    Throwable originalT = handler.originalFailure;
                     if (t != originalT) { // could be the same abort being delivered across branches
                         originalT.addSuppressed(t);
                     }
@@ -155,7 +154,7 @@ public class ParallelStep extends Step {
                     if (o.isFailure()) {
                         if (handler.originalFailure == null) {
                             // in case the plugin is upgraded whilst a parallel step is running
-                            handler.originalFailure = new SimpleEntry<String, Throwable>(e.getKey(), e.getValue().getAbnormal());
+                            handler.originalFailure = e.getValue().getAbnormal();
                         }
                         // recorded in the onFailure
                     } else {
@@ -164,7 +163,7 @@ public class ParallelStep extends Step {
                 }
                 // all done
                 if (handler.originalFailure!=null) {
-                    handler.context.onFailure(handler.originalFailure.getValue());
+                    handler.context.onFailure(handler.originalFailure);
                 } else {
                     handler.context.onSuccess(success);
                 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
@@ -3,13 +3,16 @@ package org.jenkinsci.plugins.workflow.cps.steps;
 import com.cloudbees.groovy.cps.Outcome;
 
 import groovy.lang.Closure;
+import hudson.AbortException;
 import hudson.Extension;
+import hudson.model.Result;
 import hudson.model.TaskListener;
 import java.io.IOException;
 
 import org.jenkinsci.plugins.workflow.cps.CpsVmThreadOnly;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
@@ -18,6 +21,7 @@ import org.jenkinsci.plugins.workflow.steps.StepExecution;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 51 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
```diff
@@ -517,4 +527,178 @@ public void parallelLexicalScope() throws Exception {
...
+                comparator.compare(
+                        new FlowInterruptedException(
+                                Result.FAILURE, new TimeoutStepExecution.ExceededTimeout()),
+                        new FlowInterruptedException(
+                                Result.UNSTABLE, new TimeoutStepExecution.ExceededTimeout())));
+    }
+
+    @Issue("JENKINS-49073")
+    @Test
+    public void parallelPropagatesStatus() throws Exception {
```
basil: I added these new test cases, which exercise the new functionality and fail prior to this change. Are there any other use cases I should consider or test cases I should write?

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -193,19 +283,27 @@ public String getFunctionName() {
...
             Map<String,Closure<?>> closures = new LinkedHashMap<String, Closure<?>>();
             for (Entry<String,Object> e : arguments.entrySet()) {
                 if ((e.getValue() instanceof Closure)) {
                     closures.put(e.getKey(), (Closure<?>)e.getValue());
                 }
                 else if (FAIL_FAST_FLAG.equals(e.getKey()) && e.getValue() instanceof Boolean) {
                     failFast = (Boolean)e.getValue();
+                } else if (PROPAGATE_WORST_FLAG.equals(e.getKey())
+                        && e.getValue() instanceof Boolean) {
+                    propagateWorst = (Boolean) e.getValue();
```
jglick: Does it make sense to specify _both_ `failFast: true` and `propagateWorst: true`? If not, this should throw an IAE if you try.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -193,19 +283,27 @@ public String getFunctionName() {
...
+                                    + e.getKey()
+                                    + "="
+                                    + e.getValue());
                 }
             }
-            return new ParallelStep((Map)closures, failFast);
+            return new ParallelStep((Map) closures, failFast, propagateWorst);
         }
 
         @Override public Map<String,Object> defineArguments(Step step) throws UnsupportedOperationException {
```
jglick: also needs to be updated I think, and edit `ParallelStep/help.html`

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
```diff
@@ -517,4 +527,178 @@ public void parallelLexicalScope() throws Exception {
             }
         });
     }
+
+    @Test
```
jglick:
```suggestion
    @WithoutJenkins
    @Test
```

pom.xml
```diff
@@ -167,6 +167,12 @@
             <version>1.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-build-step</artifactId>
```
jglick: Unnecessary I think; pending [JENKINS-27092](https://issues.jenkins-ci.org/browse/JENKINS-27092), can use non-sandboxed scripts which directly `throw new FlowInterruptedException(…)` for tests.

src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
basil: > Does it make sense to specify _both_ `failFast: true` and `propagateWorst: true`?

Great point! I don't think this makes sense.

> If not, this should throw an IAE if you try

Done.

pom.xml
basil: Perhaps this isn't necessary, but I think it makes the test more realistic. I saw that we're already pulling in `pipeline-input-step`, so pulling in `pipeline-build-step` as well didn't seem like a big price to pay for the benefit of more realistic integration testing. I think I'd rather keep this as-is, but let me know if you feel strongly otherwise.

jglick: No strong feeling.

basil:
My most recent test run failed due to an issue that doesn't look related to this change:

```
[ERROR] inProgressButProgramLoadFailure(org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest)  Time elapsed: 195.219 s  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 180 seconds
	at java.lang.Thread.sleep(Native Method)
	at org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest.assertCompletedCleanly(PersistenceProblemsTest.java:56)
	at org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest.lambda$inProgressButProgramLoadFailure$19(PersistenceProblemsTest.java:321)
	at org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest$$Lambda$2/1166807841.run(Unknown Source)
	at org.jvnet.hudson.test.RestartableJenkinsRule$3.evaluate(RestartableJenkinsRule.java:200)
	at org.jvnet.hudson.test.RestartableJenkinsRule$6.evaluate(RestartableJenkinsRule.java:240)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:556)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

Restarting to get another build.

basil:
Now all branches of my build failed without even running a single test. I'm just going to give up for today and try to run a build again tomorrow.

slonopotamus:
How `propagateWorst=true` interacts with `failFast=true`?

basil:
> Now all branches of my build failed without even running a single test. I'm just going to give up for today and try to run a build again tomorrow.

I got further today. My test run passed in all environments except in `windows-11-2.164.1`, where the following two errors occurred:

```
[ERROR] Tests run: 11, Failures: 0, Errors: 1, Skipped: 2, Time elapsed: 218.424 s <<< FAILURE! - in org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest
[ERROR] inProgressButProgramLoadFailure(org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest)  Time elapsed: 192.424 s  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 180 seconds
	at java.base@11.0.3/java.lang.Thread.sleep(Native Method)
	at app//org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest.assertCompletedCleanly(PersistenceProblemsTest.java:56)
	at app//org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest.lambda$inProgressButProgramLoadFailure$19(PersistenceProblemsTest.java:321)
	at app//org.jenkinsci.plugins.workflow.cps.PersistenceProblemsTest$$Lambda$67/0x0000000100125c40.run(Unknown Source)
	at app//org.jvnet.hudson.test.RestartableJenkinsRule$3.evaluate(RestartableJenkinsRule.java:200)
	at app//org.jvnet.hudson.test.RestartableJenkinsRule$6.evaluate(RestartableJenkinsRule.java:240)
	at app//org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:556)
	at app//org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at app//org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.base@11.0.3/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base@11.0.3/java.lang.Thread.run(Thread.java:834)
```


```
[ERROR] Tests run: 15, Failures: 1, Errors: 0, Skipped: 4, Time elapsed: 188.762 s <<< FAILURE! - in org.jenkinsci.plugins.workflow.cps.FlowDurabilityTest
[ERROR] testCompleteAndLoadBuilds(org.jenkinsci.plugins.workflow.cps.FlowDurabilityTest)  Time elapsed: 4.283 s  <<< FAILURE!
org.junit.ComparisonFailure: 
expected:<...e] End of Pipeline
[]> but was:<...e] End of Pipeline
[Finished: SUCCESS
]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.jenkinsci.plugins.workflow.cps.FlowDurabilityTest$2.evaluate(FlowDurabilityTest.java:420)
	at org.jvnet.hudson.test.RestartableJenkinsRule$6.evaluate(RestartableJenkinsRule.java:240)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:556)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

I can't reproduce these locally and they don't appear to be related to my changes. I'll leave it to the maintainers to comment on any possible next steps here.

basil:
> How `propagateWorst=true` interacts with `failFast=true`?

The two are mutually exclusive and an `IllegalArgumentException` is thrown if one attempts to use both simultaneously (see f1c19685c71f63470e5b2d8fc312f97e895e927d).

> I'd like `propagateWorst=true` to be default. What is the procedure of discussing this design decision?

I would also like it to be the default, but I erred on the conservative side in this change. There is precedent for changing existing behavior (and breaking backward compatibility!) to adopt a more sensible default in [jenkinsci/pipeline-build-step-plugin#24 (comment)](https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24#issuecomment-532671888):

> though I think the behavior in this PR is a natural default

Given this past precedent, it seems _possible_ to change the default behavior of the `parallel` step to propagate the worst result by default, and (to me) this seems like a "natural default", but I defer to the judgment of the maintainers regarding this matter.

dwnusbaum:
@basil Before https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24, was there any normal scenario in which users would care about the difference between `propagate: true` vs `propagate: false` ? The only one I can think of is for users manually throwing `FlowInterruptedException`, but I don't know how common that is. I am thinking it would be better to just make `propagate: true` the default so we don't have another option, but that depends on whether users care about the previous behavior.

basil:
> Before jenkinsci/pipeline-build-step-plugin#24, was there any normal scenario in which users would care about the difference between `propagate: true` vs `propagate: false`?

I think you meant to ask if there was any normal scenario in which users would care about the difference between `propagateWorst: true` vs `propagateWorst: false` in the `parallel` step (i.e., what I am proposing in this PR) rather than `propagate: true` and `propagate: false` in the `build` step (which is what was affected in jenkinsci/pipeline-build-step-plugin#24). So I'll answer the question I think you meant to ask. But if you really did mean to ask about the behavior of the `build` step, let me know and we can discuss that as well.

I took a look at the 116 Pipeline jobs in use at my company, and I didn't find a single instance where the author manually threw a `FlowInterruptedException`. In all the cases where we are using `parallel` steps (without `failFast: true`), we had initially expected the worst result to be propagated and were surprised when it wasn't. To work around this, we wrote the code described in [JENKINS-49073 (comment)](https://issues.jenkins-ci.org/browse/JENKINS-49073?focusedCommentId=375702&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-375702). This is why I think propagating the worst result would be a natural default (when `failFast: true` is _not_ being used).

Based on the above, I think we have a few options:

1. The most conservative option is to preserve the existing behavior (propagate the first result) by default, add a new option to propagate the worst result (off by default), and fail with an `IllegalStateException` when `propagateWorst` is used with `failFast`.
2. The simplest option is to propagate the worst result by default (a change in existing behavior), except when `failFast` is set. No new syntax would be added to the API. When `failFast` is set, we'd continue to propagate the first result as we do today, but there would no longer be a way to propagate the first result if `failFast` is _not_ set.
3. A compromise would be to propagate the worst result by default (a change in existing behavior), except when `failFast` is set, but to _also_ add a new `propagateFirst` option, which could be used if you want to get the old behavior of propagating the first result _without_ setting `failFast`.

I personally don't think there is a real use case for propagating the first result _without_ setting `failFast`. At least I have never wanted this in any of my pipelines. So I dislike option 3 because it unnecessarily complicates the API. So for me the real choice is between option 1 and option 2. Since I think propagating the worst result by default (except when `failFast` is set) is the most natural default, I personally favor option 2 (knowing that it breaks backward compatibility). Another reason to favor option 2 is that it keeps the syntax simple and avoids having to throw an `IllegalArgumentException` when `failFast` is used with `propagateWorst`.

dwnusbaum:
@basil I think option 2 seems like the most obvious behavior. Unfortunately, we have no real way to know how many users might be relying on the existing behavior, but given that users would have to be manually throwing `FlowInterruptedException` to notice a difference, it seems safe enough to change the default behavior, so my vote is for option 2.

basil:
Sounds good to me! I've updated the change to propagate the worst result by default when `failFast` is `false` and removed the `propagateWorst` setting.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
```diff
@@ -517,4 +528,193 @@ public void parallelLexicalScope() throws Exception {
...
+                parallelPropagatesStatusImpl(
+                        Result.FAILURE, Result.SUCCESS, Result.FAILURE, Result.UNSTABLE);
+                parallelPropagatesStatusImpl(
+                        Result.NOT_BUILT, Result.SUCCESS, Result.NOT_BUILT);
+                parallelPropagatesStatusImpl(
+                        Result.NOT_BUILT, Result.SUCCESS, Result.UNSTABLE, Result.NOT_BUILT);
+                parallelPropagatesStatusImpl(
+                        Result.NOT_BUILT, Result.SUCCESS, Result.UNSTABLE, Result.FAILURE, Result.NOT_BUILT);
+                parallelPropagatesStatusImpl(
+                        Result.NOT_BUILT, Result.NOT_BUILT, Result.NOT_BUILT);
```
basil: I'm not entirely satisfied with how the code is currently working for these test cases. For example, for three parallel builds whose results are `SUCCESS`, `UNSTABLE`, and `NOT_BUILT` we are currently propagating a status of `NOT_BUILT`. I don't consider this ideal; to me, `UNSTABLE` is "worse" than `NOT_BUILT`. Similarly, for four parallel builds whose results are `SUCCESS`, `UNSTABLE`, `FAILURE`, and `NOT_BUILT` we are currently propagating a status of `NOT_BUILT. I don't consider this ideal; to me, `FAILURE` is "worse" than `NOT_BUILT`. Finally, for two parallel builds whose status is `SUCCESS` and `NOT_BUILT` we are currently propagating a status of `NOT_BUILT`. This doesn't make sense to me. If all the child jobs had a status of `NOT_BUILT`, then I could understanding the parent job having a status of `NOT_BUILT`, but as long as a single child job has a status of `SUCCESS`, I would expect that to be propagated to the parent.

I will think some more about how to tweak the comparator algorithm to handle these cases.

basil: Another tricky question is how to handle `ABORTED`. Between `SUCCESS` and `ABORTED`, `ABORTED` is clearly worse. But what about between `FAILURE`/`UNSTABLE` and `ABORTED`? I would argue that `FAILURE`/`UNSTABLE` are "worse", because we _know_ the status of the failed/unstable job (whereas we _don't_ know the status of the aborted job, which never ran to completion). Between `NOT_BUILT` and `ABORTED` I think `ABORTED` is worse.

dwnusbaum: Re: What result is worst, I would just use the ordering defined by `Result.ordinal` (used to implement `Result.isWorseThan` and friends). Whether that ordering makes the most sense could be debated, but that's the ordering that everything else in Jenkins uses, so I would use it here as well.

Might be interesting to run pipeline-model-definition's tests with this change to see if any of them start failing.

basil: > Re: What result is worst, I would just use the ordering defined by `Result.ordinal` (used to implement `Result.isWorseThan` and friends). Whether that ordering makes the most sense could be debated, but that's the ordering that everything else in Jenkins uses, so I would use it here as well.

My comparison logic is already using `Result#isWorseThan` and `Result#isBetterThan`, so if we're OK with the behavior demonstrated by the new tests (which I think is acceptable, but not ideal) then I can leave the comparison logic as-is.

> Might be interesting to run pipeline-model-definition's tests with this change to see if any of them start failing.

Sure. I opened jenkinsci/pipeline-model-definition-plugin#360 to cover this testing.

basil: > I opened [jenkinsci/pipeline-model-definition-plugin#360](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/360) to cover this testing.

These tests passed.

dwnusbaum: To elaborate on my earlier comment, the values in `Result` have always seemed weird to me. `ABORTED` and `NOT_BUILT` feel like a different kind of thing entirely than `SUCCESS`, `FAILURE`, and `UNSTABLE`. Sometimes I think it would have made more sense for results to have 2 parts, a "completion" result, like , `NOT_BUILT`, `ABORTED`, `COMPLETED`, and a "goodness" result, like `SUCCESS`, `WARNING`, `ERROR`. Then you could have something like a Pipeline step that is `ABORTED` and `SUCCESS` in the case of a user cancelling the build, or `ABORTED` and `WARNING` in the case of something like a timeout, `NOT_BUILT` and `SUCCESS` could be used in cases like a Declarative `when` expression that is false, and here in the parallel step, the completion result would always be `COMPLETED`, and only the "goodness" result would be determined by the results of the branches. Using two values creates other problems though, and not all combinations make sense.

So all that to say, I agree that the behavior here with `NOT_BUILT` feels wrong, but I'm not sure if it makes sense to try to change it.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
```diff
@@ -517,4 +528,193 @@ public void parallelLexicalScope() throws Exception {
...
+                                "%s #1 completed with status %s",
+                                downstreamJob.getName(), downstreamResult.toString()),
+                        run);
+            }
+        }
+    }
+
+    private FreeStyleProject getDownstreamJob(Result result) throws IOException {
+        FreeStyleProject ds = story.j.createFreeStyleProject();
+        ds.getBuildersList().add(new TestBuilder() {
```
dwnusbaum: nit: Maybe simpler to use a Pipeline job for the downstream builds instead:

```
WorkflowJob ds = story.j.createProject(WorkflowJob.class);
ds.setDefinition(new CpsFlowDefinition("currentBuild.result = " + result, true));
```

pom.xml
dwnusbaum: Personally, I'd lean towards the `FlowInterruptedException` approach since it would make `parallelPropagatesStatusImpl` much easier to read, but I think either approach is fine.

dwnusbaum:
@basil My suggestions were a bit messed up and incomplete, so I filed https://github.com/basil/workflow-cps-plugin/pull/1 for improvements, it would be great if you could take a look at that PR and merge it in if you are ok with the changes.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
basil: I agree with your observations here. The awkwardness in the use cases I described above is exposing a deeper problem in how the `Result` API is defined in Jenkins core. When I get a chance I'll try to file a bug/RFE against Jenkins core with the results of this discussion. Perhaps one day that can be addressed, but I agree that it is quite orthogonal to this effort.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
basil: I don't think it makes a big difference one way or the other. I did it this way simply because I was copying an existing example from the `pipeline-build-step` plugin's tests.

basil:
> @basil My suggestions were a bit messed up and incomplete, so I filed [basil#1](https://github.com/basil/workflow-cps-plugin/pull/1) for improvements. It would be great if you could take a look at that PR and merge it in if you are ok with the changes.

Thanks for helping me take this across the finish line! At this point I think the only remaining change we need to make is to release `pipeline-build-step` and resolve the TODO in `pom.xml` in this PR by pulling in the released version.

pom.xml
```diff
@@ -167,6 +167,12 @@
...
+            <version>2.10-rc366.8d263648aa27</version> <!-- TODO: Pending release of https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24 -->
```
dwnusbaum: @basil I just released pipeline-build-step.

```suggestion
            <version>2.10</version>
```

PR #328, opened 2019-10-13
JENKINS-59762 Replace static initialization block to allow for CpsThread mocking

kamilszymanski:
This change makes it possible to test scripted pipelines (e.g. with jenkins-spock library).
Without it any mocking attempt results in:
```
Could not initialize class org.jenkinsci.plugins.workflow.cps.CpsThread
java.lang.NoClassDefFoundError: Could not initialize class org.jenkinsci.plugins.workflow.cps.CpsThread
	at org.jenkinsci.plugins.workflow.cps.CpsScript.<init>(CpsScript.java:69)
	at org.spockframework.mock.runtime.MockInstantiator.instantiate(MockInstantiator.java:33)
	at org.spockframework.mock.runtime.ProxyBasedMockFactory$CglibMockFactory.createMock(ProxyBasedMockFactory.java:155)
	at org.spockframework.mock.runtime.ProxyBasedMockFactory.create(ProxyBasedMockFactory.java:68)
	at org.spockframework.mock.runtime.JavaMockFactory.createInternal(JavaMockFactory.java:59)
	at org.spockframework.mock.runtime.JavaMockFactory.create(JavaMockFactory.java:40)
	at org.spockframework.mock.runtime.CompositeMockFactory.create(CompositeMockFactory.java:44)
	at org.spockframework.lang.SpecInternals.createMock(SpecInternals.java:51)
	at org.spockframework.lang.SpecInternals.createMockImpl(SpecInternals.java:296)
	at org.spockframework.lang.SpecInternals.createMockImpl(SpecInternals.java:286)
	at org.spockframework.lang.SpecInternals.SpyImpl(SpecInternals.java:161)
	at com.homeaway.devtools.jenkins.testing.JenkinsPipelineSpecification.setup(JenkinsPipelineSpecification.groovy:1025)
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
@@ -159,13 +159,6 @@ public final class CpsThread implements Serializable {
         this.step = step;
     }
 
-    private static final List<Class> categories;
-    static {
-        categories = new ArrayList<>();
-        categories.addAll(Continuable.categories);
-        categories.add(IteratorHack.class);
-    }
-
     /**
      * Executes CPS code synchronously a little bit more, until it hits
      * the point the workflow needs to be dehydrated.
@@ -183,7 +176,7 @@ public final class CpsThread implements Serializable {
             LOGGER.log(FINE, "runNextChunk on {0}", resumeValue);
             final Outcome o = resumeValue;
             resumeValue = null;
-            outcome = program.run0(o, categories);
+            outcome = program.run0(o, getCategories());
             if (outcome.getAbnormal() != null) {
                 LOGGER.log(FINE, "ran and produced error", outcome.getAbnormal());
             } else {
@@ -329,6 +322,12 @@ public final class CpsThread implements Serializable {
     @CpsVmThreadOnly
     public static CpsThread current() {
         return CURRENT.get();
+    }
+
+    private static List<Class> getCategories() {
+        List<Class> categories = new ArrayList<>(Continuable.categories);
+        categories.add(IteratorHack.class);
+        return categories;
     }
 
     @Override public String toString() {


kamilszymanski:
@bitwiseman I'm not convinced that we should add dependency to any mocking / bytecode generation library for the sake of testing something that is not related to the functionality provided by the plugin but rather a design / implementation detail.
Nevertheless it can be tested externally, e.g. I've replicated the issue in kamilszymanski/jenkins-pipeline-automation-starter@b2f45d7d620c7d87340c7e3c4c413a5187fd15fb (JENKINS-59762 branch in that repo) and then made sure that the locally released snapshot containing this patch is working correctly.
Does the above reasoning meets you needs?

bitwiseman:
If there is an existing test that covers this area that you can point to that still works that's fine.  Is there no way to make a unit test in this repo?  ("No" is a possible answer, I'm just asking you to consider if there is a way to touch this code within this library's tests.) Where possible I prefer to have something that ensures we don't regress functionality in the future. 

Why would adding a mocking framework be a bad thing?  (I'm just asking.)

kamilszymanski:
Expected behaviors, that depend on the changed code, are already tested in existing test suite (CpsThreadTest, CpsThreadDumpTest, CpsThreadDumpActionTest and maybe even few other ones).
Following language you've used (behavior, regress functionality) I assume we both do not see structure worth testing.
As for verifying whether the CpsThread can be mocked - it can be tested, but I'm not convinced that such test should be part of the workflow-cps-plugin test suite.
Adding test dependencies, e.g. a mocking library, is not bad, but given the above I'd say it wouldn't be justified - it creates maintenance overhead, negligible at that moment (but who knows where it could evolve) giving almost nothing (if anything) in return.

BTW. change checks have failed for the latest changes for the reasons unrelated to the changes at all (JNLP connection failure) - are you able to re-run them?

dwnusbaum:
@kamilszymanski I agree that it's not ideal to add dependencies here just to add a test for mocking `CpsThread`. Can you link to a separate PR in some other repository that depends on this version of workflow-cps that shows that whatever you are trying to accomplish works with this change?

> BTW. change checks have failed for the latest changes for the reasons unrelated to the changes at all (JNLP connection failure) - are you able to re-run them?

Unfortunately, very few users have elevated access to ci.jenkins.io to trigger a build directly. You can trigger a rebuild indirectly by closing the PR, waiting a few seconds, and then reopening it.

kamilszymanski:
@dwnusbaum I'll link a small project whose tests rely on the mentioned behavior. The latest commit (kamilszymanski/jenkins-pipeline-automation-starter@97b5304527e744534c21f13cefbe8d081e8a2380) on the JENKINS-59762 branch runs on the plugin version built by [jitpack.io](https://jitpack.io/) with the patch from this PR and passes all the tests. Tests for preceding commit (kamilszymanski/jenkins-pipeline-automation-starter@b2f45d7) fail due to the mentioned CpsThread mocking issues.
Test suite can be run by executing `./gradlew clean test`

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -183,7 +176,7 @@ public StepExecution getStep() {
             LOGGER.log(FINE, "runNextChunk on {0}", resumeValue);
             final Outcome o = resumeValue;
             resumeValue = null;
-            outcome = program.run0(o, categories);
+            outcome = program.run0(o, getCategories());
```
dwnusbaum: Instead of having to recreate the list over and over, maybe you could replace the static initialization block with something like this instead?
```
private static final List<Class> categories = ImmutableList.builder().addAll(Continuable.categories).add(IteratorHack.class).build();
```

kamilszymanski: @dwnusbaum it would not fix the underlying problem. Nevertheless I've found a way to workaround the root problem by explicitly managing the classpath (not a perfect solution though) on the call site.

kamilszymanski:
It turns out that the issues I had with mocking were related to the classpath issues, since the static block relies on static properties from another library, and hence can be resolved by managing dependencies manually.
Nevertheless I'd argue that it would be worth fixing so that people don't get confused when they run into such problem (the stacktraces do not point directly to the root problem and the naming conventions are a bit misleading).
The downside of this path is that the categories list is created each time it's requested, however the performance/resource costs of such operation are negligible.
If the mentioned downside is problematic for anyone I've created another PR, where I just fix the design (and naming convention) by getting rid off the static initialization block: #336

dwnusbaum:
> It turns out that the issues I had with mocking were related to the classpath issues, since the static block relies on static properties from another library, and hence can be resolved by managing dependencies manually.

By classpath issues, do you mean that the problem is that you needed to add a dependency on `groovy-cps` somewhere in your build configuration?

I am a little confused, because from the stack trace looks like you are using https://github.com/homeaway/jenkins-spock, which I understand was designed exactly for this use case and seems to be currently working for its users (or at least I can't find this error in any reported issues or PRs on that repo).

kamilszymanski:
@dwnusbaum you're right mentioning groovy-cps, however in my case it was already on a classpath but in a version that was missing some of the referenced static fields.
Such issues might have never been reported since Gradle, which I'm using in the project where I stumbled on this issue, is not a popular choice and most people (according to what I've googled) that have tried to use it failed to do so (e.g.: https://github.com/homeaway/jenkins-spock/issues/22).
It might also be that the problem manifests itself since I'm using jenkins-pipeline-unit in the same project.
I haven't dug further into problem sources once I've found the root cause.

dwnusbaum:
@kamilszymanski Maybe Gradle has something like maven-enforcer that you could use to make sure all of your dependencies are ok? Anyway, thanks for the PR! I'll close this one and merge #336.

PR #329, opened 2019-10-14
JENKINS-48431 Support both lightweight checkout AND build parameters

tgatinea:
[JENKINS-48431](https://issues.jenkins-ci.org/browse/JENKINS-48431) To support both lightweight checkout AND build parameters, we need to:
- recover the environment parameters of the build
- evaluate the SCM structure according to the build environment parameters

This leads to implement a new method :
public static SCMFileSystem of(@NonNull Run build, @NonNull SCM scm, @CheckForNull SCMRevision rev)

This new method offer the capacity to recover the environments parameter of the build as the current build is propagated through the new method.

Note that this modification is in relation with two other pull requests that concerns:
- scm-api-plugin: https://github.com/jenkinsci/scm-api-plugin/pull/78
- git-plugin: https://github.com/jenkinsci/git-plugin/pull/772

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-lightweight.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-lightweight.html
@@ -2,6 +2,4 @@
     If selected, try to obtain the Pipeline script contents directly from the SCM without performing a full checkout.
     The advantage of this mode is its efficiency; however, you will not get any changelogs or polling based on the SCM.
     (If you use <code>checkout scm</code> during the build, this will populate the changelog and initialize polling.)
-    Also build parameters will not be substituted into SCM configuration in this mode.
-    Only selected SCM plugins support this mode.
 </div>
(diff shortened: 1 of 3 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-lightweight.html
```diff
@@ -2,6 +2,4 @@
     If selected, try to obtain the Pipeline script contents directly from the SCM without performing a full checkout.
     The advantage of this mode is its efficiency; however, you will not get any changelogs or polling based on the SCM.
     (If you use <code>checkout scm</code> during the build, this will populate the changelog and initialize polling.)
-    Also build parameters will not be substituted into SCM configuration in this mode.
-    Only selected SCM plugins support this mode.
```
dwnusbaum, line 6: I think it is important to still note that not all SCMs that support lightweight checkouts, and that even the ones that do support lightweight checkout do not all support build parameters, so maybe change the wording to something like this instead:

```
Only selected SCM plugins support this mode.
Build parameters may be substituted into SCM configuration in this mode when running SCM API plugin X.Y.Z or newer. Not all SCM plugins support this feature.
```

jglick:
Superseded by #577?

dwnusbaum:
> Superseded by https://github.com/jenkinsci/workflow-cps-plugin/pull/577?

Yeah looks like it, so I am closing this PR.

PR #335, opened 2019-11-06
[JENKINS-58878] Make DSL$ThreadTaskImpl.eval and CpsBodyInvoker.start atomic with respect to both CpsStepContext.bodyInvokers and CpsStepContext.sync

dwnusbaum:
See [JENKINS-58878](https://issues.jenkins-ci.org/browse/JENKINS-58878) and https://github.com/jenkinsci/workflow-step-api-plugin/pull/49.

The problem AIUI was that `CpsBodyInvoker.start` could try to synchronously add a body invoker after `DSL$ThreadTaskImpl.invokeBody` had completed (or in the middle of it, leading to things like the `ArrayIndexOutOfBoundsException` described in [this comment](https://issues.jenkins-ci.org/browse/JENKINS-58878?focusedCommentId=378651&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-378651)), but before the step has switched to async mode by calling `CpsStepContext.switchToAsyncMode` in `DSL$ThreadTaskImpl.eval`, and so the body invoker would be ignored forever, since the step had switched to async mode. This only appears to  happen for implementations of `GeneralizedNonBlockingStepExecution` where the body executes very quickly (e.g. Uses of `withCredentials` where the list of credentials is empty).

The fix in this PR is to make sure that operations involving both of `CpsStepContext.bodyInvokers` and `CpsStepContext.sync` are atomic, so that `CpsBodyInvoker.start` does not try to add a synchronous body after `bodyInvokers` has been cleared but before the step has switched to async mode.

I'll update https://github.com/jenkinsci/workflow-step-api-plugin/pull/49 to pull in an incremental build from this PR and unignore the test once I have an incremental build. I'm not sure if it makes sense to add tests here, https://github.com/jenkinsci/workflow-step-api-plugin/pull/49 has a regression test for the specific bug (we could move it here instead if desired), and the existing tests in this plugin verify that this didn't break any existing behavior.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyInvoker.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyInvoker.java
@@ -136,12 +136,17 @@ public final class CpsBodyInvoker extends BodyInvoker {
             throw new IllegalStateException("The " + owner.getDisplayName() + " step has already completed.");
         }
 
-        if (owner.isSyncMode()) {
-            // we call 'launch' later from DSL.ThreadTaskImpl.
-            // in this mode, the first thread inherits the same thread, but
-            // all the other body executions are run as new threads, for the parallel.
-            owner.bodyInvokers.add(this);
-        } else {
+        boolean addedSynchrously = owner.withBodyInvokers(bodyInvokers -> {
+            if (owner.isSyncMode()) {
+                // we call 'launch' later from DSL.ThreadTaskImpl.
+                // in this mode, the first thread inherits the same thread, but
+                // all the other body executions are run as new threads, for the parallel.
+                bodyInvokers.add(this);
+                return true;
+            }
+            return false;
+        });
+        if (!addedSynchrously) {
             // when this method is called asynchronously, the body is scheduled to run in the same thread
             // that started run.
             try {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
@@ -57,12 +57,12 @@ import javax.annotation.concurrent.GuardedBy;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.CauseOfInterruption;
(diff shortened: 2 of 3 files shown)


dwnusbaum:
> In this case, in non-synch mode, we are already using runInCpsVmThread. Can we not just do the same in synch mode, updating bodyInvokers later inside the CPS VM thread

Yes, but we do not enter non-sync mode until after `DSL$ThreadTaskImpl.invokeBody` completes, so when `CpsStepContext.newBodyInvoker().start()` is called from a background thread by a `GeneralizedNonBlockingStepExecution`, `CpsBodyInvoker.start` thinks the step is in sync mode, but really, `DSL$ThreadTaskImpl.invokeBody` is in the middle of being executed, and will switch to async mode once it completes, so `CpsBodyInvoker` adds the body invoker synchronously, but too late for the new body invoker to actually be executed synchronously. 

It seems like making `CpsStepContext.start` modify `CpsStepContext.bodyInvokers` on the VM thread even in the sync case would have the same problem, since you would need to make sure that `DSL$ThreadTaskImpl.invokeBody` runs after `CpsBodyInvoker.start` adds the body invoker to `CpsStepContext.bodyInvokers`, but I don't know of a way to enforce that ordering.

If you mean to always use the async code path that uses the VM thread from `CpsBodyInvoker`, I thought about that, but it seemed complicated. I'm not totally clear on why we handle the sync and non-sync cases differently, so I was wondering if the sync case was just an optimization, and we could drop `CpsBodyInvoker` entirely and just always use the async path, but then I'm not sure what the replacement for `DSL$ThreadTaskImpl.invokeBody` would be.

bitwiseman:
@dwnusbaum
I can honestly say I don't know enough to review this.  
But if you have a minute to chat and give me some context I'm sure I can pick it up.

dwnusbaum:
@jglick FWIW, I tried changing `CpsBodyInvoker.start` to run the sync logic on the CPS VM thread as in the non-sync case, and it mostly works, but breaks tests in at least `CpsBodyInvokerTest` and `ParallelTest` for reasons that I don't really understand (in general, the problems appear to be related to not enough `FlowHead`s and/or `CpsThread`s being created, for example, `CpsThread.current()` returning null while on the VM thread leading to an NPE).

If you feel strongly that we should go down that route, any pointers you have would be appreciated. I can open a second PR with those changes if you want to see what I tried.

@bitwiseman Sure I'll set something up, although I can't say I really understand this code myself 😄

jglick:
> I was wondering if the sync case was just an optimization, and we could drop `CpsStepContext.bodyInvokers` entirely and just always use the async path

Possibly, though this might cause some test failures.

> If you feel strongly that we should go down that route

No, just wanted to check if it was something you had thought about and explicitly rejected. It sounds like you have, so just go ahead with whatever works.

dwnusbaum:
> > I was wondering if the sync case was just an optimization, and we could drop CpsStepContext.bodyInvokers entirely and just always use the async path
>
> Possibly, though this might cause some test failures.

I ended up trying that as well, but my initial approach broke all tests with block-scope steps because it made them hang indefinitely, so I didn't spend much time looking into it.

> No, just wanted to check if it was something you had thought about and explicitly rejected. It sounds like you have, so just go ahead with whatever works.

Ok, thanks for taking a look!

PR #340, opened 2019-12-02
[JENKINS-34256] Resume Pipeline execution if quiet mode is canceled

dwnusbaum:
See [JENKINS-34256](https://issues.jenkins-ci.org/browse/JENKINS-34256).

I am not sure if this is the only problem described in that ticket, since there is some discussion there about specific steps, but as far as I can tell the problem was that once the `CpsThreadGroup` suspended itself because it detected that quiet mode was enabled, nothing will call `CpsThreadGroup.scheduleRun` other than pausing and unpausing the Pipeline or restarting Jenkins, so the build is just permanently suspended.

This PR fixes the issue by adding a listener that tracks `Jenkins.isQuietingDown` to detect when quiet down mode is cancelled so that it can call `CpsThreadGroup.scheduleRun` which will resume the build.

I'm not sure if there is a better way to do this, or if the current behavior is even the behavior we want. Why not just allow the Pipelines to continue executing so that suspension/resumption works as normal when Jenkins is restarted? **EDIT**: Jesse discusses the reasoning behind the current behavior in [JENKINS-32015](https://issues.jenkins-ci.org/browse/JENKINS-32015).

Opening a draft PR for now while I read through the ticket to see if I am misunderstanding the issue.

**EDIT**: I read through all the comments, including the ones on [JENKINS-38316](https://issues.jenkins-ci.org/browse/JENKINS-38316) and found [this comment](https://issues.jenkins-ci.org/browse/JENKINS-38316?focusedCommentId=313473&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-313473) from Jesse, which is essentially what I am doing here for the fix. I think this fix is the root cause of all the issues discussed in the ticket, along with confusion from users on what "Prepare for shutdown" does for Pipeline jobs.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -2112,7 +2112,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                             }
                             @Override
                             public void onFailure(Throwable x) {
-                                LOGGER.log(Level.WARNING, "cannot resume " + exec + " after cancelling quiet mode", x);
+                                LOGGER.log(Level.WARNING, "cannot resume " + exec + " after canceling quiet mode", x);
                             }
                         });
                     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -2103,17 +2103,16 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 // Need to resume all executions by calling `CpsThreadGroup.scheduleRun()`.
                 for (FlowExecution exec : FlowExecutionList.get()) {
                     if (exec instanceof CpsFlowExecution) {
-                        CpsFlowExecution cpsExec = (CpsFlowExecution) exec;
-                        cpsExec.runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
+                        ((CpsFlowExecution) exec).runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
                             @Override
                             public void onSuccess(CpsThreadGroup g) {
-                                if (!g.isPaused()) {
+                                if (!g.isPaused()) { // If the CpsThreadGroup is paused, scheduleRun will cause it to be persisted again unnecessarily.
                                     g.scheduleRun();
                                 }
                             }
                             @Override
                             public void onFailure(Throwable x) {
-                                LOGGER.log(Level.WARNING, "cannot resume from quiet down " + exec, x);
+                                LOGGER.log(Level.WARNING, "cannot resume " + exec + " after cancelling quiet mode", x);
                             }
                         });
                     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -110,9 +110,11 @@ import groovy.lang.GroovyCodeSource;
 import hudson.AbortException;
 import hudson.BulkChange;
 import hudson.Extension;
+import hudson.Main;
 import hudson.init.Terminator;
 import hudson.model.Item;
 import hudson.model.Job;
+import hudson.model.PeriodicWork;
 import hudson.model.Queue;
 import hudson.model.Run;
 import hudson.model.Saveable;
@@ -2078,6 +2080,47 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     @Override
     protected void notifyShutdown() {
         // No-op, handled in the suspendAll terminator
+    }
+
+    // TODO: Core could use a QuietDownListener class with `onQuietDown` and `onCancelQuietDown` methods.
+    @Restricted(NoExternalUse.class)
+    @Extension
+    public static class QuietDownListener extends PeriodicWork {
+        private boolean wasQuietDown = Jenkins.get().isQuietingDown();
+
+        @Override
+        public long getRecurrencePeriod() {
+            return Main.isUnitTest
+                    ? TimeUnit.SECONDS.toMillis(1)
+                    : TimeUnit.MINUTES.toMillis(1);
+        }
+
+        @Override
+        protected void doRun() throws Exception {
+            boolean isQuietDown = Jenkins.get().isQuietingDown();
+            // Quieting down is handled by `CpsThreadGroup.scheduleRun`, so we just need to handle quiet mode being cancelled.
+            if (!isQuietDown && wasQuietDown) {
+                // Need to resume all executions by calling `CpsThreadGroup.scheduleRun()`.
+                for (FlowExecution exec : FlowExecutionList.get()) {
+                    if (exec instanceof CpsFlowExecution) {
+                        CpsFlowExecution cpsExec = (CpsFlowExecution) exec;
+                        cpsExec.runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
+                            @Override
+                            public void onSuccess(CpsThreadGroup g) {
+                                if (!g.isPaused()) {
+                                    g.scheduleRun();
+                                }
+                            }
+                            @Override
+                            public void onFailure(Throwable x) {
+                                LOGGER.log(Level.WARNING, "cannot resume from quiet down " + exec, x);
+                            }
+                        });
+                    }
+                }
+            }
+            wasQuietDown = isQuietDown;
+        }
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -110,11 +110,9 @@ import groovy.lang.GroovyCodeSource;
 import hudson.AbortException;
 import hudson.BulkChange;
 import hudson.Extension;
-import hudson.Main;
 import hudson.init.Terminator;
 import hudson.model.Item;
 import hudson.model.Job;
-import hudson.model.PeriodicWork;
 import hudson.model.Queue;
 import hudson.model.Run;
 import hudson.model.Saveable;
@@ -2080,46 +2078,6 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     @Override
     protected void notifyShutdown() {
         // No-op, handled in the suspendAll terminator
-    }
-
-    // TODO: Core could use a QuietDownListener class with `onQuietDown` and `onCancelQuietDown` methods.
-    @Restricted(NoExternalUse.class)
-    @Extension
-    public static class QuietDownListener extends PeriodicWork {
-        private boolean wasQuietDown = Jenkins.get().isQuietingDown();
-
-        @Override
-        public long getRecurrencePeriod() {
-            return Main.isUnitTest
-                    ? TimeUnit.SECONDS.toMillis(1)
-                    : TimeUnit.MINUTES.toMillis(1);
-        }
-
-        @Override
-        protected void doRun() throws Exception {
-            boolean isQuietDown = Jenkins.get().isQuietingDown();
-            // Quieting down is handled by `CpsThreadGroup.scheduleRun`, so we just need to handle quiet mode being cancelled.
-            if (!isQuietDown && wasQuietDown) {
-                // Need to resume all executions by calling `CpsThreadGroup.scheduleRun()`.
-                for (FlowExecution exec : FlowExecutionList.get()) {
-                    if (exec instanceof CpsFlowExecution) {
-                        ((CpsFlowExecution) exec).runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
-                            @Override
-                            public void onSuccess(CpsThreadGroup g) {
-                                if (!g.isPaused()) { // If the CpsThreadGroup is paused, scheduleRun will cause it to be persisted again unnecessarily.
-                                    g.scheduleRun();
-                                }
-                            }
-                            @Override
-                            public void onFailure(Throwable x) {
-                                LOGGER.log(Level.WARNING, "cannot resume " + exec + " after canceling quiet mode", x);
-                            }
-                        });
-                    }
-                }
-            }
-            wasQuietDown = isQuietDown;
-        }
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -296,11 +296,14 @@ public final class CpsThreadGroup implements Serializable {
                             } catch (IOException e) {
                                LOGGER.log(Level.WARNING, null, e);
                             }
-                            Timer.get().schedule(() -> {
-                                if (j.isQuietingDown()) {
-                                    Timer.get().schedule(this, Main.isUnitTest ? 1 : 10, TimeUnit.SECONDS);
-                                } else {
-                                    scheduleRun();
+                            Timer.get().schedule(new Runnable() {
+                                @Override
+                                public void run() {
+                                    if (j.isQuietingDown()) {
+                                        Timer.get().schedule(this, Main.isUnitTest ? 1 : 10, TimeUnit.SECONDS);
+                                    } else {
+                                        scheduleRun();
+                                    }
                                 }
                             }, Main.isUnitTest ? 1 : 10, TimeUnit.SECONDS);
                         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -40,9 +40,11 @@ import groovy.lang.GroovyShell;
 import groovy.lang.Script;
 import hudson.ExtensionList;
 import hudson.Functions;
+import hudson.Main;
 import hudson.Util;
 import hudson.model.Result;
 import jenkins.model.Jenkins;
+import jenkins.util.Timer;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
@@ -67,6 +69,7 @@ import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -121,6 +124,13 @@ public final class CpsThreadGroup implements Serializable {
 
     /** Set while {@link #runner} is doing something. */
     transient boolean busy;
+
+    /**
+     * True if the build was automatically paused because quiet mode is enabled.
+     * Used to avoid printing more than one pause message or scheduling more than one resumption task per build.
+     * Independent of {@link #paused}.
+     */
+    private transient AtomicBoolean pausedByQuietMode;
 
     /**
      * True if the execution suspension is requested.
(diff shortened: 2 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -241,6 +241,53 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
...
+            ((CpsFlowExecution) b.getExecution()).waitForSuspension();
+            assertTrue(b.isBuilding());
+            r.assertLogNotContains("I am done", b);
+            r.jenkins.doCancelQuietDown();
+            r.assertLogContains("I am done", r.assertBuildStatusSuccess(r.waitForCompletion(b)));
+        });
+    }
+
+    @Issue("JENKINS-34256")
+    @Test public void pauseAndQuietDown() {
```
dwnusbaum: Not sure about this test, probably would need to be tweaked to avoid race conditions. The goal was to test that canceling quiet mode doesn't also unpause Pipelines.

dwnusbaum: Removed the test. It was hard to follow, and I couldn't think of a good way to make it more robust without inspecting internal execution state in loops with sleep.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -2080,4 +2082,44 @@ protected void notifyShutdown() {
...
+    @Restricted(NoExternalUse.class)
+    @Extension
+    public static class QuietDownListener extends PeriodicWork {
+        private boolean wasQuietDown = Jenkins.get().isQuietingDown();
+
+        @Override
+        public long getRecurrencePeriod() {
+            return Main.isUnitTest
+                    ? TimeUnit.SECONDS.toMillis(1)
+                    : TimeUnit.MINUTES.toMillis(1);
```
dwnusbaum: Not sure about timing here. One minute seems like a decent balance between responsiveness and avoiding unnecessary work, but I am happy to tweak it based on review feedback. Ideally, we'd have a listener so this could be event-driven instead.

bitwiseman:
> I'm not sure if there is a better way to do this, or if the current behavior is even the behavior we want. Why not just allow the Pipelines to continue executing so that suspension/resumption works as normal when Jenkins is restarted?

Note I had a Jenkins user specifically ask this same question.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Not sure if we need to bother with a core patch given that this is kind of a corner case and polling should work well enough.

dwnusbaum:
> > I'm not sure if there is a better way to do this, or if the current behavior is even the behavior we want. Why not just allow the Pipelines to continue executing so that suspension/resumption works as normal when Jenkins is restarted?
>
> Note I had a Jenkins user specifically ask this same question.

I just updated the description, but it looks like the reasoning for the current behavior is discussed in [JENKINS-32015](https://issues.jenkins-ci.org/browse/JENKINS-32015). I am going to try to log a message to the build log when execution is suspended because of quiet mode so users understand why their build appears to be hanging.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -241,6 +241,23 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
...
+    @Issue("JENKINS-34256")
+    @Test public void quietDownThenCancelQuietDown() {
+        story.then(r -> {
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("semaphore 'wait'; echo 'I am done'", true));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("wait/1", b);
+            r.jenkins.doQuietDown(true, 0);
+            SemaphoreStep.success("wait/1", null);
+            ((CpsFlowExecution) b.getExecution()).waitForSuspension();
```
jglick: I would avoid using `waitForSuspension`. Really it should be deprecated—too internal. Would suffice to just sleep for a second here, I think.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -2080,4 +2082,44 @@ protected void notifyShutdown() {
         // No-op, handled in the suspendAll terminator
     }
 
+    // TODO: Core could use a QuietDownListener class with `onQuietDown` and `onCancelQuietDown` methods.
+    @Restricted(NoExternalUse.class)
+    @Extension
+    public static class QuietDownListener extends PeriodicWork {
```
jglick: You should not need a global extension to do this. Rather, when you `saveProgramIfPossible(true)` inside `scheduleRun()`, use `Timer` to set up a task which checks (say) every second to see if _this_ program can resume. (If you use an anonymous inner `Runnable` class, you can `Timer.get().submit(this)` to reschedule if Jenkins is still quieting down.)

dwnusbaum: Ok, I'll look into it. IIUC, you mean that if a build is executing something like a long-running `sh` step when quiet mode is enabled, there is no need to call `scheduleRun` on it if quiet mode is canceled _unless_ the `sh` step completed while quiet mode was enabled, is that right?

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
dwnusbaum: In https://github.com/jenkinsci/workflow-cps-plugin/pull/68 you switched from using `Thread.sleep(1000)` to `CpsFlowExecution.waitForSuspension` in `CpsFlowExecutionTest.quietDown` to try to fix some flaky tests, so I'm a bit hesitant to do the reverse. I'm happy to add a disclaimer to `CpsFlowExecution.waitForSuspension` that the method should not be used unless what you are testing involves some implementation detail of `CpsThreadGroup.scheduleRun`, if you think that would be helpful.

jglick: #68 was a different case: the test would fail if the sleep were too short. Here, at worst, the test could _pass_ if the code were buggy.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: No, that is not what I meant at all. Let me try to rephrase. You are fixing what I consider to be a corner case that many people will never hit. Yet the implementation registers a global extension that does something in the JVM every minute, even if no Pipeline builds are running, and needs to iterate all running Pipeline builds. Instead you can just create a timer task in a _particular_ build that you _know_ was suspended due to quiet mode because we just ran https://github.com/jenkinsci/workflow-cps-plugin/blob/c4ecc1d33d66449531332072db8387347e35c5ba/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java#L274-L278 which will incur that overhead if only and if a build actually needs it, and there is no need to look up `FlowExecutionList` etc. because we have references to all the relevant objects right here.

dwnusbaum: My confusion is that it doesn't seem like there is a meaningful difference between a global extension that only loops over all Pipeline builds when quiet mode is canceled and running a timer task per suspended Pipeline build while quiet mode is enabled, unless there is some reason to expect that the set of suspended Pipeline builds will normally be smaller than the set of all Pipeline builds, so I was trying to figure out if that was the case.

jglick: > there is some reason to expect that the set of suspended Pipeline builds will normally be smaller than the set of all Pipeline builds

Consider the normal case that Jenkins has not been set to quiet down in the current session at all. The current patch would be waking up every minute. I was merely proposing to activate the code in this patch only if and when a particular Pipeline build was actually suspended due to Jenkins quieting down.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -2080,4 +2082,44 @@ protected void notifyShutdown() {
...
+            return Main.isUnitTest
+                    ? TimeUnit.SECONDS.toMillis(1)
+                    : TimeUnit.MINUTES.toMillis(1);
+        }
+
+        @Override
+        protected void doRun() throws Exception {
+            boolean isQuietDown = Jenkins.get().isQuietingDown();
+            // Quieting down is handled by `CpsThreadGroup.scheduleRun`, so we just need to handle quiet mode being cancelled.
+            if (!isQuietDown && wasQuietDown) {
```
jglick: `AtomicBoolean` can express this kind of phase transition a bit better, though in this case only one thread should ever be in this listener at once due to the way `PeriodicWork` works so it does not really matter.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: …and scheduling a task which was actually specific to that build, independent of all others, which ought to make it more obvious from reading the code that it is doing the right thing.

dwnusbaum: > I was merely proposing to activate the code in this patch only if and when a particular Pipeline build was actually suspended due to Jenkins quieting down.

Sure, but then instead of waking up Jenkins once a minute (and usually doing more or less nothing), you are waking up Jenkins even more frequently than once a minute for every Pipeline build when quiet mode is enabled (and usually doing more or less nothing). Depending on how often you enable quiet mode and how many builds are running concurrently, I could see either approach being preferable, so I just wanted to check if I was missing something important that might tip the balance.

Regardless, I switched to your approach in e026365, so let me know what you think.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -241,6 +242,112 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
...
+            assertTrue(b.isBuilding());
+            r.assertLogNotContains("I am done", b);
+            r.jenkins.doCancelQuietDown();
+            r.waitForMessage("Resuming (Prepare for shutdown was canceled)", b);
+            r.assertLogContains("I am done", r.assertBuildStatusSuccess(r.waitForCompletion(b)));
+        });
+    }
+
+    @Issue("JENKINS-34256")
+    @Test public void pauseThenQuietDownThenUnpauseThenCancelQuietDown() {
```
dwnusbaum: I'm not sure about all of these tests; I think they might be flaky. I added them because I was trying to make sure there weren't any bugs around combinations of pause and quiet mode. Some of the cases are a bit weird, especially when you unpause or cancel quiet mode when quiet mode is enabled or when the build is paused, respectively. I think I could tweak `CpsFlowExecution.pause()` to avoid printing consecutive "Pausing" and "Resuming" lines, but I'm not sure if that's worth the complexity.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -271,7 +282,28 @@ public void unexport(BodyReference ref) {
                 @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="runner.submit() result")
                 public Void call() throws Exception {
                     Jenkins j = Jenkins.getInstanceOrNull();
+                    if (j != null && !j.isQuietingDown() && execution != null && pausedByQuietMode.compareAndSet(true, false)) {
+                        try {
+                            execution.getOwner().getListener().getLogger().println("Resuming (Prepare for shutdown was canceled)");
```
dwnusbaum: I added some log messages around quiet mode being enabled/disabled so users understand why there build appears to be hanging. I need to check what the user-facing text says for the "Prepare for shutdown" link in the management page to make sure this is consistent.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -271,7 +282,28 @@ public void unexport(BodyReference ref) {
...
                     if (paused.get() || j == null || (execution != null && j.isQuietingDown())) {
+                        if (j != null && j.isQuietingDown() && execution != null && pausedByQuietMode.compareAndSet(false, true)) {
+                            try {
+                                execution.getOwner().getListener().getLogger().println("Pausing (Preparing for shutdown)");
+                            } catch (IOException e) {
+                               LOGGER.log(Level.WARNING, null, e);
+                            }
+                            Timer.get().schedule(() -> {
+                                if (j.isQuietingDown()) {
+                                    Timer.get().schedule(this, Main.isUnitTest ? 1 : 10, TimeUnit.SECONDS);
```
jglick: `this` is the `Callable` from line 281, I guess. That is what you wanted? If you meant to refer to the `Runnable` at line 299, you cannot use a lambda.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
dwnusbaum: Ok, e026365 switches to using `waitForMessage` instead of a `waitForSuspension`. I think that is ok because the message is printed from inside of `scheduleRun`, but it might be flaky in some cases. I need to think about it some more.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
dwnusbaum: Ha, I was wondering about that, I tried `Runnable.this` to disambiguate earlier just in case, but that was unsupported, and then forgot about it. I guess rerunning the `Callable` technically works, but yeah I'll just use an anonymous inner class instead.

dwnusbaum: See d1322e2.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
jglick: > I'm not sure if that's worth the complexity

Likely not.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
dwnusbaum: Ok, the link to enable quiet mode says "Prepare for shutdown", and the link to cancel quiet mode says "Cancel shutdown", so I kept "Pausing (Preparing for shutdown)", but changed "Resuming (Prepare for shutdown was canceled)" to "Resuming (Shutdown was canceled)" in 3cbff11.

dwnusbaum:
> > Why not just allow the Pipelines to continue executing so that suspension/resumption works as normal when Jenkins is restarted?
>
> It might work, depending on durability modes of metadata and the exact timing of events, but you would be pushing your luck.
>
> If you are just inside some sh step that is running along for an hour or so, Jenkins can restart. The idea is to not be writing critical metadata from the master JVM at the moment it exits.

@jglick Independent of durability mode, as long as you restart Jenkins via something "safe" like `/restart` or `/safeRestart` rather than `kill -9`, [this terminator](https://github.com/jenkinsci/workflow-cps-plugin/blob/c4ecc1d33d66449531332072db8387347e35c5ba/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L1572) should save all metadata for running Pipelines before the restart, right? Or is that terminator not robust and known to not always work as desired?

jglick:
> should save all metadata for running Pipelines before the restart, right?

Probably. But also consider `SynchronousNonBlockingStepExecution`s, which are not currently resumable. You would not want, say, a `junit` step to begin running a moment before the system shuts down.

dwnusbaum:
> You would not want, say, a junit step to begin running a moment before the system shuts down.

Good point. Based on the comments in JENKINS-34256, it seems like some users would like a variant of "Prepare for shutdown" that also waits for Pipelines to complete (ignoring the fact that they can be suspended and then resumed after the restart 🤷‍♂) and others would like a variant that only suspends Pipelines once all non-Pipelines have completed. Both groups of users seem a bit confused that "Prepare for shutdown" does not actually shut down or restart Jenkins.

medianick:
As a longtime Jenkins user, my two cents: I am not in the camp confused by "Prepare for Shutdown"; I've always understood it and used it as a way to block any new builds from starting and allow any running builds to complete -- quiescing the system, essentially. My expectation for Pipeline builds in this scenario would be either that they too simply be allowed to complete (as Freestyle builds do) or -- probably better -- that they reach a safe point at which they could _pause_ (being resumable whether the shutdown were canceled or the service actually got restarted), then pause, then resume as predictably and reliably as possible.

jglick:
> You would not want, say, a `junit` step to begin running a moment before the system shuts down.

Perhaps not as much of a threat as I thought—I had forgotten about the `SynchronousNonBlockingStepExecution.blocksRestart` override. Still, there could be race conditions here. The idea behind suspending programs as soon as the first Groovy safepoint is encountered after quiet mode is entered is that otherwise Jenkins might do one last restart-blocked check, get a go-ahead because a `sh` step is running, and then initiate shutdown while a CPS VM thread is switching to new tasks.

> it seems like some users would like a variant of "Prepare for shutdown" that also waits for Pipelines to complete

Reasonable enough. Would involve waiting for all asynchronous executors to exit.

> only suspends Pipelines once all non-Pipelines have completed

I believe this could be accomplished by amending the current quiet down check in this plugin to check for heavyweight executors also being idle. Again I am not sure this is safe.

> "Prepare for shutdown" does not actually shut down or restart Jenkins

That would be `/safeExit` or `/safeRestart`, respectively. Not sure if we currently offer this as a GUI option.

> that they reach a safe point at which they could _pause_

This is the current behavior, or the intention behind it at any rate.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
car-roll: Tests seem good when I run them. I see the need for all the combinations, but yeah it does  kind of suck to have  to do all these variations. There's probably a parameterized way to simplify the code, but I don't think it's worth the effort. Maybe put it another test class?

dwnusbaum: Yeah, using a parameterized test would probably make the intention clearer, but I think the abstraction might make the individual tests more difficult to understand, so I'd prefer to leave them as-is.

dwnusbaum:
Ok, I'll merge this, and then I'll comment on the ticket to explain the expected behavior before and after my change, and to mention that if users would prefer different behavior, they should open a new ticket and explain their use case and the desired behavior.

hostalp:
> Why not just allow the Pipelines to continue executing so that suspension/resumption works as normal when Jenkins is restarted?
Could this be achieved somehow?

I see occassional problem with interaction with the ThinBackup plugin with `Wait until Jenkins/Hudson is idle to perform a backup` and `Force Jenkins to quiet mode after specified minutes` options set.

If a pipeline job is running and gets paused by the ThinBackup activity, the job will at some point log `Pausing (Preparing for shutdown)`, its execution will stop (at least I can see nothing running at that point - no slave processes, no master threads related to the job), but in Jenkins it will be still seen as active on the slave, just with the `(Analyze)` note below the running job number.

The whole Jenkins will remain stuck in that state since ThinBackup is waiting for all jobs to complete and thus keeping the Jenkins in the quiet state, but the blocking job will never complete because it's waiting for Jenkins to resume. The only wsy out of this is to abort the blocking job manually.

dwnusbaum:
@hostalp Please add general comments like the to the Jira ticket instead of the PR. Looking through the comments on [JENKINS-34256](https://issues.jenkins-ci.org/browse/JENKINS-34256), your issues sounds like it might be addressed by [JENKINS-60434](https://issues.jenkins-ci.org/browse/JENKINS-60434). If you agree, I would just add any additional information that seems relevant in a comment on that issue and then vote for it and add start watching it. If you think a different approach is needed for your use case, please open a new Jira ticket with details and link it to JENKINS-34256.

hostalp:
No problem. I didn't find the [JENKINS-60434](https://issues.jenkins-ci.org/browse/JENKINS-60434) (yes. it's the right one) initially and didn't want to open a new issue for something that was meant as a simple question whose answer would decide further steps, if any. :smiley:

PR #344, opened 2019-12-10
Add 2.78 release notes

dwnusbaum:
I would like to release the plugin today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Release date: 2019-12-10
 
-* Fix: Resume Pipeline execution when Jenkins shutdown is canceled if the execution was automatically paused because Jenkins was preparing for shutdown. ([JENKINS-34256](https://issues.jenkins-ci.org/browse/JENKINS-34256))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Changelog
 
+### 2.78
+
+Release date: 2019-12-10
+
+* Fix: Resume Pipeline execution when Jenkins shutdown is canceled if the execution was automatically paused because Jenkins was preparing for shutdown. ([JENKINS-34256](https://issues.jenkins-ci.org/browse/JENKINS-34256))
```
kshultzCB: Maybe something like:
Fix: If a running Pipeline is paused because Jenkins is preparing for shutdown, execution will now resume if that shutdown is canceled.

dwnusbaum: Yeah, I'm not crazy about my wording. I'll think about it a bit.

dwnusbaum: Ok, I tried to clarify the wording in 7e85160 by splitting it into separate sentences, one simple sentence for the fix, and then a longer one explaining the difference from the previous behavior.

PR #345, opened 2019-12-17
Bump parent pom to 3.54

slonopotamus:


slonopotamus:
All failed tests start a build but neither wait for completion nor abort it.

I'd actually say this is a bug in JMH because it pretends as if it shut down Jenkins instance but in fact there are builds still running.

Anyway, it is better to fix tests on workflow-cps-plugin side because JMH might declare having running builds during shutdown an error.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpActionTest.java
```diff
@@ -40,11 +42,13 @@
     @Test public void doProgramDotXml() throws Exception {
         WorkflowJob p = r.createProject(WorkflowJob.class);
         p.setDefinition(new CpsFlowDefinition("node {semaphore 'wait'}", true));
-        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        final QueueTaskFuture<WorkflowRun> f = p.scheduleBuild2(0);
```
dwnusbaum: It would be nice if `JenkinsRule` had a `waitAndAssertBuildStatus(Run)` method, or if `JenkinsRule.waitForCompletion` returned the `JenkinsRule`, since right now you either have to do what you did and create a variable for the `QueueTaskFuture`, or call `JenkinsRule.waitForCompletion` on its own line, both of which feel a little messy.

slonopotamus: I totally agree with you.

PR #347, opened 2020-01-20
Update groovy-cps to ignore some CPS method mismatch warnings

dwnusbaum:
See https://github.com/cloudbees/groovy-cps/pull/103. Just testing that PR and picking it up in a plugin so other plugins can test the fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
         <git-plugin.version>3.1.0</git-plugin.version>
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <groovy-cps.version>1.31</groovy-cps.version>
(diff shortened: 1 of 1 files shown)


car-roll:
Looks like there are still mismatch warnings, did it used to throw more?

dwnusbaum:
I think something else is happening, because the tests are failing with a `CpsCallableInvocation`:

```
hudson.remoting.ProxyException: CpsCallableInvocation{methodName=sleep, call=com.cloudbees.groovy.cps.impl.CpsFunction@17b65412, receiver=null, arguments=[org.jenkinsci.plugins.workflow.cps.DSL$ThreadTaskImpl@3f07eba7]}
```

I thought I merged with master upstream before deploying the snapshot, but maybe not? I'm not quite sure what is happening here, will investigate. Maybe one of the changes to proxy exceptions to avoid memory leaks is breaking the mismatch detection somehow?

dwnusbaum:
I am closing this PR, see https://github.com/cloudbees/groovy-cps/pull/103#issuecomment-576797618. That fix suppressed things that we did not want to suppress here, which was causing test failures. We could have fixed it, but it doesn't make sense to do so right now.

PR #348, opened 2020-01-23
Remove assertion of shell context variables on pipeline deserialization

car-roll:
There exists an assertion that shell context variables must be empty when the pipeline is being deserialized.  However, shell context variables can be non-empty in some cases when used with extensions that hook into the Pipeline’s execution and does not appear to be an issue. Removing the assertion as it causes false failures in unit tests.


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -174,16 +174,13 @@ private Object readResolve() {
         execution = CpsFlowExecution.PROGRAM_STATE_SERIALIZATION.get();
         setupTransients();
         assert execution!=null;
-        if (scripts != null) { // compatibility: the field will be null in old programs
+        if (scripts != null && !scripts.isEmpty()) { // compatibility: the field will be null in old programs
```
jglick:
```suggestion
        if (/* compatibility: the field will be null in old programs */ scripts != null && !scripts.isEmpty()) {
```

to clarify which part of the clause the comment applies to

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -174,16 +174,13 @@ private Object readResolve() {
...
-            assert shell.getContext().getVariables().isEmpty();
-            if (!scripts.isEmpty()) {
-                // Take the canonical bindings from the main script and relink that object with that of the shell and all other loaded scripts which kept the same bindings.
-                shell.getContext().getVariables().putAll(scripts.get(0).getBinding().getVariables());
-                for (Script script : scripts) {
-                    if (script.getBinding().getVariables().equals(shell.getContext().getVariables())) {
-                        script.setBinding(shell.getContext());
-                    }
+            // Take the canonical bindings from the main script and relink that object with that of the shell and all other loaded scripts which kept the same bindings.
+            shell.getContext().getVariables().putAll(scripts.get(0).getBinding().getVariables());
```
jglick: So this will overwrite any variable bindings present in the original shell. Is that what you meant to do?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -174,16 +174,13 @@ private Object readResolve() {
...
+            for (Script script : scripts) {
```
jglick: And is the equality check on the next line still going to work under the circumstances you describe? The new test does not cover this case (e.g., use of `load` step).

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -477,6 +477,33 @@ private void trustedShell(final boolean pos) {
...
+        story.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            SemaphoreStep.success("wait/1", null);
+            r.waitForCompletion(b);
+            r.assertBuildStatus(Result.SUCCESS, b);
+        });
+    }
+
+    @TestExtension("ExistingVariablesOnRestart")
```
jglick:
```suggestion
    @TestExtension("existingVariablesOnRestart")
```

The fact that the test passes despite this mistake is a bad sign. Did you verify the bug symptom with `src/main/` changes reverted? The test is not asserting anything about this particular variable, only that the build does not die.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -477,6 +477,33 @@ private void trustedShell(final boolean pos) {
         });
     }
 
+    @Test
+    public void existingVariablesOnRestart() throws Exception {
+        story.then(r -> {
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("echo('pre-semaphore')\n" +
```
dwnusbaum: I would update the `echo` steps to have them use the injected variable, and then check the logs and assert that it was present both before and after the restart.

```suggestion
            p.setDefinition(new CpsFlowDefinition("echo('Pre-semaphore value is ${existing}')\n" +
```

And then you can add an assertion like this before the restart, and a similar one afterwards: 
```
r.assertLogContains("Pre-semaphore value is value", b);
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
dwnusbaum: Good point, we could add a test like `RestartingLoadStepTest.accessToSiblingScripts` that also has variables injected via a `GroovyShellDecorator` to test the combined behavior.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
jglick: > manually checked that the injected variables are present before and after the restart by dumping the program state as XML

Interesting, but should be easy enough to just `echo` the variable value in an automated test as well.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
jglick: > it would probably be better only overwrite variables that are not present in the context already

Not sure. I _think_ the unconditional overwrite as currently implemented is fine; just wanted to bring up the question. Ideally the test would pin a specific behavior.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
car-roll: addressed in f15b02ddf3268eac1df80d25f14745975d68baa4

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
car-roll: addressed in f15b02ddf3268eac1df80d25f14745975d68baa4

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
car-roll: addressed in f15b02ddf3268eac1df80d25f14745975d68baa4

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
car-roll: addressed in f15b02ddf3268eac1df80d25f14745975d68baa4

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
car-roll: Added a test `clashingVariablesOnRestart` that shows the unconditional overwrite behavior

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final String A_CONTENTS = "def call(arg) {echo \"a ran on ${arg}\"}; this";
+    private static final String B_GROOVY = "b.groovy";
+    private static final String B_CONTENTS = "def m(arg) {" +
+                                             "echo \"${this} binding=${binding.variables}\";" +
+                                             "a(\"${arg} from b\")}; this";
+    private static String LOAD_PREFIX = "a = 0;" +
+                                        "node {a = load 'a.groovy'};" +
+                                        "echo \"Pre-semaphore value is ${" + EXISTING_VAR_NAME + "}\";";
+    private static String LOAD_SEMAPHORE = "semaphore 'wait';";
+    private static String LOAD_SUFFIX = "echo \"Post-semaphore value is ${" + EXISTING_VAR_NAME + "}\";" +
```
dwnusbaum: nit: I find it really hard to read tests where the Pipeline scripts is stored in fragments in fields like this, since you have to jump around the file to see what is happening. I prefer inlining the scripts into the tests so you can see everything at once even if it means there is a bit of duplication.

Also, one thing that can be useful to avoid some escaping is that Groovy supports `/` as a `GString` delimiter with the same behavior as `"`, so you could write this instead:

```suggestion
    private static String LOAD_SUFFIX = "echo(/Post-semaphore value is ${" + EXISTING_VAR_NAME + "}/)" +
```

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
```diff
@@ -0,0 +1,136 @@
...
+            r.waitForCompletion(b);
+            r.assertBuildStatus(Result.SUCCESS, b);
+            r.assertLogContains("Post-semaphore value is " + EXISTING_VAR_VALUE, b);
+        });
+    }
+
+    @Test
+    public void loadWithExistingVariables() {
+        // Testing with variables also injected via load step
+        story.addStep(new Statement() {
```
dwnusbaum:
```suggestion
        story.then(r -> {
```

And then replace uses of `story.j` with `r`. Same for the other tests.

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runners.model.Statement;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+import org.jvnet.hudson.test.TestExtension;
+import javax.annotation.CheckForNull;
+
+public class DeserializationTest {
```
dwnusbaum: Since the new tests are related to restarts and the load step, I'd move them to `RestartingLoadStepTest` instead of making a new file (even though `existingVariablesOnRestart` does not use `load`, it's exercising the logic added for `load` steps, so I think it would be fine there). You can pass multiple method names to `@TestExtension`, like `@TestExtension(value={"methodA", "methodB"})` so that `InjectedVariable` only activates for the new tests.

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
```diff
@@ -0,0 +1,136 @@
...
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                story.j.jenkins.getWorkspaceFor(p).child(A_GROOVY).write(A_CONTENTS, null);
+                ScriptApproval.get().approveSignature("method groovy.lang.Binding getVariables");
+                story.j.jenkins.getWorkspaceFor(p).child(B_GROOVY).write(B_CONTENTS, null);
+                // Control case:
+                p.setDefinition(new CpsFlowDefinition(sameVariable + LOAD_PREFIX + LOAD_SUFFIX, true));
+                WorkflowRun b = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+                story.j.assertLogContains("a ran on value from b", b);
+                story.j.assertLogContains("Pre-semaphore value is " + clashingValue, b);
```
dwnusbaum: Just for consistency:

```suggestion
                story.j.assertLogContains("Pre-semaphore value is " + clashingValue, b);
                story.j.assertLogContains("Post-semaphore value is " + clashingValue, b);
```

car-roll:
kicking off build again to see if i can get artifact uploaded

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
```diff
@@ -0,0 +1,136 @@
...
+            }
+        });
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p", WorkflowJob.class);
+                WorkflowRun b = p.getBuildByNumber(1);
+                SemaphoreStep.success("wait/1", null);
+                story.j.assertLogContains("a ran on value from b", story.j.assertBuildStatusSuccess(story.j.waitForCompletion(b)));
+                story.j.assertLogContains("Post-semaphore value is " + clashingValue, b);
```
dwnusbaum: It's not obvious to me what is happening in this case. Did you check via a debugger that the value of `INJECTED_VAR` in `shell.getContext().getVariables()` was `PRE_EXISTING`  in `readResolve` before being replaced? I would've thought that it was already `0` before the restart.

Also, in regards to https://github.com/jenkinsci/workflow-cps-plugin/pull/348/files#r370153155, I think the loaded scripts would also need to echo `INJECTED_VAR` after the main script changes the value of `INJECTED_VAR` to show that the equality check passes and that the bindings were linked successfully.

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
car-roll: Nice tip! I thought that was a typo.
Fair point, I'll go ahead and inline the scripts. I mostly did it because I had the opposite problem where I was trying to see what the difference was between the control and the test script.

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
car-roll: so i did use the debugger to check that the `INJECTED_VAR` was `PRE_EXISTING` before  the [`putAll`](https://github.com/car-roll/workflow-cps-plugin/blob/7af475fda8512d321dec623074f07b83211a03ba/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java#L180). 
I was testing the scenario where the load script changes an environment variables, and then that new value survives  a restart.  
So in this case,`INJECTED_VAR` is 0 before the restart, but then `GroovyShellDecorator` will set `INJECTED_VAR` to `PRE_EXISTING` at the restart again and then readresolve is supposed to reset `INJECTED_VAR` back to zero.

dwnusbaum: > so i did use the debugger to check that the INJECTED_VAR was PRE_EXISTING before the putAll.

Ok, sounds good, thanks for clarifying.

As far as https://github.com/jenkinsci/workflow-cps-plugin/pull/348/files#r370153155, do any of the new tests fail if [these lines](https://github.com/jenkinsci/workflow-cps-plugin/blob/5db478ab5d6781696d9b61ce75fae57774871040/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java#L184-L185) are removed? If so, then I think we're good and injecting vars via `GroovyShellDecorator` works ok with the `load` step.

car-roll: Unfortunately, this test does not fail when that equality line is removed. I'm trying to see if i can get another test to properly handle that case.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
car-roll: 2afbe31 removed the equality check and now applies all script bindings to the shell context.  @dwnusbaum created a test case called `updatedBindingsOnRestart` where the loaded script carries a stale definition if it is not given the shell context.

dwnusbaum: To expand on this, we couldn't come up with a case where you didn't want the bindings to always be mapped to the same instance of `Binding`. Before the restart, at least in the tests cases we came up with, the main script and loaded scripts always shared the same instance of `Binding`. After the restart, the loaded scripts would have the same bindings as or a subset of the bindings of the main script (but never a superset), because some bindings that deserialize correctly as part of the main script get silently dropped in the binding for the the loaded script (I am not sure why, the ones that were dropped from the loaded script had values with type `CpsClosure` and other Groovy and CPS-specific types. Simple values seem to be ok.)

When the equality check was in place, if the main script did not have any bindings that the loaded scripts silently failed to deserialize, the check would pass, and the bindings would be linked. If the main script had a binding that could not be deserialized, as in the case with `accessToSiblingVariables`, the condition would pass for the main script but not the sibling scripts, which caused a bug after the restart as seen in `updatedBindingsOnRestart`. Since we couldn't come up with a case where it made sense for the equality check to fail (please let us know if you have any ideas), it seemed best to just remove it to fix `updatedBindingsOnRestart`.

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
```diff
@@ -218,4 +226,134 @@ public void pauseInsideLoad() throws Exception {
...
+            SemaphoreStep.waitForStart("wait/1", b);
+            r.assertLogContains("before change: tmp", b);
+            r.assertLogContains("before restart: tmp2", b);
+        });
+        story.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            SemaphoreStep.success("wait/1", null);
+            r.assertBuildStatusSuccess(r.waitForCompletion(b));
+            r.assertLogContains("after restart: tmp3", b);
```
dwnusbaum: With the equality check in `CpsThreadGroup.readResolve` in place, the value here would still be `tmp2`, since the equality check would fail and the bindings would not be linked. Since the bindings are linked before the restart, I think it makes the most sense for them to be linked after the restart too.

src/test/java/org/jenkinsci/plugins/workflow/cps/DeserializationTest.java
dwnusbaum: Did you mean to delete this file? It looks like you moved the tests over to `RestartingLoadStepTest`.

dwnusbaum:
Looks like there were infra issues, trying another build:

> [ERROR] Failed to execute goal on project workflow-cps: Could not resolve dependencies for project org.jenkins-ci.plugins.workflow:workflow-cps:hpi:2.79-rc2474.eaa73531f081: Failed to collect dependencies at org.jenkins-ci.main:jenkins-core:jar:2.164.1 -> commons-httpclient:commons-httpclient:jar:3.1-jenkins-1: Failed to read artifact descriptor for commons-httpclient:commons-httpclient:jar:3.1-jenkins-1: Could not transfer artifact commons-httpclient:commons-httpclient:pom:3.1-jenkins-1 from/to repo.jenkins-ci.org (https://repo.jenkins-ci.org/public/): Remote host closed connection during handshake: SSL peer shut down incorrectly

src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
```diff
@@ -218,4 +226,134 @@ public void pauseInsideLoad() throws Exception {
...
+            SemaphoreStep.success("wait/1", null);
+            r.waitForCompletion(b);
+            r.assertBuildStatus(Result.SUCCESS, b);
+            r.assertLogContains("Post-semaphore value is " + EXISTING_VAR_VALUE, b);
+            r.assertLogContains("a ran on value from b", b);
+        });
+    }
+
+    @Test
+    public void conflictingBindingsOnRestart() throws Exception {
```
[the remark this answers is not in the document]
car-roll: yes, that's basically it. I was just writing out a test that explicitly shows what the behavior was with conflicting variables. But I think it doesn't really take much wrangling to figure out that behavior without this test.  I'll go ahead and remove it. Less code is better, since it's not testing any additional edge cases, like you said.

PR #350, opened 2020-02-07
Introduced GroovySample extension point

jglick:
Plan to introduce something in the `kubernetes` plugin. Also at some point the Declarative-specific samples should be moved out.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
@@ -1,0 +1,32 @@
+pipeline {
+   agent any
+
+   tools {
+      // Install the Maven version configured as "M3" and add it to the path.
+      maven "M3"
+   }
+
+   stages {
+      stage('Build') {
+         steps {
+            // Get some code from a GitHub repository
+            git 'https://github.com/jglick/simple-maven-project-with-tests.git'
+
+            // Run Maven on a Unix agent.
+            sh "mvn -Dmaven.test.failure.ignore=true clean package"
+
+            // To run Maven on a Windows agent, use
+            // bat "mvn -Dmaven.test.failure.ignore=true clean package"
+         }
+
+         post {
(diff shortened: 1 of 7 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
```diff
@@ -0,0 +1,32 @@
...
+   tools {
+      // Install the Maven version configured as "M3" and add it to the path.
+      maven "M3"
+   }
+
+   stages {
+      stage('Build') {
+         steps {
+            // Get some code from a GitHub repository
+            git 'https://github.com/jglick/simple-maven-project-with-tests.git'
```
olamy: what about something from here https://github.com/cloudbeers?

src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/scripted.groovy
```diff
@@ -0,0 +1,25 @@
+node {
+   def mvnHome
+   stage('Preparation') { // for display purposes
+      // Get some code from a GitHub repository
+      git 'https://github.com/jglick/simple-maven-project-with-tests.git'
```
olamy: what about something from here https://github.com/cloudbeers?

src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
jglick: For this PR my goal was simply to introduce the extension point. Content changes can be done as follow-ups (and more easily reviewed as real `*.groovy` files).

src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/scripted.groovy
jglick: same

jglick:
> Do you care about being able to suppress or filter default samples

Already done downstream in https://github.com/jenkinsci/kubernetes-plugin/pull/707.

PR #353, opened 2020-03-17
Use FailFastCause to provide a nicer cause of failure

jglick:
Previously `ParallelStep.ResultHandler.Callback.checkAllDone` made a `FailFastException` with no (`null`) message, then calls `ParallelStepExecution.stop`, which calls `BodyExecution.cancel(Throwable)`, which creates a `FlowInterruptedException` with an `ExceptionCause`, so the message is poor.

Amends https://github.com/jenkinsci/pipeline-plugin/pull/88.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
@@ -14,13 +14,18 @@ import static java.util.Arrays.*;
 import java.util.Comparator;
 import java.util.List;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
+import jenkins.model.CauseOfInterruption;
 
 import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.workflow.SingleJobTestBase;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStepTest.java
```diff
@@ -144,7 +149,14 @@ public void failure_in_subflow_will_fail_fast() throws Exception {
                 startBuilding().get();
                 assertBuildCompletedSuccessfully();
                 Assert.assertFalse("a should have aborted", jenkins().getWorkspaceFor(p).child("a.done").exists());
-
+                for (FlowNode n : new DepthFirstScanner().allNodes(e)) {
+                    ErrorAction err = n.getPersistentAction(ErrorAction.class);
+                    if (err != null) {
+                        if (err.getError() instanceof FlowInterruptedException) {
+                            assertEquals("Failed in branch b", ((FlowInterruptedException) err.getError()).getCauses().stream().map(CauseOfInterruption::getShortDescription).collect(Collectors.joining("; ")));
```
jglick: before patch, result was `Exception: null`

PR #223, opened 2018-04-20
Fix more subtle durability issues 

svanoort:
Addresses the following root causes of problems in Pipeline persistence/durability:

* Pipelines which modify the heads in an execution *without* ensuring the corresponding FlowNodes are also saved to disk - this will cause issues when trying to load the execution -- applies only to the fully durable modes (since non-durable just won't persist period)
* Pipelines which modify the `CPSThread.head` `FlowHead` and its `FlowNode` without ensuring the CpsFlowExecution is saved (via `saveOwner()` that saves the `WorkflowRun`).  This causes an NPE loading the program, because the FlowHead may map to a head in the CpsFlowExecution that will not exist at time of deserialization. 
* Provides some documentation on the persistence model so others do not endure my pain
* Removes explicit persistence calls after `FlowHead#setNewHead(FlowNode)` calls because `setNewHead` now forces flownodes to be saved *before* the head is mutated, guaranteeing an always-usable state.
* Expand the fuzzTesting cases
* Consolidate all shutdown-time persistence into the main `CpsFlowExecution#suspendAll()` terminator - *may* help avoid the occasional ConcurrentModificationException we were seeing at serialize, and is more "correct" and safer generally

TARGETTED AT:

* [JENKINS-46986](https://issues.jenkins-ci.org/browse/JENKINS-46986) specifically, and I no longer see NPEs thanks to this

TODO:

- [x] Fix this error seen at shutdown-timer persistence in live testing: 

> Apr 18, 2018 1:01:08 AM hudson.model.Executor finish1
SEVERE: Executor threw an exception
java.lang.IllegalStateException: Jenkins has not been started, or was already shut down
    at jenkins.model.Jenkins.getActiveInstance(Jenkins.java:762)
    at com.cloudbees.hudson.plugins.folder.AbstractFolder.getDescriptor(AbstractFolder.java:605)
    at jenkins.branch.MultiBranchProject.getDescriptor(MultiBranchProject.java:863)
    at jenkins.branch.MultiBranchProject.getDescriptor(MultiBranchProject.java:124)
    at com.cloudbees.hudson.plugins.folder.AbstractFolder.childNameGenerator(AbstractFolder.java:597)
    at com.cloudbees.hudson.plugins.folder.AbstractFolder.getRootDirFor(AbstractFolder.java:640)
    at jenkins.branch.MultiBranchProject.getRootDirFor(MultiBranchProject.java:843)
[...]

PR #354, opened 2020-03-30
[JENKINS-55287] Improve shutdown-related logging

dwnusbaum:
See [JENKINS-55287](https://issues.jenkins-ci.org/browse/JENKINS-55287).

While looking through some potentially-relevant code for that issue, I noticed that the logging when saving `CpsFlowExecutions` using the `PERFORMANCE_OPTIMIZED` durability level was a bit inconsistent, didn't really have a single clear message of whether everything was successful or not, didn't log anything for a few conditions that should never happen, used a default impl of `toString` (e.g. `MyClass@hash`) in some cases that were intending to print the name of the flow execution, and mentioned shutdown in some cases where Jenkins isn't shutting down.

~Opening as a draft PR for now, because I want to look through the logs of the tests before and after the change to see if anything interesting stands out, and maybe add some assertions on specific log messages to existing tests. I did check the output of some tests in `PersistenceProblemsTest` just as an overall sanity check.~

More detailed analysis of JENKINS-55287 is in https://github.com/jenkinsci/workflow-cps-plugin/pull/363 (which incorporates this PR). I think this PR is still useful, so I am marking it as ready for review so it can be reviewed and merged independently from https://github.com/jenkinsci/workflow-cps-plugin/pull/363

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1548,7 +1548,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 if (v) {
                     g.pause();
                     checkAndAbortNonresumableBuild();  // TODO Verify if we can rely on just killing paused builds at shutdown via checkAndAbortNonresumableBuild()
-                    checkpoint();
+                    checkpoint(false);
                 } else {
                     g.unpause();
                 }
@@ -1591,7 +1591,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                                 LOGGER.log(Level.WARNING, "Error waiting for Pipeline to suspend: "+exec, ex);
                             }
                         }
-                        cpsExec.checkpoint();
+                        cpsExec.checkpoint(true);
                     }
                 } catch (Exception ex) {
                     LOGGER.log(Level.WARNING, "Error persisting Pipeline execution at shutdown: "+((CpsFlowExecution) execution).owner, ex);
@@ -1972,13 +1972,18 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         }
     }
 
-    /** Save everything we can to disk - program, run, flownodes. */
-    private void checkpoint() {
+    /**
+     * Save everything we can to disk - program, run, flownodes.
+     * @param shuttingDown True if this checkpoint is happening because Jenkins is shutting down, false if it is happening because execution was paused.
+     */
+    private void checkpoint(boolean shuttingDown) {
         if (isComplete() || this.getDurabilityHint().isPersistWithEveryStep()) {
             // Nothing to persist OR we've already persisted it along the way.
             return;
         }
-        LOGGER.log(Level.INFO, "Attempting to save a checkpoint of {0} before shutdown", this);
+        LOGGER.log(Level.INFO, "Attempting to save a checkpoint of all data for {0}{1}", new Object[] {
+            this, shuttingDown ? " before shutdown" : ""
+        });
         boolean persistOk = true;
         FlowNodeStorage storage = getStorage();
         if (storage != null) {
@@ -2049,9 +2054,13 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         }
 
         if (persistOk) {
-            LOGGER.log(Level.INFO, "Successfully saved a checkpoint of {0} before shutdown", this);
+            LOGGER.log(Level.INFO, "Successfully checkpointed {0}{1}", new Object[] {
+                this, (shuttingDown ? " before shutdown" : "")
+            });
         } else {
-            LOGGER.log(Level.WARNING, "Unable to save a checkpoint of {0} before shutdown. The build will probably fail when Jenkins restarts.", this);
+            LOGGER.log(Level.WARNING, "Unable to successfully checkpoint {0}{1}", new Object[] {
+                this, (shuttingDown ? " before shutdown, so this build will probably fail when Jenkins restarts" : "")
+            });
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1978,6 +1978,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             // Nothing to persist OR we've already persisted it along the way.
             return;
         }
+        LOGGER.log(Level.INFO, "Attempting to save a checkpoint of {0} before shutdown", this);
         boolean persistOk = true;
         FlowNodeStorage storage = getStorage();
         if (storage != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
...
+                            // Logged at Level.WARNING when we call `myOutcome.get` and it throws an exception.
                             myOutcome.setException(t);
                         }
                     });
                     myOutcome.get(30, TimeUnit.SECONDS);
-                    LOGGER.log(Level.FINE, "Successfully saved program for "+this);
+                    LOGGER.log(Level.FINE, "Successfully saved program for: {0}", this);
+                } else {
+                    persistOk = false;
+                    LOGGER.log(Level.WARNING, "Unable to persist program because it was never loaded for: {0}", this);
```
dwnusbaum: I don't think this should ever happen, but definitely worth a log message if it does!

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
...
                 saveOwner();
             } catch (Exception ex) {
-                LOGGER.log(Level.WARNING, "Error saving build for "+this, ex);
+                persistOk = false;
+                LOGGER.log(Level.WARNING, "Error saving build for: " + this, ex);
             }
 
+        } else {
+            persistOk = false;
+            LOGGER.log(Level.WARNING, "No FlowNode storage for: {0}", this);
```
dwnusbaum: Ditto, should never happen, but if it does, it should be logged.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
             // Nothing to persist OR we've already persisted it along the way.
             return;
         }
+        LOGGER.log(Level.INFO, "Attempting to save a checkpoint of {0} before shutdown", this);
```
dwnusbaum: This message and the one at the end of `CpsFlowExecution.checkpoint` are now the only messages that are logged at INFO level if everything is successful. Any errors are logged at WARNING level.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
...
                 persistOk=false;
-                LOGGER.log(Level.WARNING, "Error persisting FlowNode storage before shutdown", ioe);
+                LOGGER.log(Level.WARNING, "Error persisting FlowNode storage for: " + this, ioe);
             }
 
             // Try to ensure we've saved the appropriate things -- the program is the last stumbling block.
             try {
                 final SettableFuture<Void> myOutcome = SettableFuture.create();
-                LOGGER.log(Level.INFO, "About to try to checkpoint the program for build"+this);
+                LOGGER.log(Level.FINE, "About to try to checkpoint the program for: {0}", this);
```
dwnusbaum: Now we only care about the overall failure/success, so this was moved to FINE.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
...
                 }
 
             } catch (TimeoutException te) {
                 persistOk = false;
-                LOGGER.log(Level.WARNING, "Timeout persisting program at execution checkpoint", te);
+                LOGGER.log(Level.WARNING, "Timeout persisting program for: " + this, te);
             } catch (ExecutionException | InterruptedException ex) {
                 persistOk = false;
-                LOGGER.log(Level.FINE, "Error saving program, that should be handled elsewhere.", ex);
+                LOGGER.log(Level.WARNING, "Error saving program for: " + this, ex);
```
dwnusbaum: Moved up from FINE to WARNING since the logging for exceptions in `onSuccess` and `onFailure` was moved down to FINE.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
...
                 final SettableFuture<Void> myOutcome = SettableFuture.create();
-                LOGGER.log(Level.INFO, "About to try to checkpoint the program for build"+this);
+                LOGGER.log(Level.FINE, "About to try to checkpoint the program for: {0}", this);
                 if (programPromise != null && programPromise.isDone()) {
                     runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
                         @Override
                         public void onSuccess(CpsThreadGroup result) {
                             try {
-                                LOGGER.log(Level.INFO, "Trying to save program before shutdown "+this);
+                                LOGGER.log(Level.FINE, "Trying to save program for: {0}", CpsFlowExecution.this);
```
dwnusbaum: Switched to `CpsFlowExecution.this` so that the log messages have a string like ` CpsFlowExecution[Owner[JobName/BuildNumber:JobName #BuildNumber]]` instead of `CpsFlowExecution$8@RandomHash`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1978,66 +1978,80 @@ private void checkpoint() {
...
 
+        } else {
+            persistOk = false;
+            LOGGER.log(Level.WARNING, "No FlowNode storage for: {0}", this);
+        }
+
+        if (persistOk) {
+            LOGGER.log(Level.INFO, "Successfully saved a checkpoint of {0} before shutdown", this);
+        } else {
+            LOGGER.log(Level.WARNING, "Unable to save a checkpoint of {0} before shutdown. The build will probably fail when Jenkins restarts.", this);
```
dwnusbaum: Now we have one message to look for per build that will show up if any of the program, flow nodes, or build itself were unable to be saved successfully.

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -245,7 +244,6 @@ public void inProgressNormal() throws Exception {
     }
 
     @Test
-    @Ignore
```
dwnusbaum: These tests have been ignored since they were added in https://github.com/jenkinsci/workflow-cps-plugin/pull/223. They were unignored when the tests were duplicated over in https://github.com/jenkinsci/workflow-job-plugin/pull/104 (see https://github.com/jenkinsci/workflow-job-plugin/pull/104/commits/35b435d87ee77d96857f65c26b5207a299450781), so I think the fact that they are still ignored here was just an accident. Since I was running these tests anyway to test the logging changes, I went ahead and unignored them.

PR #355, opened 2020-05-04
Fix build errors by updating minimum core version and parent POM

dwnusbaum:
Should fix enforcer errors seen in the CI builds when testing against 2.222.x.

2.176.x is the fourth-oldest LTS line at this point, so it seems like a conservative enough upgrade. The core BOM used by version 4.x of the parent POM also supports 2.164.3 if we want to be even more conservative.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
@@ -24,7 +24,7 @@ import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.RestartableJenkinsRule;
 
-import java.io.File;
+import java.nio.file.Files;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -166,8 +166,7 @@ public class PersistenceProblemsTest {
 
             // Hack but deletes the file from disk
             CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
-            File f = new File(cpsExec.getStorageDir(), finalId+".xml");
-            f.delete();
+            Files.delete(cpsExec.getStorageDir().toPath().resolve(finalId+".xml"));
             executionAndBuildResult[0] = ((CpsFlowExecution)(run.getExecution())).getResult();
             executionAndBuildResult[1] = run.getResult();
         });
@@ -313,7 +312,11 @@ public class PersistenceProblemsTest {
         story.thenWithHardShutdown( j -> {
             WorkflowRun run = runBasicPauseOnInput(j, DEFAULT_JOBNAME, build);
             CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
-            cpsExec.getProgramDataFile().delete();
+            // Wait until program.dat is written and then delete it.
+            while (!Files.exists(cpsExec.getProgramDataFile().toPath())) {
+                Thread.sleep(100);
+            }
+            Files.delete(cpsExec.getProgramDataFile().toPath());
         });
         story.then( j->{
             WorkflowJob r = j.jenkins.getItemByFullName(DEFAULT_JOBNAME, WorkflowJob.class);
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,5 @@
-buildPlugin(useAci: false, configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: false, configurations: [
+  [ platform: "linux", jdk: "8", jenkins: null ],
+  [ platform: "windows", jdk: "8", jenkins: null ],
+  [ platform: "linux", jdk: "11", jenkins: "2.222.3", javaLevel: 8 ]
+])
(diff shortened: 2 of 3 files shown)


dwnusbaum:
The test failures on Windows seem infra related (maybe mingit is broken on Windows ACI agents?), they all fail with this error:

> C:\mingit\usr\bin\sh.exe: *** fatal error - unable to load netapi32.dll, Win32 error 126

Not sure if I can fix the issues here, or what the root cause is. I'll try switching to non-ACI agents to see if that makes a difference in case the issue is related to https://github.com/jenkins-infra/pipeline-library/pull/144.

dwnusbaum:
Rebuilding:

> [ERROR] Failed to execute goal on project workflow-cps: Could not resolve dependencies for project org.jenkins-ci.plugins.workflow:workflow-cps:hpi:2.81-rc2484.f64531a6eaa2: Failed to collect dependencies at org.jenkins-ci.plugins.workflow:workflow-durable-task-step:jar:2.18 -> org.jenkins-ci.plugins.workflow:workflow-step-api:jar:2.13: Failed to read artifact descriptor for org.jenkins-ci.plugins.workflow:workflow-step-api:jar:2.13: Could not transfer artifact org.jenkins-ci.plugins:plugin:pom:2.33 from/to repo.jenkins-ci.org (https://repo.jenkins-ci.org/public/): Transfer failed for https://repo.jenkins-ci.org/public/org/jenkins-ci/plugins/plugin/2.33/plugin-2.33.pom: Connect to repo.jenkins-ci.org:443 [repo.jenkins-ci.org/130.211.20.35] failed: Connection timed out (Connection timed out) -> [Help 1]

slide:
> The test failures on Windows seem infra related (maybe mingit is broken on Windows ACI agents?), they all fail with this error:
> 
> > C:\mingit\usr\bin\sh.exe: *** fatal error - unable to load netapi32.dll, Win32 error 126
> 
> Not sure if I can fix the issues here, or what the root cause is. I'll try switching to non-ACI agents to see if that makes a difference in case the issue is related to [jenkins-infra/pipeline-library#144](https://github.com/jenkins-infra/pipeline-library/pull/144).

Can you see if an ACI build will work now? I updated the image with a possible fix.

dwnusbaum:
@slide it looks like the Git-related errors on the ACI agents are fixed now, thank you!

`PersistenceProblemsTest.inProgressButProgramLoadFailure` is timing out in CI. It is also timing out for me locally, but not every time. I'll try to figure out what is happening. I suspect that the proximate cause is that the workflow-job dependency was updated by pulling in the plugin BOM, but I'm not sure.

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -313,7 +312,11 @@ public void inProgressButProgramLoadFailure() throws Exception {
         story.thenWithHardShutdown( j -> {
             WorkflowRun run = runBasicPauseOnInput(j, DEFAULT_JOBNAME, build);
             CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
-            cpsExec.getProgramDataFile().delete();
+            // Wait until program.dat is written and then delete it.
+            while (!Files.exists(cpsExec.getProgramDataFile().toPath())) {
+                Thread.sleep(100);
+            }
+            Files.delete(cpsExec.getProgramDataFile().toPath());
```
dwnusbaum, line 319: Previously, the call to `delete` was sometimes running before `program.dat` had been written to disk, so the test would fail because the Pipeline resumed normally after the restart and just hung at an `input` step.

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -166,8 +166,7 @@ public void completedFinalFlowNodeNotPersisted() throws Exception {
 
             // Hack but deletes the file from disk
             CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
-            File f = new File(cpsExec.getStorageDir(), finalId+".xml");
-            f.delete();
+            Files.delete(cpsExec.getStorageDir().toPath().resolve(finalId+".xml"));
```
dwnusbaum, line 169: This test wasn't failing, but I switched to `Files.delete` so that the test fails immediately with an exception if the file does not exist when we try to delete it.

dwnusbaum:
Hmm looks the failures this time were infra issues with Git on Windows (CC @slide):

> hudson.plugins.git.GitException: Command "git.exe fetch --tags --force --progress -- file:///C:/Users/jenkins/Work/workspace/ugins_workflow-cps-plugin_PR-355/target/tmp/junit7871276158289604478/junit4270882418731441144/ +refs/heads/*:refs/remotes/origin/*" returned status code 128:
stdout: 
stderr: fatal: '/C:/Users/jenkins/Work/workspace/ugins_workflow-cps-plugin_PR-355/target/tmp/junit7871276158289604478/junit4270882418731441144/' does not appear to be a git repository

Trying a rebuild now.

slide:
Strange, I'll try running some stuff locally

slide:
I'm looking into the failures, it's just slow going.

dwnusbaum:
@slide No worries, I will just use non-ACI agents for now.

PR #356, opened 2020-05-07
Test case for JENKINS-57253

basil:
Adding a test case for [JENKINS-57253](https://issues.jenkins-ci.org/browse/JENKINS-57253). The last assertion currently fails due to a leaked executor, as described in the bug, so I have marked the test with `@Ignore`. When the bug is fixed, the `@Ignore` annotation can be removed.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
@@ -45,12 +45,14 @@ import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
 import org.jenkinsci.plugins.workflow.steps.StepExecutions;
+import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 
 import static org.junit.Assert.*;
 
 import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.ClassRule;
+import org.junit.Ignore;
 import org.junit.Rule;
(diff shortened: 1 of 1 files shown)


dwnusbaum:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/356 for the fix.

dwnusbaum:
Also, did you happen to check whether things work correctly if you do put a `loop` or `switch` around the `node` step? If that causes problems, then we also need some kind of runtime behavior change in groovy-cps.

basil:
Thank _you_ for the fix!

> Actually, I am pretty sure that `break` would be disallowed in something like `for { node { break; } }` because the closure is a new scope as far as `break` and `continue` are concerned.

Indeed, I confirmed that `for { node { break; } }` fails at the compilation stage as you predicted.

car-roll:
@basil fyi I merged https://github.com/cloudbees/groovy-cps/pull/109 but I don't think it's getting released until we get the repo moved into `jenkinsci`.
I noticed that this test did not have an `@Ignore`. Is this issue already fixed via some other way?

basil:
> @basil fyi I merged [cloudbees/groovy-cps#109](https://github.com/cloudbees/groovy-cps/pull/109) but I don't think it's getting released until we get the repo moved into `jenkinsci`. I noticed that this test did not have an `@Ignore`. Is this issue already fixed via some other way?

Sorry, when I was merging with `master` I accidentally added some unrelated changes to this. I just corrected that. It now has the right set of changes with `@Ignore`.

car-roll:
@basil what  confused  me though was that the checks showed green. Did this test actually  pass without any fix?

basil:
> @basil what confused me though was that the checks showed green. Did this test actually pass without any fix?

They were green because I had posted an unrelated set of changes that happened to pass.

car-roll (approved):

PR #359, opened 2020-06-01
Merge groovy-cps into workflow-cps

dwnusbaum:
This PR merges https://github.com/cloudbees/groovy-cps into this repository to simplify development (for example I do not actually have a way to release groovy-cps right now outside of security fixes). If we decide to merge this PR, we would archive https://github.com/cloudbees/groovy-cps replace its contents with a link to this repository.

I tried to maintain the history of groovy-cps by merging it into this repository rather than just copying the source code, so most of the commits in this PR can be ignored. The interesting commits are 9409e1e and 531b438. EDIT: I was hoping that GitHub would display the combined diff for 531b438, but it looks like it is diffing against the first parent instead... Maybe it would have been better to do a minimal merge that only addressed the conflicts even though things weren't compiling so I could fix everything else in a separate commit to make it easier to review? Or I guess I could have gone to groovy-cps and made those changes in `master` first.

Just opening a draft for now in case some issues come up when running the tests.

TODO before merge:
* [ ] Test the changes
    * Looks like `groovy-cps-dgm-builder` cannot run on Java 9+ (see https://github.com/jenkinsci/workflow-cps-plugin/pull/360#issuecomment-638983758 and the later comment), so this PR would mean that workflow-cps could no longer be built using Java 9+, which will complicate testing, especially for things like the PCT. 
* [ ] Make sure nothing is broken in terms of packaging and we aren't bundling any new dependencies in the plugin
* [ ] Is it ok that groovy-cps is using the Apache 2.0 license while workflow-cps is using the MIT license?
* [ ] Add components to repository-permissions updater for the new parent, groovy-cps, and groovy-cps-dgm-builder


lib/src/main/java/com/cloudbees/groovy/cps/green/GreenThread.java
```diff
@@ -0,0 +1,199 @@
...
+import com.cloudbees.groovy.cps.Outcome;
+import com.cloudbees.groovy.cps.impl.ThrowBlock;
+import java.io.Serializable;
+
+/**
+ * Represents a green thread.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+public abstract class GreenThread implements Runnable, Serializable {
```
dwnusbaum, line 16: SpotBugs didn't like that `GreenThreadState` is serializable and this class was not since `GreenThreadState.g` is a `GreenThread`. I added `Serializable`, but I don't think the `green` package is used at all, maybe it should just be deleted?

lib/src/main/java/com/cloudbees/groovy/cps/impl/LocalVariableBlock.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @author Kohsuke Kawaguchi
+ */
+public class LocalVariableBlock extends LValueBlock {
+    private final String name;
+    private SourceLocation loc;
+
+    public LocalVariableBlock(SourceLocation loc, String name) {
+        this.name = name;
+        this.loc = loc;
```
dwnusbaum, line 21: SpotBugs noticed that `loc` was never actually assigned in this class, which seemed like a bug, so I added this line. Maybe it will improve the quality of some error messages? Or maybe `loc` was left `null` intentionally and I should suppress the error instead?

lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Arguments to the call.
+     */
+    private final Block[] argExps;
+
+    private final SourceLocation loc;
+
+    private final boolean safe;
+
+    @SuppressFBWarnings(value = "EI_EXPOSE_REP2", justification = "argExps is not modified by existing callers after being passed to this method and we want to avoid cloning every array")
```
dwnusbaum, line 45: SpotBugs didn't like that we are directly storing the mutable `argExps` array passed in as a parameter. This code is only used internally, and nothing ever tries to mutate `argExps`, so I think it is fine to suppress the warning, especially since we don't really want to increase the overhead of this code.

lib/src/test/java/com/cloudbees/groovy/cps/BasicTest.java
```diff
@@ -0,0 +1,324 @@
...
+        } catch (InvocationTargetException x) {
+            throw new AssertionError(x);
+        }
+    }
+
+
+    // 3    => 3
+    @Test
+    public void constant() {
+        assertThat(run(b.constant(3)), equalTo(3));
```
dwnusbaum, line 44: Updating to Java 8 caused this class to fail to compile due to ambiguity between `assertEquals(Object, Object)` and `assertEquals($primitive, $primitive)` in various tests. To fix it, I switched to using `assertThat` and `equalTo`, which should be more or less the same as using the `(Object, Object)` overloads of `assertEquals`.

dwnusbaum:
Looks like there is an issue running `groovy-cps-dgm-builder` during the build of `groovy-cps`:

```
Exception in thread "main" java.lang.NoClassDefFoundError: com/sun/tools/javac/file/ZipArchive
 	at com.cloudbees.groovy.cps.tool.Driver.run(Driver.java:49)
 	at com.cloudbees.groovy.cps.tool.Driver.main(Driver.java:29)
 Caused by: java.lang.ClassNotFoundException: com.sun.tools.javac.file.ZipArchive
 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
 	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
 	... 2 more
```

dwnusbaum:
Actually, it looks like `groovy-cps-dgm-builder` should have all of those dependencies in its jar-with-dependencies because of the `sorcerer-javac` dependency. I see `com/sun/tools/javac/file/ZipArchive.class` in the jar locally, so I'm not sure why they wouldn't be found on the CI build, and I think a plain JRE should be good enough to run `groovy-cps-dgm-builder`.

dwnusbaum:
> Remote call on EC2 (aws) - Windows 2019 (i-0fbd89ef00515ecbb) failed. The channel is closing down or has closed down

lib/src/main/java/com/cloudbees/groovy/cps/green/GreenThread.java
jglick: I do not think this package was ever used. Looks like it can be deleted.

jglick:
> a plain JRE should be good enough to run `groovy-cps-dgm-builder`.

Possibly requires a JDK; I think the difference is that the CI build is using Java 11, and this was only ever intended to be run from 8.

dwnusbaum:
> Possibly requires a JDK; I think the difference is that the CI build is using Java 11, and this was only ever intended to be run from 8.

Yes, I spent some time in #360 realizing that `groovy-cps-dgm-builder` requires Java 8, and it is unclear to me if there is any way to get it running against both Java 8 and Java 9+, or even what exactly is needed to get it both building and running against Java 9+ in the first place.

For now I am going to close this PR because I do not think I will have capacity to look into it further anytime soon.

PR #360, opened 2020-06-04
Investigating CI failures in #359

dwnusbaum:
A clone of #359 that I am using to try to figure out why the CI builds are failing in that PR so I don't clutter up the history in that PR.


dwnusbaum:
Ok, my best guess is that module-related changes in JDK 9+ make it so that the copy of `javac` in https://github.com/kohsuke/sorcerer/javac cannot be loaded correctly (maybe because our classes are in the unnamed module, but the `com.sun.tools` package is in a module, so we have a split package situation?). Maybe we could shade the dependencies in the jar-with-dependencies to fix the issue, or use `--upgrade-module-path` or `--patch-module` on Java 11?

dwnusbaum:
Shading is no good because it breaks `sorcerer-javac`, probably because the package names are important. I spent more time than I intended yesterday trying to see if we could replace our uses of internal APIs with only the public APIs from what become the `java.compiler` and `jdk.compiler` modules in Java 9+, but I couldn't figure out how to get things to work with both Java 8 and Java 9+, and I'm not sure if everything we need is even available. Probably the best solution is to just require Java 8 for now and leave things as-is, although that means we can't easily test on Java 9+. Maybe with some effort we could hack up sorcerer-javac enough to get it working on Java 9+, but at that point it might just be easier to maintain `CpsDefaultGroovyMethods.java` and friends as source code in groovy-cps itself.

PR #362, opened 2020-06-16
Expose assertGenerateSnippet to pipeline-model-definition

Vlatombe:
Essentially DRY for [DirectiveGeneratorTest](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/d9f665d27c269f8294e3897f54f1f7d1f3d6a8e5/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java#L587-L602). Upstream of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/396

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
@@ -98,7 +98,7 @@ import org.kohsuke.stapler.StaplerRequest;
      *
      * @param o The object to translate.
      * @param nestedExp
-     *      true if this object is written as a nested expression (in which case we always produce parenthesis for readability)
+     *      true if this object is written as a nested expression (in which case we always produce parentheses for readability)
      * @return A string translation of the object.
      */
     public static String object2Groovy(Object o, boolean nestedExp) throws UnsupportedOperationException {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
@@ -89,7 +89,20 @@ import org.kohsuke.stapler.StaplerRequest;
      * @return A string translation of the object.
      */
     public static String object2Groovy(Object o) throws UnsupportedOperationException {
-        return object2Groovy(new StringBuilder(), o, false).toString();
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -89,7 +89,20 @@ static String step2Groovy(Step s) throws UnsupportedOperationException {
...
+        return object2Groovy(o, false);
+    }
+
+    /**
+     * Publicly accessible version of {@link #object2Groovy(StringBuilder, Object, boolean)} that translates an object into
+     * the equivalent Pipeline Groovy string.
+     *
+     * @param o The object to translate.
+     * @param nestedExp
+     *      true if this object is written as a nested expression (in which case we always produce parenthesis for readability)
```
jglick, line 101:
```suggestion
     *      true if this object is written as a nested expression (in which case we always produce parentheses for readability)
```

dwnusbaum:
Hmm, I have seen `RestartingLoadStepTest.updateBindingsOnRestart` flake a few times recently. I think that after the restart the `PlaceholderExecutable` for the `node` step is still present in the program and starts executing (not sure exactly why, based on the logs it looks like it already completed, maybe `program.dat` isn't saved consistently?), and so [this line in `PlaceholderExecutable.run`](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/f45d07a56a63c83f50f048453d91dcf166171c57/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L798) fails since the flow node for the `node` step is already marked as completed.

Vlatombe:
Yeah it passed in https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-362/2/ anyway. Just wanted to have a clean green status, looks like it's not happening.

Vlatombe:
@dwnusbaum do you plan on cutting a release for this plugin soon? I'd like to move forward with the downstream changes

dwnusbaum:
@Vlatombe Yes, I was hoping to get #361 and #363 into the next release, but there have been a ton of infra issues lately so the builds aren't passing.

PR #363, opened 2020-06-18
[JENKINS-55287] Improve error messages for nonresumable Pipelines

dwnusbaum:
See [JENKINS-55287](https://issues.jenkins-ci.org/browse/JENKINS-55287). Subsumes https://github.com/jenkinsci/workflow-cps-plugin/pull/354 (and will subsume https://github.com/jenkinsci/workflow-cps-plugin/pull/234 eventually).

Based on the cases I've seen, I think the root cause of JENKINS-55287 is that Jenkins is crashing, so it is expected that PERFORMANCE_OPTIMIZED Pipelines are unable to resume. From that perspective, the issue is more that the error message is misleading and causes users to think they have hit a bug in Pipeline, when really this case is expected and the fix is to investigate and understand why Jenkins is crashing.

The field `CpsFlowExecution.persistedClean` was supposed to help detect these cases and print a more relevant error message, but there were two issues:
* It was always set to true right after a build started because of [this call to `CpsFlowExecution.saveOwner`](https://github.com/jenkinsci/workflow-cps-plugin/blob/455249a3b9b65450e4de0739a46668e2f5c1202e/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L537) that occurs right after the build starts. In the cases of JENKINS-55287 I have seen, the only persisted flow node is always the `FlowStartNode`, meaning that this is the only time the build was ever saved, and so the build is always considered resumable unless Jenkins shuts down normally and there is an error while saving the build in `CpsFlowExecution.suspendAll`. To fix this we only modify `persistedClean` if Jenkins is shutting down.
* The logic that checks if a Pipeline can be resumed used to only happen _after_ the flow graph was initialized, but in most (all?) cases the flow graph will be incomplete and so that will fail. That means that we also need to check if the Pipeline is resumable either before trying to initializing flow nodes (pessimistic) and fail immediately, or after we receive an error loading flow nodes (optimistic) and then change the reported error as necessary. **EDIT**: Part of the problem here is that the flow graph is persisted after every step for `PERFORMANCE_OPTIMIZED` pipelines, see https://github.com/jenkinsci/workflow-cps-plugin/pull/234. I don't think it makes sense to try to update the logic here without fixing that first, so I will look into fixing that separately.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -2001,7 +2001,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             // Nothing to persist OR we've already persisted it along the way.
             return;
         }
-        LOGGER.log(Level.INFO, "Attempting to save a checkpoint of {0} before shutdown", this);
+        LOGGER.log(Level.INFO, "Attempting to checkpoint all data for {0}{1}", new Object[] {
+            this, (shuttingDown ? " before shutdown" : "")
+        });
         boolean persistOk = true;
         FlowNodeStorage storage = getStorage();
         if (storage != null) {
@@ -2077,9 +2079,13 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         }
 
         if (persistOk) {
-            LOGGER.log(Level.INFO, "Successfully saved a checkpoint of {0} before shutdown", this);
+            LOGGER.log(Level.INFO, "Successfully checkpointed {0}{1}", new Object[] {
+                this, (shuttingDown ? " before shutdown" : "")
+            });
         } else {
-            LOGGER.log(Level.WARNING, "Unable to save a checkpoint of {0} before shutdown. The build will probably fail when Jenkins restarts.", this);
+            LOGGER.log(Level.WARNING, "Unable to successfully checkpoint {0}{1}", new Object[] {
+                this, (shuttingDown ? " before shutdown, so this build will probably fail when Jenkins restarts" : "")
+            });
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -745,8 +745,10 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                     saveOwner();
                 }
             } else {  // See if we can/should resume build
+                // Always set up the shell in case the build is not resumable and loadProgramFailed is called without loadProgramAsync being called first.
+                scriptClass = parseScript().getClass();
                 if (canResume()) {
-                    // Before we resume, we need to unset persistedClean in case Jenkins restarts again.
+                    // Before we resume, we need to unset persistedClean (and persist the change) in case Jenkins restarts again.
                     persistedClean = null;
                     saveOwner();
                     loadProgramAsync(getProgramDataFile());
@@ -780,8 +782,6 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         programPromise = result;
 
         try {
-            scriptClass = parseScript().getClass();
-
             final RiverReader r = new RiverReader(programDataFile, scriptClass.getClassLoader(), owner);
             Futures.addCallback(
                     r.restorePickles(pickleFutures = new ArrayList<>()),
@@ -1483,6 +1483,10 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             } finally {
                 if (synchronous) {
                     bc.abort(); // hack to skip save—we are holding a lock
+                } else if (!getDurabilityHint().isPersistWithEveryStep() && !isComplete()
+                        && nodes.stream().noneMatch(node -> node instanceof FlowStartNode || node instanceof FlowEndNode)) {
+                    // Do not save in PERFORMANCE_OPTIMIZED mode unless this is the end of the build.
+                    bc.abort();
                 } else {
                     try {
                         bc.commit();
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -746,10 +746,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 }
             } else {  // See if we can/should resume build
                 if (canResume()) {
-                    // TODO: I think we need to null out persistedClean here and resave the execution so that future
-                    // resumptions are not misled by the fact that persistedClean was true before this resumption.
-                    // CpsThreadGroup.run nulls out persistedClean after every step but it does not save that value
-                    // for PERFORMANCE_OPTIMIZED Pipelines.
+                    // Before we resume, we need to unset persistedClean in case Jenkins restarts again.
+                    persistedClean = null;
+                    saveOwner();
                     loadProgramAsync(getProgramDataFile());
                 } else {
                     // TODO if possible, consider trying to close out unterminated blocks to keep existing graph history
@@ -1560,6 +1559,10 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                     checkAndAbortNonresumableBuild();  // TODO Verify if we can rely on just killing paused builds at shutdown via checkAndAbortNonresumableBuild()
                     checkpoint(false);
                 } else {
+                    // Pausing the build sets persistedClean to true so the build can resume, so if we unpause the build
+                    // we need to unset persistedClean again.
+                    persistedClean = null;
+                    saveOwner();
                     g.unpause();
                 }
                 try {
@@ -2049,12 +2052,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 persistOk=false;
                 LOGGER.log(Level.WARNING, "Error persisting FlowNode storage for: " + this, ioe);
             }
-            if (shuttingDown) {
-                // Only modify persistedClean if Jenkins is shutting down.
-                // TODO: We could set persistedClean to true when the build is just being paused, but then we would need
-                // to null it out and save again when the build was unpaused.
-                persistedClean = persistOk;
-            }
+            persistedClean = persistOk;
             try {
                 saveOwner();
             } catch (Exception ex) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1558,7 +1558,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 if (v) {
                     g.pause();
                     checkAndAbortNonresumableBuild();  // TODO Verify if we can rely on just killing paused builds at shutdown via checkAndAbortNonresumableBuild()
-                    checkpoint();
+                    checkpoint(false);
                 } else {
                     g.unpause();
                 }
@@ -1601,7 +1601,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                                 LOGGER.log(Level.WARNING, "Error waiting for Pipeline to suspend: "+exec, ex);
                             }
                         }
-                        cpsExec.checkpoint();
+                        cpsExec.checkpoint(true);
                     }
                 } catch (Exception ex) {
                     LOGGER.log(Level.WARNING, "Error persisting Pipeline execution at shutdown: "+((CpsFlowExecution) execution).owner, ex);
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -712,17 +712,26 @@ public void onLoad(FlowExecutionOwner owner) throws IOException {
         this.owner = owner;
 
         try {
+            initializeStorage();  // Throws exception and bombs out if we can't load FlowNodes
+        } catch (Exception ex) {
             try {
-                initializeStorage();  // Throws exception and bombs out if we can't load FlowNodes
-            } catch (Exception ex) {
-                LOGGER.log(Level.WARNING, "Error initializing storage and loading nodes, will try to create placeholders for: "+this, ex);
+                if (!canResume()) {
```
dwnusbaum, line 718: Alternatively, we could check this even before we attempt to initialize storage and just bail out right away. I don't know if that is too pessimistic, maybe there is a case where the Pipeline is not resumable but flow nodes are all up to date and so the optimistic behavior here is useful because you will still be able to see the existing flow graph in those cases?

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -712,17 +712,26 @@ public void onLoad(FlowExecutionOwner owner) throws IOException {
...
+                    // usually that only the FlowStartNode has been persisted and so we cannot find the current head node in storage.
+                    String message = "Unable to resume " + owner.getExecutable() + " because it was not saved before Jenkins shut down. Did Jenkins crash?";
+                    LOGGER.log(Level.WARNING, message, ex);
+                    ex = new AbortException(message); // Change the exception displayed in the build log to one that is more useful for users.
+                } else {
+                    LOGGER.log(Level.WARNING, "Unable to resume " + owner.getExecutable() + " because its flow nodes could not be loaded", ex);
+                }
                 createPlaceholderNodes(ex);
-                return;
+            } catch (Exception ex2) {
```
dwnusbaum, line 728: This is all mostly refactoring. The old code had this structure:
```
try {
  try {
    // code that can throw
  } catch (first) {
    // code that can throw
    return
  }
} catch (second) {
   throw second
}
```

Now it has this structure:

```
try {
  // code that can throw
} catch (first) {
  try {
    // code that can throw
  } catch (second) {
    second.addSuppressed(first)
    throw second
  }
  return
}
```

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -737,12 +746,17 @@ public void onLoad(FlowExecutionOwner owner) throws IOException {
                 }
             } else {  // See if we can/should resume build
                 if (canResume()) {
+                    // TODO: I think we need to null out persistedClean here and resave the execution so that future
+                    // resumptions are not misled by the fact that persistedClean was true before this resumption.
+                    // CpsThreadGroup.run nulls out persistedClean after every step but it does not save that value
+                    // for PERFORMANCE_OPTIMIZED Pipelines.
```
dwnusbaum, line 752: Maybe possible to set a transient field after a restart so that `CpsThreadGroup.run` saves the build even for PERFORMANCE_OPTIMIZED Pipelines? Wouldn't be perfect, but would also address the issue noted in `CpsFlowExecution.checkpoint`.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1526,10 +1540,6 @@ public boolean isPaused() {
         return false;
     }
 
-    private void setPersistedClean(boolean persistedClean) {  // Workaround for some issues with anonymous classes.
```
dwnusbaum, line 1529: Unused

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1951,24 +1961,35 @@ void saveExecutionIfDurable() {
      *  Key note: to avoid deadlocks we need to ensure that we don't hold a lock on this CpsFlowExecution when running saveOwner
      *   or pre-emptively lock the run before locking the execution and saving. */
     void saveOwner() {
+        /**
+         * We only want to update persistedClean if the program is over (in which case it doesn't matter), or if Jenkins
+         * is about to shut down (in which case it controls whether PERFORMANCE_OPTIMIZED Pipelines can be resumed).
+         */
+        boolean updatePersistedClean = done || Jenkins.get().isTerminating();
```
dwnusbaum, line 1968: TODO: Do we even care about changing it if the build is done? Also, I did not actually check how `Jenkins.isTerminating` works, so I should probably make sure it is true when terminators (`CpsFlowExecution.suspendAll`) are running.

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -244,7 +243,6 @@ public void inProgressNormal() throws Exception {
     }
 
     @Test
-    @Ignore
```
dwnusbaum, line 247: These tests have been ignored since they were added. I am not sure why, but I went ahead and unignored them.

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -263,14 +261,14 @@ public void inProgressMaxPerfCleanShutdown() throws Exception {
         });
     }
 
+    @Issue("JENKINS-55287")
     @Test
-    @Ignore
     public void inProgressMaxPerfDirtyShutdown() throws Exception {
```
dwnusbaum, line 266: This is the exact situation reported in JENKINS-55287 as far as I can tell.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -2031,7 +2052,12 @@ public void onFailure(Throwable t) {
                 persistOk=false;
                 LOGGER.log(Level.WARNING, "Error persisting FlowNode storage before shutdown", ioe);
             }
-            persistedClean = persistOk;
+            if (Jenkins.get().isTerminating()) {
+                // Only modify persistedClean if Jenkins is shutting down.
+                // TODO: We could set persistedClean to true when the build is just being paused, but then we would need
+                // to null it out and save again when the build was unpaused.
```
dwnusbaum, line 2058: Looks like `FlowDurabilityTest.testPauseForcesLowDurabilityToPersist` is failing because it expects that pausing the build will make it be able to resume, so I will need to think about this.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: Ha, ok so `Jenkins.isTerminating` is only true _right after_ terminators run, see [Jenkins.cleanUp](https://github.com/jenkinsci/jenkins/blob/dd6c322dec260e41a0020b302378af85f79e1ee4/core/src/main/java/jenkins/model/Jenkins.java#L3417-L3419). Not a big deal, we'll just have to pass a flag to `saveOwner` when called from `suspendAll`.

This is why `FlowDurabilityTest.testDurableAgainstCleanRestartSurvivesIt` is failing.

src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
basil: You probably already knew this, but I just noticed that these test cases seem to be duplicated in `workflow-job-plugin`'s [`CpsPersistenceTest`](https://github.com/jenkinsci/workflow-job-plugin/blob/f369d5bdfc39ac902f544a125e971414cb51a463/src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java). Once the dust settles, it might be worth doing a sweep of the two copies to ensure that they are in sync with each other.

dwnusbaum: Yeah, I think I will probably delete the copies in workflow-job after this. Ideally we would run the PCT in the CI builds here with just the core Pipeline plugins so that changes are tested against the newest versions of everything.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -737,12 +746,16 @@ public void onLoad(FlowExecutionOwner owner) throws IOException {
                 }
             } else {  // See if we can/should resume build
                 if (canResume()) {
+                    // Before we resume, we need to unset persistedClean in case Jenkins restarts again.
+                    persistedClean = null;
+                    saveOwner();
```
dwnusbaum, line 751: Should add a test case for this. The test would use a PERFORMANCE_OPTIMIZED Pipeline, restart Jenkins normally, then restart Jenkins again with a hard shutdown and make sure we get the error message that the Pipeline can't be resumed rather than a raw exception.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1548,8 +1557,12 @@ public void pause(final boolean v) throws IOException {
                 if (v) {
                     g.pause();
                     checkAndAbortNonresumableBuild();  // TODO Verify if we can rely on just killing paused builds at shutdown via checkAndAbortNonresumableBuild()
-                    checkpoint();
+                    checkpoint(false);
                 } else {
+                    // Pausing the build sets persistedClean to true so the build can resume, so if we unpause the build
+                    // we need to unset persistedClean again.
+                    persistedClean = null;
+                    saveOwner();
```
dwnusbaum, line 1565: We should add a test case for this. The Pipeline would need to be paused, then unpaused, and then we do a hard shutdown and make sure we get the error message that the Pipeline can't be resumed rather than a raw exception.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1972,72 +1984,95 @@ void saveOwner() {
...
+    /**
+     * Save everything we can to disk - program, run, flownodes.
+     * @param shuttingDown True if this checkpoint is happening because Jenkins is shutting down, false if it is happening because execution was paused.
+     */
+    private void checkpoint(boolean shuttingDown) {
         if (isComplete() || this.getDurabilityHint().isPersistWithEveryStep()) {
             // Nothing to persist OR we've already persisted it along the way.
             return;
         }
+        LOGGER.log(Level.INFO, "Attempting to checkpoint all data for {0}{1}", new Object[] {
```
dwnusbaum, line 1996: The logging-related changes were originally part of #354 (there are some slight modifications here), so it might be easier to review that PR first.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: Addressed in https://github.com/jenkinsci/workflow-cps-plugin/pull/363/commits/2214c421c47c01e84a2c7218912f0587799f8252.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: Changed in https://github.com/jenkinsci/workflow-cps-plugin/pull/363/commits/b360ae0cf8547bf0d56bfc36b8c39aaf10a04aa0 and https://github.com/jenkinsci/workflow-cps-plugin/pull/363/commits/eb08b5b6488961ce5c011eee774c5389a67c5d93.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: Addressed by https://github.com/jenkinsci/workflow-cps-plugin/pull/363/commits/2214c421c47c01e84a2c7218912f0587799f8252.

dwnusbaum:
> Authorization failed for https://repo.jenkins-ci.org/public/org/jenkins-ci/plugins/plugin/4.1/plugin-4.1.pom 403 Forbidden

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: I came up with a test case, but it doesn't fail without this line because of a random save that happens when `CpsFlowExecution.notifyListeners` is called. IIUC, https://github.com/jenkinsci/workflow-cps-plugin/pull/234 was trying to stop that save from occurring for PERFORMANCE_OPTIMIZED Pipelines, so I will look into making related changes in this PR.

dwnusbaum:
I don't remember exactly what was wrong with this approach, but back in February I spent a but more time looking into this and decided it would be better to just immediately throw an exception from `CpsFlowExecution.onLoad` if `CpsFlowExecution.canResume` was false to avoid issues with a partially-initialized `CpsFlowExecution`, especially since in these scenarios the data that was persisted is unlikely to be useful anyway since it will be very incomplete.

See my [JENKINS-55287-2](https://github.com/jenkinsci/workflow-cps-plugin/compare/master...dwnusbaum:JENKINS-55287-2) branch for what that approach looked like. I can't remember why I didn't file a PR for that branch at the time. I think that I might have thought we needed a simultaneous patch to `WorkflowRun.getExecution` in `workflow-job` to clean up the error handling so that we don't have multiple exceptions printed and to make it possible for the error message from the exception thrown in `workflow-cps` to be printed to the build logs rather than just showing up in the Jenkins system logs. I also think I wanted to do some more testing in relation to #234, since both of my fixes to improve the log messages pick up that fix as well, and I recall that we saw some test failures when that patch was filed originally. I think I also thought a few more tests were needed to check some of the edge cases that I noticed while investigating this issue.

PR #364, opened 2020-06-19
Add SnippetizerTester#assertParseStep

Vlatombe:
I find this useful to ensure backward compatibility when changing
symbols or step parameter names.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
@@ -105,11 +105,23 @@ public class SnippetizerTester {
      */
     public void assertRoundTrip(Step step, String expected) throws Exception {
         assertEquals(expected, Snippetizer.step2Groovy(step));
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTester.java
```diff
@@ -105,11 +105,23 @@ protected void assertGenerateSnippet(@Nonnull String url, @Nonnull String json,
      */
     public void assertRoundTrip(Step step, String expected) throws Exception {
         assertEquals(expected, Snippetizer.step2Groovy(step));
+        assertParseStep(step, expected);
+    }
+
+    /**
+     * Given a DSL script, make sure it gets parsed a the provided step object.
```
dwnusbaum, line 112:
```suggestion
     * Given a DSL script, make sure it gets parsed as the provided step object.
     * Useful for testing backwards compatibility when changing symbols or step parameter names. In other cases prefer {@link #assertRoundTrip}.
```

PR #366, opened 2020-07-17
 Try to diagnose why RestartingLoadStepTest.updatedBindingsOnRestart is flaky

dwnusbaum:
I recently received a bug report with similar symptoms as described in the `@Ignore` message added to  `RestartingLoadStepTest.updatedBindingsOnRestart` in https://github.com/jenkinsci/workflow-cps-plugin/pull/363/commits/b3a6847717d6962f6e3b509f4ed1679c2e766523 (from https://github.com/jenkinsci/workflow-cps-plugin/pull/363, same flake was also mentioned in https://github.com/jenkinsci/workflow-cps-plugin/pull/362#issuecomment-645430600), so I want to investigate in case the flakes we've seen recently are related.

PR #367, opened 2020-07-22
[JENKINS-63164] Clear CpsBodyExecution.thread when the body completes

dwnusbaum:
See [JENKINS-63164](https://issues.jenkins-ci.org/browse/JENKINS-63164). Kind of reimplements #245 after #279 replaced it with a different approach that works better than #245 but only affects the `parallel` and `load` steps.

JENKINS-63164 is the root cause of `RestartingLoadStepTest.updatedBindingsOnRestart` being flaky, see #366.

I am really not sure about the best way to fix the issue. See https://github.com/jenkinsci/workflow-cps-plugin/pull/367#discussion_r458869636 for some discussion of other options.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -325,8 +325,16 @@ class CpsBodyExecution extends BodyExecution {
 
     private void setOutcome(Outcome o) {
         synchronized (this) {
-            if (bodyToUnexport != null && thread != null) {
-                thread.group.unexport(bodyToUnexport);
+            if (thread != null) {
+                if (bodyToUnexport != null) {
+                    thread.group.unexport(bodyToUnexport);
+                }
+                // onSuccess and onFailure are part of the program and reference the CpsBodyExecution as their outer
+                // class. They are usually unreachable from the program once the body completes, but in some cases they
+                // may still be reachable (e.g. closures that outlive the body). If the CpsBodyExecution is still part
+                // of the program, we need to make sure that we do not reference to a dead CpsThread because that may
+                // cause resumption issues (JENKINS-63164).
+                thread = null;
             }
             if (outcome!=null)
                 throw new IllegalStateException("Outcome is already set");
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
@@ -12,7 +12,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
-import static org.hamcrest.Matchers.*;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepNode;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -325,8 +325,16 @@ public synchronized Object get(long timeout, TimeUnit unit) throws InterruptedEx
...
+            if (thread != null) {
+                if (bodyToUnexport != null) {
+                    thread.group.unexport(bodyToUnexport);
+                }
+                // onSuccess and onFailure are part of the program and reference the CpsBodyExecution as their outer
+                // class. They are usually unreachable from the program once the body completes, but in some cases they
+                // may still be reachable (e.g. closures that outlive the body). If the CpsBodyExecution is still part
+                // of the program, we need to make sure that we do not reference to a dead CpsThread because that may
+                // cause resumption issues (JENKINS-63164).
+                thread = null;
```
dwnusbaum, line 337: I'm not sure about this. Some other ideas I had:

* Update `CpsThreadGroup.run` so that when it removes a thread from `CpsThreadGroup.threads` it calls some new method `CpsThread.destroy` or similar that nulls out fields (mainly `contextVariables`) in `CpsThread` in case something is still referencing it. Might fix other issues as well that we aren't aware of.
* Modify `groovy-cps` so that the `Env` captured by `CpsClosure` does not reference things like `CallEnv.returnAddress` from nested environments. Not sure about correctness/feasibility of this, seems complicated. At least in the simple case here we only need the locals from the nested environments and if we only stored those variables there would be no issues, but perhaps there could be a references from the locals to some of the fields in `CallEnv`, and maybe we need to maintain the actual `Env` references as-is for semantics.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -214,14 +217,36 @@ public boolean start() throws Exception {
         });
         rr.then(r -> {
             WorkflowRun b = r.jenkins.getItemByFullName("demo", WorkflowJob.class).getBuildByNumber(1);
+            SemaphoreStep.success("wait/1", null);
+            r.assertBuildStatusSuccess(r.waitForCompletion(b));
```
dwnusbaum, line 221: Weird diff, but all I did here was removed the redundant call to `SemaphoreStep.waitForStart` and the comment/code regarding "Before the fix for JENKINS-53709" since https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/104 changed the behavior.

dwnusbaum:
Incrementals publisher failed (`InterruptedException` in `BourneShellScript.launchWithCookie`) but the tests are passing so I am marking this as ready for review.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
bitwiseman: The first option sounds better to me.  Is there a downside to that?

dwnusbaum: > The first option sounds better to me. Is there a downside to that?

I'm not really sure, I can prototype it. If `CpsBodyExecution.thread` is actually usable as a captured variable in the closure, then nulling out all of the fields in `CpsThread` could result in some weird behavior/NPEs, but I suspect that the `CpsThread` is not actually usable.

dwnusbaum:
Closing in favor of https://github.com/jenkinsci/workflow-cps-plugin/pull/368.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
dwnusbaum: Noting some other ideas that came up in a discussion today in case they are useful to someone in the future:

* Replacing `CpsBodyExecution.thread` with `CpsBodyExecution.threadId` and only holding a transient reference to the `CpsThread` itself, loading the value by looking up the `threadId` in `CpsThreadGroup`. Not sure if this would be possible, and would require some adjustments to maintain serialization compatiblity.
* Hooking into `RiverWriter` or similar code in `workflow-support` and just not serializing `CpsThread`s that are not part of the root `CpsThreadGroup`. Seems complicated, and I'm not sure about compatibility.

PR #368, opened 2020-07-22
[JENKINS-63164] Clear fields in CpsThread when thread is removed from CpsThreadGroup

dwnusbaum:
See #367. This is the first alternative discussed in https://github.com/jenkinsci/workflow-cps-plugin/pull/367#discussion_r458869636.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -194,7 +194,7 @@ public final class CpsThreadGroup implements Serializable {
     }
 
     @CpsVmThreadOnly
-    public CpsThread addThread(Continuable program, FlowHead head, ContextVariableSet contextVariables) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -419,6 +419,7 @@ private boolean run() {
                     t.fireCompletionHandlers(o); // do this after ErrorAction is set above
 
                     threads.remove(t.id);
+                    t.cleanUp();
```
bitwiseman, line 422: I think I like this better than #367 because it seems like there's fewer paths in which this call could be missed but I might not be understanding.

dwnusbaum:
@bitwiseman I added some assertions for `CpsThread.program` in fa8bd21 like you requested. As far as I can tell, that field previously could have been final and marked as non-null, so I'm not totally sure why the assertion in `CpsThread.run` existed. For existing code, `program` is only accessed when a live `CpsThread` is accessed through a `CpsThreadGroup`, so the assertions should never be triggered.

The other fields we clean up are all nullable, so it doesn't really make sense to add assertions around them.

PR #371, opened 2020-07-30
Add 2.82 release notes

dwnusbaum:
I am planning on releasing the plugin today. Only change is #368, so in case something goes wrong with that change we can just block the release from the update center without needing to worry about other changes that users might care about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### 2.82
+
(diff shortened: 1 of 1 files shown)


dwnusbaum:
The Windows branch died in the CI build due to infra issues. Merging anyway since this PR only touches `CHANGELOG.md` and the most recent build of the master branch is fine.

PR #375, opened 2020-08-25
[JENKINS-63500] Fix SnippetizerTest.oneOrMoreArgsStepDocs in Jenkins 2.236 and newer

dwnusbaum:
See [JENKINS-63500](https://issues.jenkins-ci.org/browse/JENKINS-63500).

Doc generation fails for the `input` step in new versions of Jenkins, so we switch this test over to a different step that doesn't have anything to do with parameters to avoid the issue.

https://github.com/jenkinsci/workflow-cps-plugin/commit/e2b1c95be13c17fc3dbf15cb55a55e16ffa217fc just moves some imports around, and https://github.com/jenkinsci/workflow-cps-plugin/commit/f9f4918b6b5b397c9a2fde53d6d3671a2e87ef62 is the fix.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
@@ -35,10 +35,16 @@ import org.jenkinsci.plugins.workflow.steps.CoreStep;
 import org.jenkinsci.plugins.workflow.steps.EchoStep;
 import org.jenkinsci.plugins.workflow.steps.PwdStep;
 import org.jenkinsci.plugins.workflow.steps.ReadFileStep;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -301,7 +303,7 @@ public void singleArgStepDocs() throws Exception {
 
     @Test
     public void oneOrMoreArgsStepDocs() throws Exception {
-        SnippetizerTester.assertDocGeneration(InputStep.class);
+        SnippetizerTester.assertDocGeneration(TimeoutStep.class);
```
dwnusbaum, line 306: `timeout` has one required parameter (`time`), and one optional parameter (`unit`), which is I think the scenario this test was trying to cover.

jglick: Perhaps. I do not really recall. Feel free to switch any of these to mock steps using `@TestExtension` if that is clearer or easier.

PR #377, opened 2020-09-17
If resume is blocked on a build, there is no purpose in ever saving program.dat

jglick:
Inspired by https://github.com/jenkinsci/jenkinsfile-runner/pull/364.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -177,13 +177,14 @@ public class FlowDurabilityTest {
         return run;
     }
 
-    static WorkflowRun createAndRunSleeperJob(Jenkins jenkins, String jobName, FlowDurabilityHint durabilityHint) throws Exception {
+    private WorkflowRun createAndRunSleeperJob(Jenkins jenkins, String jobName, FlowDurabilityHint durabilityHint, boolean resumeBlocked) throws Exception {
         Item prev = jenkins.getItemByFullName(jobName);
         if (prev != null) {
             prev.delete();
         }
 
         WorkflowJob job = jenkins.createProject(WorkflowJob.class, jobName);
+        job.setResumeBlocked(resumeBlocked);
         CpsFlowDefinition def = new CpsFlowDefinition("node {\n " +
                 "sleep 30 \n" +
                 "} \n" +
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -192,7 +193,7 @@ static WorkflowRun createAndRunSleeperJob(Jenkins jenkins, String jobName, FlowD
         provider.registerHint(jobName, durabilityHint);
         job.setDefinition(def);
         WorkflowRun run = job.scheduleBuild2(0).getStartCondition().get();
-        Thread.sleep(4000L);  // Hacky but we just need to ensure this can start up
+        story.j.waitForMessage("Sleeping for", run);
```
jglick, line 196: May make this less flaky. Noticed because it prevented me from using the debugger.

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -736,12 +737,13 @@ public void testResumeBlocked() throws Exception {
             @Override
             public void evaluate() throws Throwable {
                 Jenkins jenkins = story.j.jenkins;
-                WorkflowRun run = createAndRunSleeperJob(story.j.jenkins, jobName, FlowDurabilityHint.MAX_SURVIVABILITY);
-                run.getParent().setResumeBlocked(true);
```
jglick, line 740: This was happening too late—after the build had started. (That corner case is covered in the next test. This test covers the normal case that the job is permanently marked incapable of resuming.)

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -762,20 +764,16 @@ public void evaluate() throws Throwable {
     @Issue("JENKINS-49961")
     public void testResumeBlockedAddedAfterRunStart() throws Exception {
         final String jobName = "survivesEverything";
-        final String[] logStart = new String[1];
```
jglick, line 765: Some dead code I noticed.

dwnusbaum (approved):

oleg-nenashev (approved):
In Jenkinsfile Runner it might make sense to persist the final state of the program when exiting Pipeline, but indeed there is no need in intermediate stages. Good for performance if the Pipeline gets paused due to whatever reason

jglick:
> In Jenkinsfile Runner it might make sense to persist the final state of the program when exiting Pipeline

`program.dat` already gets deleted (if it were ever created!) when the build exits; its only purpose is to support resuming a build across a controller restart, which is an impossibility in JFR.

PR #370, opened 2020-07-28
[JENKINS-63254][JENKINS-47101] Insecure Groovy String Interpolation Warnings

car-roll:
[JENKINS-63254](https://issues.jenkins-ci.org/browse/JENKINS-63254)

PoC version 2 for groovy string interpolation interception.

This version does not require any modifications to core. Sensitive variables are registered in `BindingStep` via API from `EnvironmentExtender`. Scanning occurs in `DSL.flattenGString()`.

Requires:
https://github.com/jenkinsci/credentials-binding-plugin/pull/105
https://github.com/jenkinsci/workflow-step-api-plugin/pull/57
https://github.com/jenkins-infra/jenkins.io/pull/3814

This PR also subsumes https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/132

Strings that have been interpolated through `DSL.flattenGstring` are now tracked through `DSL.parseArgs()` and stored in `NamedArgsAndClosure`. The interpolated strings are then compared against the list of sensitive variables from `EnvironmentExpander` via the api established in https://github.com/jenkinsci/workflow-step-api-plugin/pull/57. Interpolated string warnings are displayed in the console output as well as the build page.

Special note: changes were made to `DSL.invokeDescribable`, specifically when there is no associated meta-step for a symbol (i.e. when building up a nested argument for a step). Previously, the method would return an `UninstantiatedDescribable` and it would be instantiated later in `DSL.invokeStep`. Now, the `UninstantiatedDescribable` is packed into `NamedArgsAndClosure` and later unpacked when `invokeStep` calls `parseArgs`.

The changes in `ArgumentsActionImpl` address [Jennkins-47101](https://issues.jenkins-ci.org/browse/JENKINS-47101). It uses the new `workflow-step` API to identify sensitive variables.  The `SAFE_ENVIRONMENT_VARIABLES` list has been removed and instead of checking for safe variables, we now check for sensitive variables. Sensitive variables  are no longer dropped with  `NotStoredReason.MASKED_VALUE`, but are now replaced with the variables name (e.g. "echo s3cret" becomes "echo ${PASSWORD}"). This allows us to display the signature of the steps that contain interpolation warnings.
Example screenshot here:
Screen Shot 2020-10-03 at 2 19 08 PM

PR #378, opened 2020-09-23
[JENKINS-47101] Implement ArgumentsActionImpl with EnvironmentExpander

car-roll:
Remove the safe list from `ArgumentsActionImpl` and replace with the enhanced `EnvironmentExpander` from #370 
This should remove the need for a safe list that can cause incorrect application of masked variables.


src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
```diff
@@ -156,18 +156,20 @@ public void testStringSafetyTest() throws Exception {
...
+
+        Assert.assertFalse("EnvVars that do not occur are safe", ArgumentsActionImpl.isStringSensitive("I have no passwords", new EnvVars(passwordBinding), sensitiveVariables));
+
+//        HashMap<String, String> safeBinding = new HashMap<>();
+//        safeBinding.put("harmless", "secret");
+//        HashSet<String> safeVars = new HashSet<>();
+//        safeVars.add("harmless");
+//        passwordBinding.put("harmless", "secret");
+//        Assert.assertTrue("Input containing whitelisted bound value is safe", ArgumentsActionImpl.isStringSafe(input, new EnvVars(safeBinding), safeVars));
+//        Assert.assertFalse("Input containing one safe and one unsafe bound value is unsafe", ArgumentsActionImpl.isStringSafe(input, new EnvVars(passwordBinding), safeVars));
```
jglick, line 172: If this is obsolete, just delete it, do not comment it out.

jglick (commented):
IIUC this subsumes #370 and just adds https://github.com/jenkinsci/workflow-cps-plugin/pull/378/commits/6e8b5eb99200853905b7eedb6c91ec7678600d42?

src/test/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImplTest.java
car-roll: yeah, i meant to, just forgot to do that in the clean up.

car-roll:
@jglick sorry for the confusion. I should have just made this a PR to #370 
#370 is still going through review, but I just wanted to take a first stab at doing this  to see how much work would need to be done. I'm still inclined to keep these two separate PRs.

jglick:
> I should have just made this a PR to #370

Perhaps. There are downsides to that too. Unfortunately GH does not provide first-class support for “stacked” PRs. For example, you can display the incremental diff as https://github.com/car-roll/workflow-cps-plugin/compare/interpolation-v2...remove-safe-list but then reviewers cannot add line comments. You can display the incremental diff in the **Files** tab here but you cannot create a permalink to that as the heads of one or both PRs move.

car-roll:
Closing this PR and merging with #370

PR #382, opened 2020-10-09
Add Dependabot configuration

dwnusbaum:
Also updates the BOM.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  # Intentionally not adding automatic NPM updates because I do
+  # not trust the plugin's tests to catch issues caused by updates.
+  - package-ecosystem: "maven"
+    directory: "/"
+    reviewers:
+      - "dwnusbaum"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  # Intentionally not adding automatic NPM updates because I do
+  # not trust the plugin's tests to catch issues caused by updates.
+  - package-ecosystem: "maven"
+    directory: "/"
+    reviewers:
+      - "dwnusbaum"
+    schedule:
+      interval: "daily"
```
olamy, line 10: maybe only weekly?

dwnusbaum: I think that most of the dependencies here have their version controlled via https://github.com/jenkinsci/bom, so we shouldn't see too many updates. Maybe we can try daily for now and move to weekly if there is too much noise?

olamy: yup go for it. I just sometimes receive too many dependabot email. I wonder if gmail will flag this as a spam :)

olamy (approved):

PR #397, opened 2020-11-12
[JENKINS-64203] Save original build Cause(s) on pipeline Replay

ababushk:
This PR adds code to copy build Cause(s) from replayed build to a new one. This is useful when pipeline is triggered with plugins like GitLab plugin which store additional data in build Cause and make it available in pipeline via environment variables


ababushk:
I'll merge latest changes and re-open the PR

PR #398, opened 2020-11-12
[JENKINS-64203] Copy original build Cause(s) on Replay

ababushk:
This PR adds code to copy build Cause(s) from replayed build to a new one. This is useful when pipeline is triggered with plugins like GitLab plugin which store additional data in build Cause and make it available in pipeline via environment variables


ababushk:
Greetings @dwnusbaum, could you please take a look?

PR #399, opened 2020-11-17
[JENKINS-64185] Replace use of NamedArgsAndClosure as return value for invokeDescribable

car-roll:
There are some conditions in which the NamedArgsAndClosure class is passed to other packages where it is not recognized. This fix here goes back to an earlier implementation in #370  where we create a subclass of `UninstantiatedDescribable` that stores interpolated strings. This should ensure that existing code that used to look for `UninstantiatedDescribable` can continue to work as expected.

Note: not sure what tests to add to reflect this problem. Did manual testing with the Gerrit plugin

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -435,7 +435,6 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         // The only time a closure is valid is when the resulting Describable is immediately executed via a meta-step
         NamedArgsAndClosure args = parseArgs(_args, metaStep!=null && metaStep.takesImplicitBlockArgument(),
                 UninstantiatedDescribable.ANONYMOUS_KEY, singleArgumentOnly, new HashSet<>());
-        UninstantiatedDescribable ud = new UninstantiatedDescribable(symbol, null, args.namedArgs);
 
         if (metaStep==null) {
             // there's no meta-step associated with it, so this symbol is not executable.
@@ -448,11 +447,9 @@ public class DSL extends GroovyObjectSupport implements Serializable {
             // where this UninstantiatedDescribable is ultimately used, the symbol
             // might be resolved with a specific type.
 
-            // we are returning the NamedArgsAndClosure instead of the UninstantiatedDescribable in order to preserve
-            // the discovered interpolated strings that are stored in the NamedArgsAndClosure.
-            args.uninstantiatedDescribable = ud;
-            return args;
+            return new UninstantiatedDescribableWithInterpolation(symbol, null, args.namedArgs, args.interpolatedStrings);
         } else {
+            UninstantiatedDescribable ud = new UninstantiatedDescribable(symbol, null, args.namedArgs);
             Descriptor d = SymbolLookup.get().findDescriptor((Class)(metaStep.getMetaStepArgumentType()), symbol);
             try {
                 // execute this Describable through a meta-step
@@ -544,8 +541,6 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         final Closure body;
         final List<String> msgs;
         final Set<String> interpolatedStrings;
-        // UninstantiatedDescribable is set when the associated symbol is being built as the parameter of a step to be invoked.
-        UninstantiatedDescribable uninstantiatedDescribable = null;
 
         private NamedArgsAndClosure(Map<?,?> namedArgs, Closure body, @Nonnull Set<String> foundInterpolatedStrings) {
             this.namedArgs = new LinkedHashMap<>(preallocatedHashmapCapacity(namedArgs.size()));
@@ -567,9 +562,9 @@ public class DSL extends GroovyObjectSupport implements Serializable {
          * Gathers all the interpolated strings from each instance of {@link NamedArgsAndClosure}.
          */
         private static Object collectInterpolatedStrings(Object argValue, Set<String> interpolatedStrings) {
-            if (argValue instanceof NamedArgsAndClosure) {
-                interpolatedStrings.addAll(((NamedArgsAndClosure) argValue).interpolatedStrings);
-                return ((NamedArgsAndClosure) argValue).uninstantiatedDescribable;
+            if (argValue instanceof UninstantiatedDescribableWithInterpolation) {
+                interpolatedStrings.addAll(((UninstantiatedDescribableWithInterpolation) argValue).getInterpolatedStrings());
+                return argValue;
             } else if (argValue instanceof Map) {
                 Map<Object, Object> r = new LinkedHashMap<>(preallocatedHashmapCapacity(((Map) argValue).size()));
                 for (Map.Entry e : ((Map<Object, Object>) argValue).entrySet()) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -289,7 +289,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         ClassLoader originalLoader = Thread.currentThread().getContextClassLoader();
         try {
             TaskListener listener = context.get(TaskListener.class);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/UninstantiatedDescribableWithInterpolation.java
```diff
@@ -0,0 +1,29 @@
+package org.jenkinsci.plugins.workflow.cps;
+
+import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
+
+import java.util.Map;
+import java.util.Set;
+
+public class UninstantiatedDescribableWithInterpolation extends UninstantiatedDescribable {
+    private final Set<String> interpolatedStrings;
```
dwnusbaum, line 9:
```suggestion
    private static final long serialVersionUID = 1L;
    private final Set<String> interpolatedStrings;
```

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -448,11 +447,9 @@ protected Object invokeDescribable(String symbol, Object _args) {
             // where this UninstantiatedDescribable is ultimately used, the symbol
             // might be resolved with a specific type.
 
-            // we are returning the NamedArgsAndClosure instead of the UninstantiatedDescribable in order to preserve
-            // the discovered interpolated strings that are stored in the NamedArgsAndClosure.
-            args.uninstantiatedDescribable = ud;
-            return args;
+            return new UninstantiatedDescribableWithInterpolation(symbol, null, args.namedArgs, args.interpolatedStrings);
         } else {
+            UninstantiatedDescribable ud = new UninstantiatedDescribable(symbol, null, args.namedArgs);
```
dwnusbaum, line 452: Are you intentionally not adding the interpolated strings here? Does it not matter because we will scan everything in `invokeStep` in this case anyway?

src/main/java/org/jenkinsci/plugins/workflow/cps/UninstantiatedDescribableWithInterpolation.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
```
dwnusbaum, line 28: You could omit these right? Or did you just want to make it clear that `interpolatedStrings` was not part of the equality check? Should `interpolatedStrings` be part of the equality check? I'm not sure if it matters.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -448,11 +447,9 @@ protected Object invokeDescribable(String symbol, Object _args) {
             // where this UninstantiatedDescribable is ultimately used, the symbol
             // might be resolved with a specific type.
 
-            // we are returning the NamedArgsAndClosure instead of the UninstantiatedDescribable in order to preserve
-            // the discovered interpolated strings that are stored in the NamedArgsAndClosure.
-            args.uninstantiatedDescribable = ud;
-            return args;
+            return new UninstantiatedDescribableWithInterpolation(symbol, null, args.namedArgs, args.interpolatedStrings);
```
dwnusbaum, line 450:
```suggestion
            // Note: Declarative relies on an instance of UninstantiatedDescribable being returned here in some cases where there is no Pipeline step involved, e.g. for the `triggers` directive.
            return new UninstantiatedDescribableWithInterpolation(symbol, null, args.namedArgs, args.interpolatedStrings);
```

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
car-roll: Yes, in this case the interpolated strings were scanned already in the `NamedArgsAndClosure` here: https://github.com/jenkinsci/workflow-cps-plugin/blob/7f00f8c8d40f5a3e75abc1a4a1ca14f13ab929dd/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java#L436
and then used in `invokeStep` as a new `NamedArgsAndClosure`. I guess I could have left it outside of the `if-else` statement, I just reverted it to match a bit closer to the old code.

src/main/java/org/jenkinsci/plugins/workflow/cps/UninstantiatedDescribableWithInterpolation.java
car-roll: In this case, it was spotbugs complaining to me that I had no overriden `equals` and `hashcode`

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
dwnusbaum: Ok, the way you have it seems fine to me, just wanted to make sure it was intentional.

src/main/java/org/jenkinsci/plugins/workflow/cps/UninstantiatedDescribableWithInterpolation.java
car-roll: added in b828472cff3abc7ccce59127b6a8a1e5845a2eb3

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
car-roll: added in b828472cff3abc7ccce59127b6a8a1e5845a2eb3

dwnusbaum (approved):

Costallat:
Tested here and it worked again, tested manually triggering the job and also using Gerrit.

Costallat (approved):

Costallat (approved):
Tested here, and it worked

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -544,8 +542,6 @@ private static int preallocatedHashmapCapacity(int elementsToHold) {
         final Closure body;
         final List<String> msgs;
         final Set<String> interpolatedStrings;
-        // UninstantiatedDescribable is set when the associated symbol is being built as the parameter of a step to be invoked.
-        UninstantiatedDescribable uninstantiatedDescribable = null;
 
         private NamedArgsAndClosure(Map<?,?> namedArgs, Closure body, @Nonnull Set<String> foundInterpolatedStrings) {
```
dwnusbaum, line 546: @car-roll Do we need to do any cleanup in `NamedArgsAndClosure` now? I guess we should remove `implements Serializable` and remove "The instance is returned to the Groovy program ..." from the Javadoc.

car-roll: Technically the whole paragraph might have to go, right? Because we are now using the `UninstantiatedDescribable` again. Maybe that paragraph should go to `UninstantiatedDescribableWithInterpolation`?

dwnusbaum: Yeah you are right, everything other than "This class holds the argument map and optional body of the step that is to be invoked." is no longer accurate. Feel free to move it to the new class if you want.

src/main/java/org/jenkinsci/plugins/workflow/cps/UninstantiatedDescribableWithInterpolation.java
```diff
@@ -0,0 +1,30 @@
+package org.jenkinsci.plugins.workflow.cps;
+
+import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
+
+import java.util.Map;
+import java.util.Set;
+
+public class UninstantiatedDescribableWithInterpolation extends UninstantiatedDescribable {
```
dwnusbaum, line 8: We should restrict this class (some imports will need to be added along with this suggestion):

```suggestion
@Restricted(NoExternalUse.class)
public class UninstantiatedDescribableWithInterpolation extends UninstantiatedDescribable {
```

dwnusbaum (approved):

PR #401, opened 2020-11-25
[JENKINS-64282] Skip if sensitive variable has an empty string

car-roll:
See [JENKINS-64282](https://issues.jenkins.io/browse/JENKINS-64282).

Empty string password parameters will trigger a false warning of groovy string interpolation

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -376,7 +376,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         }
 
         List<String> scanResults = sensitiveVariables.stream()
-                .filter(e -> interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
+                .filter(e -> !(envVars.get(e).isEmpty()) && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
                 .collect(Collectors.toList());
 
         if (scanResults != null && !scanResults.isEmpty()) {
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
@@ -31,6 +31,10 @@ import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
 import hudson.Functions;
 import hudson.model.AbstractDescribableImpl;
 import hudson.model.Descriptor;
+import hudson.model.ParametersAction;
+import hudson.model.ParametersDefinitionProperty;
+import hudson.model.PasswordParameterDefinition;
+import hudson.model.PasswordParameterValue;
 import hudson.model.Result;
 
 import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


dwnusbaum (changes requested):
@car-roll If this is specific to password parameters, it might be clearer to modify the handler for password parameters in `workflow-support` to not include password parameters where the value is null or empty instead: https://github.com/jenkinsci/workflow-support-plugin/blob/f92c6f41f911d6e82eefea984a98ed6d13186ed4/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java#L196-L199

I guess this could affect credentials as well though, so maybe it's good to patch `workflow-cps` generically, but I can't think of a case where an empty secret in a credential would be anything other than user error.

If you want to patch things here, I think you also need to change `ArgumentsActionImpl` to ignore the empty string as well, because from a quick local test it looks like `"abc".replace("", "${foo}")` is `{foo}a{foo}b{foo}c{foo}`:
https://github.com/jenkinsci/workflow-cps-plugin/blob/9b4aa81a5d938572435e850dd971995de3cd4ac0/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java#L97-L99

car-roll:
@dwnusbaum thanks for the catch. I agree the changes make more sense in `workflow-support`. I'll make a PR there and keep this PR just to harden `workflow-cps`.

src/test/java/org/jenkinsci/plugins/workflow/cps/DSLTest.java
```diff
@@ -604,6 +609,56 @@ public void namedSoleParamForStep() throws Exception {
         WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
     }
 
+    @Test public void passwordParameter() throws Exception {
```
dwnusbaum, line 612: I added `ArgumentsActionImpl.passwordParametersSanitized` in https://github.com/jenkinsci/workflow-cps-plugin/pull/370/commits/78111484e999fe68e87b7345d45ff68121c0d396 that is basically a subset of this test, so I would probably delete my old test so we don't test the same thing twice and since this test adds some assertions for the warnings that the old test didn't have so it is better to keep this one.

dwnusbaum (approved):

garazdawi:
Could this PR be the source of https://issues.jenkins.io/browse/JENKINS-64729? I've encountered the same problem as the reporter.

src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -376,7 +376,7 @@ private void logInterpolationWarnings(String stepName, @CheckForNull ArgumentsAc
         }
 
         List<String> scanResults = sensitiveVariables.stream()
-                .filter(e -> interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
+                .filter(e -> !(envVars.get(e).isEmpty()) && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
```
turboBasic, line 379: shouldn't it be `.filter(e -> !(envVars.get(e,'').isEmpty()) && ...` ?

PR #411, opened 2021-01-13
Handle tables-to-divs conversion

bitwiseman:
This is a simplistic start at the tables to divs conversion for this plugin based on https://github.com/jenkinsci/scm-api-plugin/commit/0c2723bb5029805565ffe128b98bfca6574d418b . 

This may be all that is needed, but we should run tests against some downstream plugins just to be sure.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -42,35 +42,6 @@ THE SOFTWARE.
                 <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
                 <j:set var="item" value="${it.getItem(request)}"/>
                 <d:taglib uri="local">
-                    <!-- this tag wraps the dropdown list in a div but only in div-based forms
-                        using a div in table-based forms will crash the layout generation -->
-                    <d:tag name="snippetizerOptionsWrapper">
-                       <j:choose>
-                           <j:when test="${divBasedFormLayout}">
-                               <style>
-                                   .snippetizer-options-wrapper .setting-input.dropdownList {
-                                       max-width: calc(100% - 2rem);
-                                   }
-
-                                   .snippetizer-help {
-                                       position: relative;
-                                   }
-
-                                   .snippetizer-help a.help-button {
-                                       position: absolute;
-                                       top: -2rem;
-                                   }
-                               </style>
-
-                               <div class="snippetizer-options-wrapper">
-                                   <d:invokeBody />
-                               </div>
-                           </j:when>
-                           <j:otherwise>
-                               <d:invokeBody />
-                           </j:otherwise>
-                       </j:choose>
-                    </d:tag>
 
                     <d:tag name="listSteps">
                         <j:forEach var="quasiDescriptor" items="${quasiDescriptors}">
@@ -80,23 +51,7 @@ THE SOFTWARE.
                                     <j:set var="it" value="${item}"/>
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
-                                    
-                                    <j:if test="${help != null}">
-                                        <j:choose>
-                                            <j:when test="${divBasedFormLayout}">
-                                                <div class="snippetizer-help">
-                                                    <f:entry help="${help}" />
-                                                </div>
-                                            </j:when>
-                                            <j:otherwise>
-                                                <tr>
-                                                    <td colspan="3"/>
-                                                    <f:helpLink url="${help}"/>
-                                                </tr>
-                                            </j:otherwise>
-                                        </j:choose>
-                                        <f:helpArea/>
-                                    </j:if>
+                                    <f:entry title="STEP: &lt;code&gt;${quasiDescriptor.symbol}&lt;code&gt;" help="${help}" />
                                     <st:include from="${descriptor}" page="${descriptor.configPage}" optional="true">
                                         <f:block>
                                             ${%This step has not yet defined any visual configuration.}
@@ -108,13 +63,11 @@ THE SOFTWARE.
                     </d:tag>
                 </d:taglib>
 
-                <local:snippetizerOptionsWrapper>
-                    <f:dropdownList name="prototype" title="${%Sample Step}">
-                        <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(false)}"/>
-                        <f:dropdownListBlock title="${%— Advanced/Deprecated —}"/>
-                        <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(true)}"/>
-                    </f:dropdownList>
-                </local:snippetizerOptionsWrapper>
+                <f:dropdownList name="prototype" title="${%Sample Step}">
+                    <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(false)}"/>
+                    <f:dropdownListBlock title="${%— Advanced/Deprecated —}"/>
+                    <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(true)}"/>
+                </f:dropdownList>
 
                 <j:set var="id" value="${h.generateId()}"/>
                 <f:block>
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -51,7 +51,7 @@ THE SOFTWARE.
                                     <j:set var="it" value="${item}"/>
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
-                                    <f:entry title="&lt;strong&gt;&lt;code&gt;${symbol}&lt;code&gt;&lt;/strong&gt;" help="${help}" />
+                                    <f:entry title="STEP: ${symbol}" help="${help}" />
                                     <st:include from="${descriptor}" page="${descriptor.configPage}" optional="true">
                                         <f:block>
                                             ${%This step has not yet defined any visual configuration.}
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -51,7 +51,7 @@ THE SOFTWARE.
                                     <j:set var="it" value="${item}"/>
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
-                                    <f:entry title="STEP: ${symbol}" help="${help}" />
+                                    <f:entry title="${symbol}" help="${help}" />
                                     <st:include from="${descriptor}" page="${descriptor.configPage}" optional="true">
                                         <f:block>
                                             ${%This step has not yet defined any visual configuration.}
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -42,16 +42,16 @@ THE SOFTWARE.
                 <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
                 <j:set var="item" value="${it.getItem(request)}"/>
                 <d:taglib uri="local">
-
                     <d:tag name="listSteps">
                         <j:forEach var="quasiDescriptor" items="${quasiDescriptors}">
                             <j:set var="descriptor" value="${quasiDescriptor.real}"/>
-                            <f:dropdownListBlock title="${quasiDescriptor.symbol}: ${descriptor.displayName}" staplerClass="${descriptor.clazz.name}" lazy="descriptor,item">
+                            <j:set var="symbol" value="${quasiDescriptor.symbol}"/>
+                            <f:dropdownListBlock title="${symbol}: ${descriptor.displayName}" staplerClass="${descriptor.clazz.name}" lazy="descriptor,item,symbol">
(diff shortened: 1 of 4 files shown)


timja (approved):

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -50,11 +78,12 @@ THE SOFTWARE.
                                     <j:set var="it" value="${item}"/>
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
+                                    
                                     <j:if test="${help != null}">
-                                        <tr>
-                                            <td colspan="3"/>
+                                        <local:row>
+                                            <local:td />
```
timja, line 84: is this actually needed on a tables to divs core? what does the UI render like on new versions?

this is just adding an empty div

dwnusbaum: From a quick test with 2.276, with and without this change, after tables-to-divs the help icon for the step itself and the first parameter of the step itself end up side by side. Before tables-to-divs, the help icon for the step itself was one line up (it has always looked kind of weird; there might be a Jira ticket about this). If you expand the help text, each icon gets its own line as before and things look normal.

Screen Shot 2021-01-27 at 16 59 45

If I change the whole section to the following (the local taglib would have to be tweaked to be able to add the `br` tag only after tables-to-divs), then the post-table-to-divs appearance looks pretty close to the previous appearance:

```
<div class="tr">
    <f:helpLink url="${help}"/>
</div>
<f:helpArea/>
<br/>
```

There might be a better way to accomplish the same result. Here's a screenshot with that change:

Screen Shot 2021-01-27 at 17 10 32

timja (changes requested):
so this fix seems to break the snippet generator, 

without this change I can't see anything visually wrong (other than the side by side help buttons which isn't good)

but with this fix I can only expand the help for archiveArtifacts

If I try expand say 'fingerprints' help, nothing happens, interestingly if I switch back to 'archiveArtifacts' the help is expanded for 'fingerprints'

I assume some javascript is going wrong

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
timja: I was able to fix it with:

```
                                    <j:if test="${help != null}">
                                        <div class="setting-name help-sibling snippetizer-help">
                                            <f:helpLink url="${help}"/>
                                        </div>
                                        <f:helpArea/>
                                    </j:if>
```

image

image

Don't have time to push a fix right now though

bitwiseman:
@timja 
So, I should close this PR and leave the plugin as is?

timja:
It also needs a bit of css (height: 20px) from my earlier comment did it) to fix the help icons that are side by side unless you fixed that another way

bitwiseman:
@timja  Thanks, I think i got it now.  It picks up the correct help items on Jenkins v2.276.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/workflow.css
```diff
@@ -99,6 +99,10 @@ z.root>dt {
   padding-bottom: 0
 }
 
+.snippetizer-help {
```
timja, line 102: this css file appears not to be loaded on http://localhost:8083/jenkins/job/pipeline/pipeline-syntax/

the css wasn't applied, and when I searched the dom I couldn't find the css file
image

bitwiseman: @timja Okay, so I still don't have that right.

bitwiseman: @timja 
I agree it doesn't look great, but I feel this should not block merging the PR.  Would you accept that or do we need to address it in this PR?  Either way, I'm going to bed for the night, back online in about six hours.

MRamonLeon:
Maybe @fqueiruga could help

fqueiruga:
This is the best I can do, with VERY hacky code:

Captura de pantalla 2021-02-15 a las 12 01 00
Captura de pantalla 2021-02-15 a las 12 01 04

fqueiruga:
Pasting the interesting bits:

```
<d:taglib uri="local">
  <d:tag name="listSteps">
    <j:forEach var="quasiDescriptor" items="${quasiDescriptors}">
      <j:set var="descriptor" value="${quasiDescriptor.real}" />
      <f:dropdownListBlock
        title="${quasiDescriptor.symbol}: ${descriptor.displayName}"
        staplerClass="${descriptor.clazz.name}"
        lazy="descriptor,item"
      >
        <l:ajax>
          <j:set var="it" value="${item}" />
          <j:set var="instance" value="${null}" />
          <j:set var="help" value="${descriptor.helpFile}" />

          <j:if test="${help != null}">
            <j:choose>
              <j:when test="${divBasedFormLayout}">
                <div class="snippetizer-help">
                  <f:entry help="${help}" />
                </div>
              </j:when>
              <j:otherwise>
                <tr>
                  <td colspan="3" />
                  <f:helpLink url="${help}" />
                </tr>
              </j:otherwise>
            </j:choose>
            <f:helpArea />
          </j:if>
          <st:include
            from="${descriptor}"
            page="${descriptor.configPage}"
            optional="true"
          >
            <f:block>
              ${%This step has not yet defined any visual configuration.}
            </f:block>
          </st:include>
        </l:ajax>
      </f:dropdownListBlock>
    </j:forEach>
  </d:tag>
</d:taglib>

<style>
  .snippetizer-options-wrapper .setting-input.dropdownList {
    max-width: calc(100% - 2rem);
  }

  .snippetizer-help {
    position: relative;
  }

  .snippetizer-help a.help-button {
    position: absolute;
    top: -2rem;
  }
</style>

<div class="snippetizer-options-wrapper">
  <f:dropdownList name="prototype" title="${%Sample Step}">
    <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(false)}" />
    <f:dropdownListBlock title="${%— Advanced/Deprecated —}" />
    <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(true)}" />
  </f:dropdownList>
</div>

```

jtnord (approved):
I guess this is an improvement as it makes things less broken, but I would have thought that https://github.com/jenkinsci/workflow-cps-plugin/pull/411/files#r576049627 needs to be fixed.

timja:
@jtnord, felix is working on it atm, chatting via gitter

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/workflow.css
fqueiruga: I have a proposal on https://github.com/jenkinsci/workflow-cps-plugin/pull/411#issuecomment-779145393 . I used inline styling because I didn't want to get into why the css wasn't loaded. if the file is not being used we might as well remove it.

fqueiruga (changes requested):

fqueiruga:
@bitwiseman I have created a PR on your fork for this branch to fix the help icon placement.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/workflow.css
basil: Based on some examples in Jenkins core, I think to include the CSS you'd need to add:

```xml
<st:adjunct includes="org.jenkinsci.plugins.workflow.cps.Snippetizer.workflow"/>
```

bitwiseman:
@fqueiruga 

> I have a proposal on #411 (comment) . I used inline styling because I didn't want to get into why the css wasn't loaded. if the file is not being used we might as well remove it.

It used in `src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/html.groovy`.  Not sure where that is used yet, but seems like it isn't extraneous.

bitwiseman:
@fqueiruga 
Wow, that PR is really cool.  Is there any way to do this without resorting to hackery?

bitwiseman:
I've merged the changes from @fqueiruga because they are definably better than what I have.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -42,6 +42,36 @@ THE SOFTWARE.
...
+
+                                   .snippetizer-help a.help-button {
+                                       position: absolute;
+                                       top: -2rem;
+                                   }
+                               </style>
+
+                               <div class="snippetizer-options-wrapper">
+                                   <d:invokeBody />
+                               </div>
```
bitwiseman, line 67: @fqueiruga  
Could we include this div regardless of whether we're using div-based layout?  Or would that explode?

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/workflow.css
```diff
@@ -99,6 +99,10 @@ z.root>dt {
...
+  height: 20px;
+}
+
```
timja, line 105: this isn't being uses
```suggestion
```

timja (approved):

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
fqueiruga: it will explode, that's why i created this tag that noops on table layouts.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -50,13 +51,7 @@ THE SOFTWARE.
...
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
-                                    <j:if test="${help != null}">
-                                        <tr>
-                                            <td colspan="3"/>
-                                            <f:helpLink url="${help}"/>
-                                        </tr>
-                                        <f:helpArea/>
-                                    </j:if>
+                                    <f:entry title="STEP: &lt;code&gt;${quasiDescriptor.symbol}&lt;code&gt;" help="${help}" />
```
bitwiseman, line 54: Instead of doing some magic to position the help for a step, we could add an entry and use it to anchor the help for the step.  
I want this to look like "STEP: `symbol`" but it seems quasiDescriptor.symbol} is not available when the block is expanded. 
@timja  @fqueiruga  do you know the magic to get the value down in the block?

bitwiseman: I suppose I could use this instead, keep things simple. 
```suggestion
                                    <f:entry title="Step parameters" help="${help}" />
```

bitwiseman:
Here's the new look in div-based and table-based layouts:

Screen Shot 2021-02-16 at 9 29 49 AM

Screen Shot 2021-02-16 at 9 24 06 AM

fqueiruga (approved):

fqueiruga:
Fine by me as long as it works but needs some comments.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -45,18 +45,13 @@ THE SOFTWARE.
...
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
-                                    <j:if test="${help != null}">
-                                        <tr>
-                                            <td colspan="3"/>
-                                            <f:helpLink url="${help}"/>
-                                        </tr>
-                                        <f:helpArea/>
-                                    </j:if>
+                                    <f:entry title="&lt;strong&gt;&lt;code&gt;${symbol}&lt;code&gt;&lt;/strong&gt;" help="${help}" />
```
dwnusbaum, line 54: I would change this to just be `title="${symbol}"`, or use `Step: ` as a prefix or something if that looks weird. Right now the tooltip for the help icon is literally "Help for feature \<strong>\<code>step name here\</code>\</strong>" in recent versions of Jenkins due to the fix for [SECURITY-1955](https://www.jenkins.io/security/advisory/2020-08-12/#SECURITY-1955).

dwnusbaum (approved):
When testing against pre-tables-to-divs versions of Jenkins, I noticed that whitespace accumulates between the dropdown and the first element displayed for the step when you select later steps in the dropdown, but then I checked `master` and it has the same behavior 🤷‍♂️. For example with this PR:

Screen Shot 2021-02-16 at 13 00 08

And with `master`:

Screen Shot 2021-02-16 at 13 12 34

When testing against post-tables-to-divs versions of Jenkins I did not see this issue, so I think we can ignore it.

timja (approved):
lgtm, help feature title is looking a bit weird as Devin said

dwnusbaum:
Also just to note, I did some quick testing to confirm the behavior and the main functionality of the snippet generator works fine before and after tables-to-divs with or without this PR. The only issue after tables-to-divs without this PR is that the layout of the help icons is messed up as shown in the first screenshot in https://github.com/jenkinsci/workflow-cps-plugin/pull/411#discussion_r565667800.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -51,7 +51,7 @@ THE SOFTWARE.
                                     <j:set var="it" value="${item}"/>
                                     <j:set var="instance" value="${null}"/>
                                     <j:set var="help" value="${descriptor.helpFile}"/>
-                                    <f:entry title="&lt;strong&gt;&lt;code&gt;${symbol}&lt;code&gt;&lt;/strong&gt;" help="${help}" />
+                                    <f:entry title="STEP: ${symbol}" help="${help}" />
```
timja, line 54: Is the prefix even needed? / why is it capitalised? 

image

```suggestion
                                    <f:entry title="${symbol}" help="${help}" />
```

timja (approved):

PR #412, opened 2021-01-28
Fix build on mac m1

timja:
cc @bitwiseman @dwnusbaum

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.14</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -72,7 +72,6 @@
         <workflow-support-plugin.version>3.7</workflow-support-plugin.version>
         <node.version>12.19.0</node.version>
         <npm.version>6.14.8</npm.version>
-        <frontend-version>1.10.0</frontend-version>
```
timja, line 75: this is overriding the parent and building breaks on my machine, node and npm could probably do with an update too but just wanting to compile for now

dwnusbaum (approved):
Thanks!

PR #414, opened 2021-02-10
Add new step LoadString

bmiguel-teixeira:
We have recently came up against this particular issue that this PR should patch with minimal noise. This is a non-breaking change.

Consider the following scenario.
The `load` instruction loads a `dummy.groovy`. However, this `dummy.groovy` also pulls a new node. And while this `dummy.groovy` is executing in Node Y, the original Node X that launched the `dummy.groovy` still has a tied down executor doing nothing (waiting for execution of `dummy.groovy`).

The `load` function is specifically tied to a path forcing the need of a node for proper execution. This PR introduces a new step `loadString` which offers the same functionality  as `load` by passing the file content instead if the path.

This new method can then be called outside of a node context, and allow regular execution.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyInvoker.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyInvoker.java
@@ -40,6 +40,7 @@ import javax.annotation.Nonnull;
 
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.*;
 import org.jenkinsci.plugins.workflow.cps.steps.LoadStep;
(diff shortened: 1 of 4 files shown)


dwnusbaum (changes requested):
@bmiguel-teixeira Thanks for the PR! I think there is already a way to do this today using `evaluate`, for example: 

```
def script = 'echo("Hello, world!")'
evaluate(script)
```

Does `evaluate` not work for your use case? I would really prefer not to add another step similar to `load`.

brunomiguel-teixeira:
Well...I was unaware of that feature in groovy. We have tested the `evaluate` locally and it works like a charm. 😅 

Thanks a lot. I'll be closing this PR since `evaluate` does pratically the same thing.

PR #188, opened 2017-11-13
Add new durability modes for improving performance + other goodies

svanoort:
Replaces #170 

# Functional tasks
- [x] Pass and use DurabilityHints to set up storage
- [x] Use durability hint to decide when we should persist program and what storage to use for node (plus when to persist them).
- [x] Correctly defer writing FlowNode actions, with test cases
- [x] Provide testcases for the key durability modes covering: resume build, complete build, try to resume build when we can't do so due to system failure.
- [x] Tests verify no dangling tasks left
- [x] Test that hard failure of the master (not clean shutdown) doesn't hose things upon restart (clean failures)
- [x] CPS Program persists before restart if we're set up to survive clean restarts but not continuously persistent
- [x] Tests pass for dirty restarts
- [x] Optional: try to cleanly terminate flow graphs with incomplete persistence state (best-effort basis)
- [x] Apply any refactors to the DurabilityHint API

# Administrivia
- [x] Re-merge master
- [x] Deploy the upstream SNAPSHOTs

PR #415, opened 2021-02-15
[JENKINS-63305] Fix bug that ignores durability settings with lightweight checkout

car-roll:
[JENKINS-63305](https://issues.jenkins.io/browse/JENKINS-63305)

Pipeline job durability settings are ignored when lightweight checkout is used. This is due to an incorrect instance check of the `Queue.Executable`.


dwnusbaum (approved):

dwnusbaum:
Also just noting that I think this has been broken since the code was introduced in https://github.com/jenkinsci/workflow-cps-plugin/pull/188.

car-roll:
bump for ci issues

PR #419, opened 2021-02-19
[JENKINS-64729] use EnvVars get() instead of TreeMap get()

car-roll:
Fix bug where `TreeMap.get()` was  used  instead of `EnvVars.get()`.
The difference being that `EnvVars.get()` provides a default value if the key  does not exists as opposed to just `null` with `TreeMap.get()`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -376,7 +376,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         }
 
         List<String> scanResults = sensitiveVariables.stream()
-                .filter(e -> !(envVars.get(e, "").isEmpty()) && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
+                .filter(e -> !envVars.get(e, "").isEmpty() && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
                 .collect(Collectors.toList());
 
         if (scanResults != null && !scanResults.isEmpty()) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -376,7 +376,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         }
 
         List<String> scanResults = sensitiveVariables.stream()
-                .filter(e -> !(envVars.get(e).isEmpty()) && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -376,7 +376,7 @@ private void logInterpolationWarnings(String stepName, @CheckForNull ArgumentsAc
         }
 
         List<String> scanResults = sensitiveVariables.stream()
-                .filter(e -> !(envVars.get(e).isEmpty()) && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
+                .filter(e -> !(envVars.get(e, "").isEmpty()) && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
```
dwnusbaum, line 379: nit: looks like there were some extra parentheses that could be removed:

```suggestion
                .filter(e -> !envVars.get(e, "").isEmpty() && interpolatedStrings.stream().anyMatch(g -> g.contains(envVars.get(e))))
```

dwnusbaum (commented):
Looks good, but do we need to change this code as well? https://github.com/jenkinsci/workflow-cps-plugin/blob/92457d6fe5a97860d7f341df8ba61597a7eb0ce3/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java#L98-L99

dwnusbaum (approved):
Looks like this covers everything from https://github.com/jenkinsci/workflow-cps-plugin/pull/401, but it would be good to check if there is any other use of `EnvVars` related to secret masking in the plugin.

car-roll:
@dwnusbaum I did an inspection of other `EnvVars` usage here and I found another  usage  of `TreeMap.get()` being used for EnvVars here at  the end of the same  line we  modified here:
` .filter(e -> !envVars.get(e, "").isEmpty() && interpolatedStrings.stream().anyMatch(g -> **g.contains(envVars.get(e))**))`
But in this case  I  think we want the `get()`  to return null since if it returned  any empty string, you  would  always get a match.
Another place was here: https://github.com/jenkinsci/workflow-cps-plugin/blob/92457d6fe5a97860d7f341df8ba61597a7eb0ce3/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java#L102, but in this case I personally think  there should be  an error since I'm not inclined to let slide  a sensitive  variable that  has no value.

dwnusbaum:
@car-roll I think both of those places are fine. The checks added in this PR mean we'll never execute the code in either location with a null environment variable because `.get(blah, "").isEmpty()` will be true so we will short-circuit the `&&` in the first case and execute `continue` in the block right above the second case.

turboBasic:
@car-roll thanks for the fix. Do you have any rough estimation when the next tagged release is expected?

caolandix:
This is bug is affecting my org. Has this been released?

PR #422, opened 2021-03-03
Don't leave zombie work units on exception in onStart

mrginglymus:


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -250,4 +254,78 @@ public boolean start() throws Exception {
...
+
+        @Override
+        public StepExecution start(StepContext context) throws Exception {
+            return new WithStartFailureStepExecution(context);
+        }
+
+        static class WithStartFailureStepExecution extends AbstractStepExecutionImpl {
+
+            WithStartFailureStepExecution(final StepContext context) {
+                super(context);
```
kieraCurtis, line 304: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

dwnusbaum: This is a constructor, not a regular method.

kieraCurtis: noticed.

jglick:
This could be mergeable if you just added

```java
@Ignore("TODO JENKINS-xxxxx or some other explanation")
```

PR #424, opened 2021-03-05
[JENKINS-50855] Clearly differentiate between Replay and Rebuild

jglick:
[JENKINS-50855](https://issues.jenkins.io/browse/JENKINS-50855) has been a major source of confusion since #187.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
@@ -96,7 +96,7 @@ public class ReplayAction implements Action {
     }
 
     @Override public String getDisplayName() {
-        return Messages.ReplayAction_displayName();
+        return isEnabled() ? Messages.ReplayAction_displayName() : Messages.ReplayAction_rebuild_displayName();
     }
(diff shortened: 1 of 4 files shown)


bitwiseman (approved):

bitwiseman:
@jglick  Can we get a test to verify this?

jglick:
This is all GUI. I verified it interactively and have no plans to spend time on an automated test.

car-roll (approved):

basil:
One amusing side effect of this change is that if a user lacks the `Run/Replay` permission and [Rebuilder](https://plugins.jenkins.io/rebuild/) is installed, there are now two "Rebuild" buttons (rather than the old behavior of a "Rebuild" and "Replay" button): one to rebuild the job with Rebuilder (which supports customizing the parameters for the rebuilt job) and the other to rebuild the job with Pipeline: Groovy (which does not, hence why administrators still install Rebuilder).

jglick:
That plugin does not appear to ensure that branch project builds use the same revision for `checkout scm`, or other Pipeline specifics, so **Rebuild** from that plugin without customizing parameters is not necessarily the same as **Rebuild** here.

PR #440, opened 2021-05-08
Skip test if jenkins version is too new to fix PCT failure

res0nance:
Due to jenkinsci/jenkins#5275 a test breaks PCT and is blocking jenkinsci/bom#511.

Because now there is no way to produce a null value we skip the test on newer versions of jenkins

Tested manually with `mvn clean install -Djenkins.version=2.289`

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/ParamsVariableTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/ParamsVariableTest.java
@@ -66,7 +66,7 @@ public class ParamsVariableTest {
 
     @Issue("JENKINS-42367")
     @Test public void nullValue() throws Exception {
-        Assume.assumeTrue(Jenkins.getVersion().isOlderThan(new VersionNumber("2.281")));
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/ParamsVariableTest.java
```diff
@@ -63,6 +66,7 @@
 
     @Issue("JENKINS-42367")
     @Test public void nullValue() throws Exception {
+        Assume.assumeTrue(Jenkins.getVersion().isOlderThan(new VersionNumber("2.281")));
```
jglick, line 69:
```suggestion
        Assume.assumeTrue(Jenkins.getVersion().isOlderThan(new VersionNumber("2.281"))); // TODO delete test when updating baseline
```

jglick (approved):
Fine, or just delete the test.

res0nance:
ping @car-roll

car-roll (approved):

PR #448, opened 2021-06-05
Inlining `InterceptingExecutorService`

basil:
Jenkins core defines `jenkins.util.InterceptingExecutorService`, which extends `com.google.common.util.concurrent.ForwardingExecutorService`. This exposes a Guava API in the Jenkins public API, which is a liability. If the Guava API changed, the Jenkins API might also have to change, which could result in incompatibilities. Supporting such an API also implies that Jenkins core must expose Guava to plugins, something we would rather avoid if possible.

Analyzing usages in both open source and CloudBees proprietary plugins with `usage-in-plugins`, I see that the only usage of this class is in `workflow-cps`. It seems better to move this class to `workflow-cps`. That way `workflow-cps` still gets what it needs, while Jenkins core gets out of the business of exposing Guava classes in its public API.

Concomitant with this PR, I am deprecating `jenkins.util.InterceptingExecutorService` in Jenkins core. Once this PR is widely adopted, I plan to delete `jenkins.util.InterceptingExecutorService` from Jenkins core.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/InterceptingExecutorService.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/InterceptingExecutorService.java
@@ -1,0 +1,84 @@
+package org.jenkinsci.plugins.workflow.cps;
+
+import com.google.common.util.concurrent.ForwardingExecutorService;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * {@link ExecutorService} that wraps all the tasks that run inside.
+ *
+ * @author Kohsuke Kawaguchi
+ * @since 1.557
+ */
+public abstract class InterceptingExecutorService extends ForwardingExecutorService {
+    private final ExecutorService base;
+
+    public InterceptingExecutorService(ExecutorService base) {
+        this.base = base;
+    }
+
+    protected abstract Runnable wrap(Runnable r);
+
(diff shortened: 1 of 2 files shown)


dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/cps/InterceptingExecutorService.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * {@link ExecutorService} that wraps all the tasks that run inside.
+ *
+ * @author Kohsuke Kawaguchi
+ * @since 1.557
+ */
+public abstract class InterceptingExecutorService extends ForwardingExecutorService {
```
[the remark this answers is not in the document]
basil: That makes sense to me and sounds like a good next step for future work. In the meantime this PR is a strict improvement over the current status quo.

src/main/java/org/jenkinsci/plugins/workflow/cps/InterceptingExecutorService.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * {@link ExecutorService} that wraps all the tasks that run inside.
+ *
+ * @author Kohsuke Kawaguchi
+ * @since 1.557
```
daniel-beck, line 19:
```suggestion
```

src/main/java/org/jenkinsci/plugins/workflow/cps/InterceptingExecutorService.java
daniel-beck, line 21: If possible, `@Restricted(NoExternalUse.class)`, plenty of plugins depend on this one.

car-roll (approved):

PR #380, opened 2020-09-30
Update parent POM

jglick:
PCT compatibility with https://github.com/jenkinsci/jenkins/pull/4944.

PR #452, opened 2021-06-30
Improve durability of `SURVIVABLE_NONATOMIC` while retaining the skipping of `fsync(2)`

basil:
See [JENKINS-66001](https://issues.jenkins.io/browse/JENKINS-66001). Downstream of jenkinsci/jenkins#5599, jenkinsci/workflow-api-plugin#163, and jenkinsci/workflow-support-plugin#120. Part 4 of a 5-part series to make Pipeline's `SURVIVABLE_NONATOMIC` mode behave [as advertised in the documentation](https://www.jenkins.io/doc/book/pipeline/scaling-pipeline/#what-am-i-giving-up-with-this-durability-setting-trade-off) (part 5 is in jenkinsci/workflow-job-plugin#199).

I tested this by running a Pipeline job with 100 steps in `SURVIVABLE_NONATOMIC` mode while simultaneously attaching a remote debugger to `XmlFile` and monitoring `fsync(2)` calls with [`syncsnoop.bt`](https://github.com/iovisor/bpftrace/blob/6cb740241edddc4b88e039bfc9385e6a3bad3abb/tools/syncsnoop.bt). I confirmed that `AtomicFileWriter` was being used in the Java debugger, and I confirmed that `fsync(2)` was not being used with `syncsnoop.bt`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,12 +23,12 @@
   ~ THE SOFTWARE.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.18</version>
+        <version>4.19</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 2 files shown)


basil:
A clean CI run of this PR will depend on removing 2.276 from the `Jenkinsfile`, which I do not have commit access to do.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -504,9 +505,17 @@ private TimingFlowNodeStorage createStorage() throws IOException {
...
+            wrappedStorage = new SimpleXStreamFlowNodeStorage(this, getStorageDir());
+            wrappedStorage.setAvoidAtomicWrite(!hint.isAtomicWrite());
+            wrappedStorage.setAvoidForce(!hint.isForce());
+        } else {
+            wrappedStorage = new BulkFlowNodeStorage(this, getStorageDir());
+            /*
+             * avoidAtomicWrite and avoidForce are always true for BulkFlowNodeStorage; see its
+             * constructor for details.
+             */
+        }
```
basil, line 518: Discovered this preëxisting bug while testing this series of changes. While we deal with calling `setAvoidAtomicWrite` (and, as of this PR, `setAvoidForce`) for the wrapping `TimingFlowNodeStorage` returned by this method earlier on line 402, nothing ever dealt with calling `setAvoidAtomicWrite` (and, as of this PR, `setAvoidForce`) on the wrapped `SimpleXStreamFlowNodeStorage` or `BulkFlowNodeStorage` being created here. During my testing, I found that some code was calling the wrapped storage and not the wrapper, and that call wasn't respecting the durability settings. This hunk fixes that problem.

pom.xml
```diff
@@ -64,21 +64,21 @@
     <properties>
         <revision>2.93</revision>
         <changelist>-SNAPSHOT</changelist>
-        <gitHubRepo>${project.artifactId}-plugin</gitHubRepo>
-        <jenkins.version>2.222.4</jenkins.version>
+        <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
```
jglick, line 67: Thanks, looks like I messed this up in #380.

jglick (approved):

PR #473, opened 2021-10-28
Pass through `FlowInterruptedException` from `Pickle.rehydrate`

jglick:


car-roll (approved):

jglick:
@car-roll could this be released please, so the downstream https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/176 can be readied? I would recommend setting up JEP-229 on these repos to minimize the labor of publishing changes.

PR #487, opened 2021-11-29
Ensure compatibility with Guava 31+

abayer:
Remove usage of `ImmutableMap`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -36,7 +36,6 @@ import com.cloudbees.jenkins.support.api.Component;
 import com.cloudbees.jenkins.support.api.Container;
 import com.cloudbees.jenkins.support.api.Content;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -485,7 +484,7 @@ public String getScript() {
     }
 
     public Map<String,String> getLoadedScripts() {
-        return ImmutableMap.copyOf(loadedScripts);
+        return Collections.unmodifiableMap(loadedScripts);
```
jglick, line 487: No, this is not a replacement. Perhaps you meant
```suggestion
        return Collections.unmodifiableMap(new HashMap<>(loadedScripts));
```

?

jglick (changes requested):

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
abayer: Fixed - thanks!

jglick (approved):

car-roll (approved):

PR #491, opened 2021-12-08
Make workflow editor horizontally resizable as well

five-iron:
Add expandable handle to east side of the workflow scrip editor. If anyone knows why this was not configured this way originally please feel free to explain.


jglick:
Would need to be revisited after #653.

jglick:
Does not appear to have any impact when I try it in Firefox. Anyway the editor consumes the full horizontal width available in the config screen, so I am not sure what more you would need?

five-iron:
Ah yeah this is no longer needed I think after the configuration page UI update a while back. Thanks!

PR #493, opened 2021-12-16
Exclude `javax.servlet:servlet-api`

basil:
With recent versions of Maven Enforcer, any plugin depending on `workflow-cps` starts getting errors like this:

```
Require upper bound dependencies error for javax.servlet:servlet-api:0 [provided] paths to dependency are:
+-org.jenkins-ci.plugins:gitlab-plugin:1.5.25-SNAPSHOT
  +-org.jenkins-ci.plugins.workflow:workflow-cps:2.94 [test]
    +-org.jenkins-ci.ui:ace-editor:1.1 [test] (managed) <-- org.jenkins-ci.ui:ace-editor:1.1 [test]
      +-javax.servlet:servlet-api:0 [provided] (managed) <-- javax.servlet:servlet-api:2.4 [provided]
```

While the root cause of the problem is that `ace-editor` needs refreshing, excluding the problematic `provided` version of `javax.servlet:servlet-api` in `workflow-cps` at the point that `ace-editor` is consumed is easily a fraction of the work and chases away the problem, so I think it is a good temporary measure.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,12 @@
         <dependency>
             <groupId>org.jenkins-ci.ui</groupId>
             <artifactId>ace-editor</artifactId>
+            <exclusions>
+                <exclusion> <!-- TODO remove when ace-editor is refreshed -->
+                    <groupId>javax.servlet</groupId>
+                    <artifactId>servlet-api</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.cloudbees</groupId>


basil:
@car-roll I think getting this into the next release (and subsequently the BOM) would certainly reduce the friction associated with the upcoming Maven Enforcer upgrade. While it isn't the ideal solution, `ace-editor` has 273,063 installations and hasn't been refreshed in 6 years, so I don't quite feel like tackling it just yet. =)

car-roll:
the tests seem to pass without `servlet-api`, but does it affect anything in the UI that the code here wouldn't be testing?

basil:
> the tests seem to pass without `servlet-api`, but does it affect anything in the UI that the code here wouldn't be testing?

No. The dependency is `provided`, and the version shipped by core is an empty JAR. So this has zero effect at runtime.

car-roll (approved):

car-roll:
@basil https://github.com/jenkinsci/workflow-cps-plugin/releases/tag/2648.va9433432b33c

basil:
Thank you!

PR #506, opened 2022-02-16
Update tests for SECURITY-2463 and SECURITY-2595 for compatibility with recent versions of Git plugin and Windows

dwnusbaum:
The tests for security fixes for https://www.jenkins.io/security/advisory/2022-02-15/ fail against recent version of Git plugin due to https://github.com/jenkinsci/git-plugin/pull/1207, and also fail when running on Windows due to a difference in the behavior.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinitionTest.java
@@ -272,7 +272,10 @@ public class CpsScmFlowDefinitionTest {
         FileUtils.copyDirectory(new File(sampleRepo.getRoot(), ".git"), gitDirInSvnRepo);
         String jenkinsRootDir = r.jenkins.getRootDir().toString();
         // Add a Git post-checkout hook to the .git folder in the SVN repo.
-        Files.write(gitDirInSvnRepo.toPath().resolve("hooks/post-checkout"), ("#!/bin/sh\ntouch '" + jenkinsRootDir + "/hook-executed'\n").getBytes(StandardCharsets.UTF_8));
+        Path postCheckoutHook = gitDirInSvnRepo.toPath().resolve("hooks/post-checkout");
+        // Always create hooks directory for compatibility with https://github.com/jenkinsci/git-plugin/pull/1207.
+        Files.createDirectories(postCheckoutHook.getParent());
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -179,7 +179,7 @@ public boolean isLightweight() {
             }
 
             FilePath scriptFile = dir.child(expandedScriptPath);
-            if (!new File(scriptFile.getRemote()).getCanonicalFile().toPath().startsWith(dir.absolutize().getRemote())) { // TODO JENKINS-26838
+            if (!new File(scriptFile.getRemote()).getCanonicalFile().toPath().startsWith(new File(dir.getRemote()).getCanonicalPath())) { // TODO JENKINS-26838
```
dwnusbaum, line 182: The only intended difference here is that if `JENKINS_HOME` is a symlink, the code should no longer throw an exception in non-dangerous scenarios.

wmeeus:
Would this also resolve that this check fails in case `JENKINS_HOME/workspace` is a symlink pointing outside the `$JENKINS_HOME` tree? (Use case: cloud setup with `JENKINS_HOME` in persistent storage and the workspace in local, faster ephemeral storage)

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
jglick, line 182: Did you mean
```suggestion
            if (!new File(scriptFile.getRemote()).getCanonicalFile().toPath().startsWith(new File(dir.getRemote()).getCanonicalFile().toPath())) { // TODO JENKINS-26838
```
?

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
jglick: or https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/139#discussion_r809047683

dwnusbaum: Discussed in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/139#discussion_r809043590.

dwnusbaum:
@wmeeus Yes, if you are using symlinks for `JENKINS_HOME` itself, or for the workspace directory, this PR should prevent false positives related to the recent security fixes.

PR #519, opened 2022-03-30
Use a concurrent map to manage timings

Vlatombe:
When using a large number of concurrent steps, writing timings back can
become a contention point as it locks the `CpsFlowExecution` object.
Replace the timings implementation by a concurrent map and only use atomic operations.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -92,6 +92,7 @@ import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.Stack;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
@@ -384,8 +385,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     }
 
     /** accumulated time in ns of a given {@link TimingKind#name}; {@link String} key for pretty XStream form */
-    @GuardedBy("this")
-    @CheckForNull Map<String, Long> timings;
+    @NonNull Map<String, Long> timings = new ConcurrentHashMap<>();
 
     @Deprecated
     public CpsFlowExecution(String script, FlowExecutionOwner owner) throws IOException {
@@ -413,6 +413,10 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     private Object readResolve() {
         if (loadedScripts==null)
             loadedScripts = new HashMap<>();   // field added later
+        // Convert timings to concurrent hash map
+        if (!(timings instanceof ConcurrentHashMap)) {
+            timings = timings == null ? new ConcurrentHashMap<>() : new ConcurrentHashMap<>(timings);
+        }
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -384,8 +385,7 @@ public void setResumeBlocked(boolean resumeBlocked) {
     }
 
     /** accumulated time in ns of a given {@link TimingKind#name}; {@link String} key for pretty XStream form */
-    @GuardedBy("this")
-    @CheckForNull Map<String, Long> timings;
+    @NonNull Map<String, Long> timings = new ConcurrentHashMap<>();
```
jglick, line 388: We just got rid of concurrent persistent fields in #518, let us not now reintroduce one! :grin: (see that PR for how to run the relevant tests in Java 17)

jglick (commented):
Nonblocking collections are probably overkill here. I think it would suffice to use a simple `HashMap` if we just synchronized on `timings` rather than `this`, and either did some trick with `volatile` or initialized it both in the constructor and `readResolve` so we do not run into race conditions with the null check.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
Vlatombe: Actually, the test command passes for me. Probably related to https://github.com/jenkinsci/workflow-cps-plugin/pull/519/files#diff-1c6ff2489c8f8b9b41176212ef2e09f3e26ee2967a9ed4a8b483710026d2cbe7R1640

jglick: Oh I see.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -424,16 +429,7 @@ private Timing(TimingKind kind) {
...
-                if (timings == null) {
-                    timings = new HashMap<>();
-                }
-                Long orig = timings.get(kind.name());
-                if (orig == null) {
-                    orig = 0L;
-                }
-                timings.put(kind.name(), orig + System.nanoTime() - start);
-            }
+            timings.merge(kind.name(), System.nanoTime() - start, Long::sum);
```
jglick, line 432: nice

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: I'm not sure whether that code actually changes the serialized type. It might just be that XStream has a special converter for for `ConcurrentHashMap` but not `ConcurrentSkipListMap`, e.g. https://github.com/x-stream/xstream/blob/2d2f5e9b407a4b0877a4e8cd9f3b18d019fe4b0d/xstream/src/java/com/thoughtworks/xstream/converters/collections/MapConverter.java#L71-L75.

dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
car-roll: I think I am seeing this very problem where a test is deserializing an old build, but it is failing on the `close` method due to `timings` being null

PR #528, opened 2022-04-24
Migrate from readResolve to unmarshal

car-roll:
If a build is loaded from a saved state, and the build existed before #519 (i.e. v2686.v7c37e0578401), It is possible for the `timings` field to be `NULL`. This causes an NPE when the Timing is closed: https://github.com/jenkinsci/workflow-cps-plugin/blob/3f09155513c1c8ed3727ec16220701bee0c31a9c/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L432

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1604,6 +1604,8 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     // the execution in Groovy CPS should hold that lock (or worse, hold that lock in the runNextChunk method)
     // so that the execution gets suspended while we are getting serialized
 
+    // Note: XStream ignores readResolve and writeReplace methods on types with custom Converter implementations, so use marshal and unmarshal instead.
+
     public static final class ConverterImpl implements Converter {
         private final ReflectionProvider ref;
         private final Mapper mapper;
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -428,7 +428,13 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             this.kind = kind;
             start = System.nanoTime();
         }
+
+        @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
         @Override public void close() {
+            // it is possible for timings to be null if an old build (< v2686.v7c37e0578401) is being loaded from a saved state
+            if (timings == null) {
+                timings = new ConcurrentHashMap<>();
+            }
             timings.merge(kind.name(), System.nanoTime() - start, Long::sum);
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -428,7 +428,13 @@ private Timing(TimingKind kind) {
             this.kind = kind;
             start = System.nanoTime();
         }
+
+        @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
         @Override public void close() {
+            // it is possible for timings to be null if an old build (< v2686.v7c37e0578401) is being loaded from a saved state
+            if (timings == null) {
```
car-roll, line 435: I basically just put back the null check that was removed in #519

olamy (approved):

Vlatombe:
I wonder if this is not a symptom of inconsistent (or at least, surprising) deserialization, where `unmarshal` is called but not `readResolve`.

dwnusbaum:
> I wonder if this is not a symptom of inconsistent (or at least, surprising) deserialization, where unmarshal is called but not readResolve.

I tested this really quickly by adding logger calls in each method, and it does look like only `unmarshal` gets called during deserialization.

> Should the current unmarshal method be patched to call readResolve at the end?

Maybe? I took a quick look through `xstream` to see if there was any discussion or documentation about combining custom converters with `readResolve`, but I did not notice anything. I could also see an argument for moving the logic in `readResolve` into `unmarshal` and deleting `readResolve`, at least as long as that doesn't cause any other issues.

jglick:
> moving the logic in `readResolve` into `unmarshal` and deleting `readResolve`

That sounds appropriate.

car-roll:
When I move the code to https://github.com/jenkinsci/workflow-cps-plugin/blob/3f09155513c1c8ed3727ec16220701bee0c31a9c/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L1715 `timings` is still null.

Looking at the stack trace, I realize it is only calling `WorkflowRun.reload()`: https://github.com/jenkinsci/workflow-job-plugin/blob/04c4e75dce438ba5af785fbf3a7711cacd54f41e/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L538
So it is completely bypassing `CpsFlowExecution.unmarshal()` and just calling `hudson.XmlFile.unmarshal` via `WorkflowJob.getLastBuild()`

dwnusbaum:
@car-roll I think you would need to inline `readResolve` later on (e.g. [right before the `return` statement](https://github.com/jenkinsci/workflow-cps-plugin/blob/3f09155513c1c8ed3727ec16220701bee0c31a9c/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L1755)), not add the logic into the `else if` block, because if `timings` is missing from the XML file (i.e. if it is null), that block never gets entered.

car-roll:
@dwnusbaum  thank you for pointing out that glaring error 🤦 
All is well now!

dwnusbaum (approved):
Might be worth adding a quick comment to `ConverterImpl`:
```
// Note: XStream ignores readResolve and writeReplace methods defined on types with custom Converter implementations, so use marshal and unmarshal instead.
```

jglick:
@car-roll https://github.com/jenkinsci/workflow-cps-plugin/pull/528#discussion_r858722052

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
car-roll: yikes, removed in #529

PR #535, opened 2022-05-11
Never print warnings from `CpsStepContext.completed`, but use `addSuppressed`

jglick:
Amending #490, and a string of earlier attempts to log repeated attempts to mark a step as failed, all of which had problems. Most recently, in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180 I find a case where first a `node` block is marked failed by `FlowInterruptedException`, and then a nested `sh` step throws `MissingContextVariableException` as well. Harmless; can use `addSuppressed` to at least track everything.


car-roll (approved):

jglick:
Lots of test flakes in this plugin recently. Not sure if there has been some infrastructure change or what.

jglick:
Can cause a stack overflow error.

jglick:
I believe this is safe to release at any time.

dwnusbaum (approved):

PR #539, opened 2022-05-19
[JENKINS-45327] Pickle EnvActionImpl to prevent multiple instances from existing after resumption

dwnusbaum:
See [JENKINS-45327](https://issues.jenkins.io/browse/JENKINS-45327).

Right now, if you use `env` in a Pipeline in a way that causes `EnvActionImpl` to be serialized into the program's state, then after resumption you end up with a distinct instance of `EnvActionImpl` that is unrelated to the action attached to the `WorkflowRun`, so the `owner` for the deserialized `EnvActionImpl` is null, and `getProperty` and `setProperty` throw NPEs.

This PR fixes the issue by pickling `EnvActionImpl` so that it can be rehydrated using the action attached to the `WorkflowRun`, which maintains the expected singleton semantics and prevents issues.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -215,17 +215,18 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
      * in case {@code env} is serialized into the program.
      */
     private static class EnvActionImplPickle extends Pickle {
+        @Override
         public ListenableFuture<?> rehydrate(FlowExecutionOwner owner) {
-            IOException cause = null;
             try {
                 Queue.Executable executable = owner.getExecutable();
                 if (executable instanceof Run) {
                     return Futures.immediateFuture(EnvActionImpl.forRun((Run)executable));
+                } else {
+                    return Futures.immediateFailedFuture(new IllegalStateException("Invalid executable: " + executable));
                 }
             } catch (IOException e) {
-                cause = e;
+                return Futures.immediateFailedFuture(e);
             }
-            return Futures.immediateFailedFuture(new RuntimeException("Unable to find Run for EnvActionImpl: " + owner, cause));
         }
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -24,6 +24,8 @@
 
 package org.jenkinsci.plugins.workflow.cps;
 
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
 import groovy.lang.GroovyObjectSupport;
 import hudson.EnvVars;
 import hudson.Extension;
@@ -32,7 +34,6 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.util.LogTaskListener;
 import java.io.IOException;
-import java.io.Serializable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -32,28 +34,30 @@
...
 import org.jenkinsci.plugins.workflow.steps.EnvironmentExpander;
 import org.jenkinsci.plugins.workflow.support.actions.EnvironmentAction;
+import org.jenkinsci.plugins.workflow.support.pickles.SingleTypedPickleFactory;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 @ExportedBean
-public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAction.IncludingOverrides, Serializable, RunAction2 {
```
dwnusbaum, line 56: This will cause Pipelines which already have a serialized instance of `EnvActionImpl` to fail, but those instances would have thrown an NPE anyway if they were used, so my thinking is that this should not affect a significant number of users. We could leave it in place if we want to be extra careful though.

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -199,4 +203,29 @@ public Collection<EnvironmentContributor> getEnvironmentContributors() {
...
+            IOException cause = null;
+            try {
+                Queue.Executable executable = owner.getExecutable();
+                if (executable instanceof Run) {
+                    return Futures.immediateFuture(EnvActionImpl.forRun((Run)executable));
+                }
+            } catch (IOException e) {
+                cause = e;
+            }
+            return Futures.immediateFailedFuture(new RuntimeException("Unable to find Run for EnvActionImpl: " + owner, cause));
```
dwnusbaum, line 228: I guess this doesn't need to be a `RuntimeException` any more (at first I was just `throw`ing the exception here). Maybe clearer to use `IOException`, idk.

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -199,4 +203,29 @@ public Collection<EnvironmentContributor> getEnvironmentContributors() {
...
+        protected Pickle pickle(EnvActionImpl object) {
+            return new EnvActionImplPickle();
+        }
+    }
+
+    /**
+     * Prevents multiple instances of {@link EnvActionImpl} from existing for a single Pipeline after a Jenkins restart
+     * in case {@code env} is serialized into the program.
+     */
+    private static class EnvActionImplPickle extends Pickle {
```
dwnusbaum, line 217: This has no internal state, we just use it as a hook to return `EnvActionImpl.forRun(owner.getExecutable())` during deserialization, completely discarding the serialized object.

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
jglick, line 228:
```suggestion
            try {
                Queue.Executable executable = owner.getExecutable();
                if (executable instanceof Run) {
                    return Futures.immediateFuture(EnvActionImpl.forRun((Run)executable));
                } else {
                    return Futures.immediateFailedFuture(new IllegalStateException("unexpected: " + executable));
                }
            } catch (IOException e) {
                return Futures.immediateFailedFuture(e);
            }
```

?

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -527,4 +527,26 @@ public boolean isAllowed(String groovyResourceUrl) {
...
+                    "echo('EnvActionImpl instance: ' + System.identityHashCode(e))\n", false));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("wait/1", b);
+        });
+        sessions.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getLastBuild();
+            SemaphoreStep.success("wait/1", null);
+            r.assertBuildStatus(Result.SUCCESS, r.waitForCompletion(b));
+            // Check that the EnvActionImpl attached to the run is the same object as the one in the program.
```
jglick, line 548: Rather than this low-level check, why not simply assert that the final env includes `foo=bar`, which is the actual goal?

dwnusbaum: Yeah good point. At first I just wanted to show that the instance was the same, but checking the value of `foo` on the action attached to the run does that indirectly anyway and tests the user-facing behavior which seems better, so I updated the test in 06761ef2c7b6aac6da9659476ca93feed3d0238c.

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
dwnusbaum: Done in 06761ef2c7b6aac6da9659476ca93feed3d0238c.

src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
jglick: I think removing `Serializable` makes sense here.

jglick (approved):

car-roll (approved):

PR #540, opened 2022-05-20
Call `CpsVmExecutorService.shutdown` from `CpsFlowExecution.suspendAll`

jglick:
https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/195/commits/19c2fdeab67597016b9cb960cda999839421d112 turned up a curious test failure. It turned out that a synchronous nonblocking step completed long after Jenkins had shut down, but since the test uses `JenkinsSessionRule` rather than `RealJenkinsRule`, the JVM continued running and the completion was actually processed by finishing the build (successfully) in the old copy of Jenkins. Needless to say, things got confused when the same build in the new copy of Jenkins tried to continue running, and began doing stuff with the still uncompleted program state despite the build record on disk claiming the build had completed.

There are probably multiple things that could be fixed here (and I will try to make the test behave more realistically to start with), but one problem seems to be that `suspendAll`, which runs during Jenkins shutdown and pauses all running Pipeline builds and makes sure `program.dat` is flushed to disk etc., was neglecting to actually turn off the CPS VM thread, which then continued to accept new tasks. We do not expect or want any tasks to be running in the CPS VM thread after the program has supposedly been serialized and Jenkins is about to exit, as the behavior could be anomalous; consider a `sleep` step which just happens to finish sleeping during this window. Better to reject any further tasks and let `StepExecution.onResume` and other mechanisms do their job.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
@@ -436,12 +436,9 @@ public class CpsStepContext extends DefaultStepContext { // TODO add XStream cla
                     }
                     outcome = new Outcome(null, new AlreadyCompleted());
                 }
-
-                /**
-                 * Program state failed to load.
-                 */
                 @Override
                 public void onFailure(Throwable t) {
+                    LOGGER.log(Level.WARNING, "Failed to proceed after " + CpsStepContext.this, t);
                 }
             });
         } catch (IOException x) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1584,6 +1584,15 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                             }
                         }
                         cpsExec.checkpoint(true);
+                        cpsExec.runInCpsVmThread(new FutureCallback<CpsThreadGroup>() {
+                            @Override public void onSuccess(CpsThreadGroup g) {
+                                LOGGER.fine(() -> "shutting down CPS VM threadin for " + cpsExec);
+                                g.shutdown();
+                            }
+                            @Override public void onFailure(Throwable t) {
+                                LOGGER.log(Level.WARNING, null, t);
+                            }
+                        });
                     }
                 } catch (Exception ex) {
                     LOGGER.log(Level.WARNING, "Error persisting Pipeline execution at shutdown: "+((CpsFlowExecution) execution).owner, ex);
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -640,6 +640,10 @@ public final class CpsThreadGroup implements Serializable {
         }
     }
 
+    void shutdown() {
+        runner.shutdown();
+    }
+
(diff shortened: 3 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -436,12 +436,9 @@ public void onSuccess(CpsThreadGroup g) {
                     }
                     outcome = new Outcome(null, new AlreadyCompleted());
                 }
-
-                /**
-                 * Program state failed to load.
-                 */
                 @Override
                 public void onFailure(Throwable t) {
+                    LOGGER.log(Level.WARNING, "Failed to proceed after " + CpsStepContext.this, t);
```
jglick, line 441: e.g.

```
…	WARNING	o.j.p.w.cps.CpsStepContext$2#onFailure: Failed to proceed after CpsStepContext[7:waitWithoutAgent]:Owner[p/1:p #1]
org.jenkinsci.remoting.util.ExecutorServiceUtils$FatalRejectedExecutionException: Cannot execute the command java.util.concurrent.FutureTask@…. The executor service is shutting down
	at hudson.remoting.SingleLaneExecutorService.execute(SingleLaneExecutorService.java:116)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at jenkins.util.InterceptingExecutorService.submit(InterceptingExecutorService.java:46)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$5.onSuccess(CpsFlowExecution.java:916)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$5.onSuccess(CpsFlowExecution.java:912)
	at org.jenkinsci.plugins.workflow.support.concurrent.Futures$1.run(Futures.java:147)
	at org.jenkinsci.plugins.workflow.support.concurrent.DirectExecutor.execute(DirectExecutor.java:33)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1277)
	at com.google.common.util.concurrent.AbstractFuture.addListener(AbstractFuture.java:761)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.addListener(AbstractFuture.java:136)
	at org.jenkinsci.plugins.workflow.support.concurrent.Futures.addCallback(Futures.java:157)
	at org.jenkinsci.plugins.workflow.support.concurrent.Futures.addCallback(Futures.java:97)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.runInCpsVmThread(CpsFlowExecution.java:912)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.scheduleNextRun(CpsStepContext.java:382)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.completed(CpsStepContext.java:327)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.onSuccess(CpsStepContext.java:319)
	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:49)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

dwnusbaum (approved):

PR #544, opened 2022-05-26
[JENKINS-68610] Re-enable admin auto approval when job config submitted via CLI/REST

yaroslavafenkin:
[JENKINS-68610](https://issues.jenkins.io/browse/JENKINS-68610)


src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -897,7 +904,6 @@ public void cpsScriptNonAdminConfiguration() throws Exception {
 
         jenkins.submit(config);
 
-        assertEquals(1, ScriptApproval.get().getPendingScripts().size());
```
yaroslavafenkin, line 900: Removed because this sometimes causes test failures locally, it was added in https://github.com/jenkinsci/workflow-cps-plugin/commit/6bd4e8bf837ba1df4394efdd003c70bb1b9ac33f by me. This assertion doesn't make much sense on its own, checking that specific script didn't get approved on the next line should be enough.

dwnusbaum (approved):

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
dwnusbaum: Does it fail because the value is 0 or because it is more than 1? And does it fail intermittently or consistently?

yaroslavafenkin: It's consistently more than 1, and I've got no idea how it wasn't failing before 🤔  
I could alternatively clean approved & pending scripts before each test to make that one work, but again don't see much value added in this assertion.

Pldi23 (approved):
LGTM

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
dwnusbaum: I suspect that scripts saved by non-admins result in two pending scripts since we do not call `ApprovalContext.withKey` here, which is [the only thing](https://github.com/jenkinsci/script-security-plugin/blob/4bd517d6dbf0926c3d506005e11187006d8f556f/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java#L455-L458) that would prevent multiple pending scripts with the same content since we call `configuring` twice in most cases (constructor + newInstance/readResolve). Maybe we should pass `Run.getExternalizableId()` (maybe with `CpsFlowDefinition.class.getName()` as well?) as the key? `SecureGroovyScript` has [a convenience method](https://github.com/jenkinsci/script-security-plugin/blob/4bd517d6dbf0926c3d506005e11187006d8f556f/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java#L164) to use the current `Item` as the key which should prevent this issue for it.

dwnusbaum (changes requested):
I think we should try to fix https://github.com/jenkinsci/workflow-cps-plugin/pull/544#discussion_r883680143 while we're here even if it hasn't been reported as a bug yet.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
dwnusbaum: Actually, `pendingScripts` is already defined as a set and equality on its members is just the script and the language, so what I noted shouldn't matter at all. Are the pending scripts in this case distinct?

yaroslavafenkin: I haven't looked at it under debugger. I'll check now.

yaroslavafenkin: So the other pending one is from another test

dwnusbaum: Ah ok, false alarm then, sorry!

ckulla:
Works like a charm! Tested the plugin from here: https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/workflow/workflow-cps/2706.vd6763c1272b_f/

ckulla:
@dwnusbaum @yaroslavafenkin Please merge this PR and release a new version of the plugin.

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
jglick: Maybe just move this test case to another class not using `@ClassRule`?

src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -0,0 +1,252 @@
+package org.jenkinsci.plugins.workflow.cps;
```
yaroslavafenkin, line 1: Should I add the license thing here?

PR #568, opened 2022-07-20
Resolve a race condition in FlowDurabilityTest causing test flakes

Riliane:
A race condition in `FlowDurabilityTest.testResumeBlockedAddedAfterRunStart` caused the test to flake.

```
org.junit.ComparisonFailure: expected:<[sleep]> but was:<[End of Pipeline]>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.jenkinsci.plugins.workflow.cps.FlowDurabilityTest.assertIncludesNodes(FlowDurabilityTest.java:665)
at org.jenkinsci.plugins.workflow.cps.FlowDurabilityTest$23.evaluate(FlowDurabilityTest.java:788)
```

The last "sleep" node was not loaded on restart.
Even though all individual nodes were persisted properly, a change to the head node ID was being done by another thread and was not completing before the original thread went on to restart Jenkins.
`CpsThreadGroup.notifyNewHead` starts a thread every time it needs to write a new head to the build.xml file. (in line 499)
Timeline of the failure:

- the sleep step node is persisted with id 5 in an xml file
- `notifyNewHead` is called with node with id 5 which is the sleep step node - thread updating the head starts
- main thread starts writing snapshot
- thread created in `notifyNewHead` reaches the writing stage - too late, snapshot has been written with the old head!

The solution is to wait for CpsFlowExecution to go into SUSPENDED state as it finishes persisting build.xml before going on with the restart.
The same flake seems to be happening in two more tests - likely same solution applies.

- [n/a] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -250,6 +250,20 @@ public class FlowDurabilityTest {
         assertHasTimingAction(run.getExecution());
     }
 
+    /** A minimal check of resuming the job with no node checks, in case of a dirty shutdown.
+     * Currently dirty shutdowns have the potential to lose the head node ID if they happen at the moment of writing
+     * it to build.xml
+     */
+    static void verifyDirtyResumed(JenkinsRule rule, WorkflowRun run, String logStart) throws Exception {
+        assert run.isBuilding();
+        assertHasTimingAction(run.getExecution());
+        rule.waitForCompletion(run);
+        Assert.assertEquals(Result.SUCCESS, run.getResult());
+        verifyCompletedCleanly(rule.jenkins, run);
+        //no checking nodes
+        rule.assertLogContains(logStart, run);
+    }
+
     /** If it's a {@link SemaphoreStep} we test less rigorously because that blocks async GraphListeners. */
     static void verifySafelyResumed(JenkinsRule rule, WorkflowRun run, boolean isSemaphore, String logStart) throws Exception {
         assert run.isBuilding();
@@ -716,7 +730,6 @@ public class FlowDurabilityTest {
                 WorkflowRun run = createAndRunSleeperJob(story.j.jenkins, jobName, FlowDurabilityHint.MAX_SURVIVABILITY, false);
                 FlowExecution exec = run.getExecution();
                 if (exec instanceof CpsFlowExecution) {
-                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
                     assert ((CpsFlowExecution) exec).getStorage().isPersistedFully();
                 }
                 logStart[0] = JenkinsRule.getLog(run);
@@ -727,7 +740,7 @@ public class FlowDurabilityTest {
             @Override
             public void evaluate() throws Throwable {
                 WorkflowRun run = story.j.jenkins.getItemByFullName(jobName, WorkflowJob.class).getLastBuild();
-                verifySafelyResumed(story.j, run, false, logStart[0]);
+                verifyDirtyResumed(story.j, run, logStart[0]);
             }
         });
     }
@@ -746,7 +759,6 @@ public class FlowDurabilityTest {
                 FlowExecution exec = run.getExecution();
                 Assert.assertTrue(((CpsFlowExecution) exec).isResumeBlocked());
                 if (exec instanceof CpsFlowExecution) {
-                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
                     assert ((CpsFlowExecution) exec).getStorage().isPersistedFully();
                 }
                 Assert.assertFalse(((CpsFlowExecution) exec).getProgramDataFile().exists());
@@ -761,7 +773,6 @@ public class FlowDurabilityTest {
             public void evaluate() throws Throwable {
                 WorkflowRun run = story.j.jenkins.getItemByFullName(jobName, WorkflowJob.class).getLastBuild();
                 verifyFailedCleanly(story.j.jenkins, run);
-                assertIncludesNodes(nodesOut, run);
             }
         });
     }
@@ -779,7 +790,6 @@ public class FlowDurabilityTest {
                 WorkflowRun run = createAndRunSleeperJob(story.j.jenkins, jobName, FlowDurabilityHint.MAX_SURVIVABILITY, false);
                 FlowExecution exec = run.getExecution();
                 if (exec instanceof CpsFlowExecution) {
-                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
                     assert ((CpsFlowExecution) exec).getStorage().isPersistedFully(); // single node xmls written
                 }
                 nodesOut.addAll(new DepthFirstScanner().allNodes(run.getExecution()));
@@ -793,7 +803,6 @@ public class FlowDurabilityTest {
             public void evaluate() throws Throwable {
                 WorkflowRun run = story.j.jenkins.getItemByFullName(jobName, WorkflowJob.class).getLastBuild();
                 verifyFailedCleanly(story.j.jenkins, run);
-                assertIncludesNodes(nodesOut, run);
             }
         });
     }
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -777,7 +777,8 @@ public class FlowDurabilityTest {
                 WorkflowRun run = createAndRunSleeperJob(story.j.jenkins, jobName, FlowDurabilityHint.MAX_SURVIVABILITY, false);
                 FlowExecution exec = run.getExecution();
                 if (exec instanceof CpsFlowExecution) {
-                    assert ((CpsFlowExecution) exec).getStorage().isPersistedFully();
+                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
+                    assert ((CpsFlowExecution) exec).getStorage().isPersistedFully(); // single node xmls written
                 }
                 nodesOut.addAll(new DepthFirstScanner().allNodes(run.getExecution()));
                 nodesOut.sort(FlowScanningUtils.ID_ORDER_COMPARATOR);
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -716,6 +716,7 @@ public class FlowDurabilityTest {
                 WorkflowRun run = createAndRunSleeperJob(story.j.jenkins, jobName, FlowDurabilityHint.MAX_SURVIVABILITY, false);
                 FlowExecution exec = run.getExecution();
                 if (exec instanceof CpsFlowExecution) {
+                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
                     assert ((CpsFlowExecution) exec).getStorage().isPersistedFully();
                 }
                 logStart[0] = JenkinsRule.getLog(run);
@@ -745,6 +746,7 @@ public class FlowDurabilityTest {
                 FlowExecution exec = run.getExecution();
                 Assert.assertTrue(((CpsFlowExecution) exec).isResumeBlocked());
                 if (exec instanceof CpsFlowExecution) {
+                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
                     assert ((CpsFlowExecution) exec).getStorage().isPersistedFully();
                 }
                 Assert.assertFalse(((CpsFlowExecution) exec).getProgramDataFile().exists());
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -255,8 +255,7 @@ public class FlowDurabilityTest {
      * Currently dirty shutdowns have the potential to lose the head node ID if they happen at the moment of writing
      * it to build.xml
      */
-    static void verifyDirtyResumed(JenkinsRule rule, WorkflowRun run, String logStart) throws Exception {
-        assert run.isBuilding();
+    private static void verifyDirtyResumed(JenkinsRule rule, WorkflowRun run, String logStart) throws Exception {
         assertHasTimingAction(run.getExecution());
         rule.waitForCompletion(run);
         Assert.assertEquals(Result.SUCCESS, run.getResult());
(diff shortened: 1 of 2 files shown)


jmdesprez (approved):
LGTM

Pldi23 (commented):
Changes seems clear to me, let's fix other flaky tests in this class (I assume the fix would be similar). 
I also see the PR was not build because of test failure, what are your thoughts is it other flake? If yes probably good idea would be to fix it also to avoid multiple releases?

Riliane:
@Pldi23 it's a different test class:
```
org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionRJRTest.smokes
connect: Address is invalid on local machine, or port is not valid on remote machine
```
Probably an unstable connection issue. I could take a look into the logs and see if I can hunt that down

jglick:
also see https://github.com/jenkinsci/workflow-cps-plugin/pull/570#issuecomment-1192679404

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -777,7 +777,8 @@ public void evaluate() throws Throwable {
                 WorkflowRun run = createAndRunSleeperJob(story.j.jenkins, jobName, FlowDurabilityHint.MAX_SURVIVABILITY, false);
                 FlowExecution exec = run.getExecution();
                 if (exec instanceof CpsFlowExecution) {
-                    assert ((CpsFlowExecution) exec).getStorage().isPersistedFully();
+                    ((CpsFlowExecution) exec).waitForSuspension(); // till done writing head node ID into build.xml
```
jglick, line 780: `waitForSuspension` should be deprecated. Generally there is something higher-level you want to wait for.

Riliane:
@jglick yes, `testFullyDurableSurvivesDirtyRestart` is the exact same flake - the writing of the head ID into the build.xml is cut short, the head is not set to the sleep node, the sleep node is not reloaded on restart, and the test then checks in `verifySafelyResumed` whether the head node is a semaphore or sleep one.

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
jglick: > if there was discussion of deprecation

Basically dating to the early days of ~Workflow~ Pipeline when the method was even introduced. We managed to get rid of most calls and there are just a few remaining ones to be cleaned up.

> were there suggestions on what else does this better?

Depends on the circumstance. If you are waiting for a build to get a certain point, `SemaphoreStep` is the usual idiom.

I am not very familiar with the test cases here. Basically the comment

> till done writing head node ID into build.xml

should guide you. Is that the actual thing you want to wait for? If so, sleep until `build.xml` contains it.

Riliane: I do not need to wait for the build progress. I do not need the sleep step to end, we are shutting down when it's still going. We just need a little time to persist properly.

> If so, sleep until build.xml contains it.
Do you mean a cycle of checking the contents of the file and sleeping more if they are incorrect?

jglick: Something like that, yes. (http://www.awaitility.org/ does that, but a simple `while` loop suffices.) Again, assuming that you are confident that the thing you are truly interested in waiting for is a particular field in this file.

Riliane: I honestly do not think a loop of file read is a great solution.
Firstly, now, instead of waiting _until the last head is done being reacted to and written by another thread_, we wait _until head is set to number X in the file_. That number X will be different and have to be counted separately for each pipeline if we want to reuse this pattern in some other test.
Secondly, what if a bug is introduced into the code, which the tests should be helping us find, and the correct head is _never_ written in the file? Then the waiting will simply time out, which seems less informative than restarting and getting "expected head X, got head Y" that we were getting here.

You could phrase it like that - the particular field in this file - or phrase it as waiting for the thread that reacts to a new head being set to be done with its job (writing included). `waitForSuspension` did it because the purpose of it seems to be waiting until `CpsThreadGroup` threads are done with their jobs.

Alternatively, we are currently considering if this case couldn't actually be something that happens during a real life Jenkins dirty restart, if Jenkins does shut down in a very specific moment while this thread is still not done, and an expected consequence of it being a "dirty restart".

jglick: > if this case couldn't actually be something that happens during a real life Jenkins dirty restart

Probably it can.

As I think I mentioned, I am not very familiar with any of the tests in this class, but from a brief look my guess is that `assertIncludesNodes` is just not useful to test and should be deleted unless we can actually make hard guarantees about behavior, which seems unlikely in the absence of transactional storage.

Riliane: If we decide to accept it as acceptable collateral of a dirty restart and do that, what about `testFullyDurableSurvivesDirtyRestart` which also checks the head node (that it is a semaphore step or sleep) in `verifySafelyResumed`? (Going to investigate how the job continues if it loses the sleep node and the head node is set as `{`.)

(Also I wonder if this can't cause a step to run twice - step starts but head is not written in time, shutdown happens, head is set back, step restarts)

mikecirioli: @jglick I think part of our struggle is in understanding what the expected behavior is for a durable pipeline based on the various states of persistence it can be in at the moment jenkins is killed.  (for example, the question in the previous comment).  Is this something you have some knowledge of (or know someone who does?)

jglick: > I wonder if this can't cause a step to run twice

I suspect it could, if you were very unlucky. Most builds spend the vast majority of their time inside some `sh` step, during which time there should be no step metadata changing (JENKINS-52165 complicates this a bit—the push mode is cleaner but not yet default). If Jenkins happens to crash at the moment that some metadata is being written, there can be problems. We try to use `AtomicFile` to defend against corruption of a _single_ file (even that is not necessarily reliable, depending on the filesystem; I think #452 etc. pertains to this), but the current simplistic build storage mechanism (which was really intended as an MVP) divides important information between

* `FlowExecution` parts of `build.xml`
* `program.dat`
* `workflow/*.xml`

and these files are not written in a properly journaled order.

> Is this something you have some knowledge of

Not a lot; the author of this test (and the whole system of picking different durability levels) no longer works on Jenkins.

Riliane: If all the head node related assertions were commented out and the pipeline was let continue with that malformed head, it logged "Sleeping for 30 sec" a total of _three_ times (once before restart, twice after) and then a different assertion in verifySucceededCleanly failed because it counted 9 nodes instead of 8. So the pipeline continuing "normally" in this case is quite dubious.

jglick: Indeed that sounds dubious. If you have the ambition to rewrite the storage system better (and we would have to consider whether to move to a transactional database or something), I am not sure what there is to do except relax these tests, or just delete them if they are not really proving anything interesting.

dwnusbaum: > I wonder if this can't cause a step to run twice

Yes, I have seen steps run twice when trying to create automated tests for edge cases related to resumption and persistence. I feel like I have even seen it in a support ticket or two, but I cannot recall any specifics.

> that assertIncludesNodes is just not useful to test and should be deleted unless we can actually make hard guarantees about behavior

I agree with this. Generally speaking, I think that any test which uses `addStepWithDirtyShutdown` (or `thenWithHardShutdown`) has nondeterministic behavior given the current implementation of `workflow-cps` unless it sets up very precise initial conditions, which IMO kind of defeats the point of testing with `addStepWithDirtyShutdown` in the first place.

Personally, I would first try to swap the tests that use `addStepWithDirtyShutdown` to have only the most basic assertions (e.g., `verifyFailedCleanly`, `verifySafelyResumed`). If even that is not reliable, then I would probably just `@Ignore` them.

mikecirioli: thanks for the info @dwnusbaum && @jglick , that is very helpful!

Riliane: thank you for the info, it does help!
I'll however go a bit back to `verifySafelyResumed` - this also checks the last node upon restart as well as the final amount of nodes in the finished pipeline. How do we define _safely resumed_ if we skip these assertions? What should we test for here?

mikecirioli: Since we control the content of the pipeline we should have a good expectation of what a correctly resumed and completed run should look like, and would could just compare the final pipeline state to that.  Unfortunately, I think we end up back in the same place (flake town) because of the unpredictability of `addStepWithDirtyShutdown`.   It seems like `workflow-cps` makes a best effort to safely persist its state so that it can resume but does not make any guarantees.  I am leaning toward more or less ignoring these flakey tests until such time as workflow-cps has been made more bulletproof.

dwnusbaum: > How do we define safely resumed if we skip these assertions?

I think that checking the final build result and that the logs contain some expect content at least once is the most that we can currently expect to be reliable.

I would probably delete the assertions related to `heads` and `node`, replace all calls to `verifySucceededCleanly` with `j.assertBuildStatus(Result.SUCCESS, run); verifyCompletedCleanly(run)`, and then delete `verifySucceededCleanly`.

jglick: > the unpredictability of `addStepWithDirtyShutdown`

Of course that is the whole purpose of this method—to shake out problems like these.

> `workflow-cps` makes a best effort to safely persist its state

Maybe “some” effort if not the best. :grimacing:

Riliane: maybe the node checks could stay for clean shutdown tests (eg `testDurableAgainstCleanRestartSurvivesIt`), since they do guarantee us correct node persistence AFAIK?

dwnusbaum: Sure, that seems fine assuming we have never seen it flake.

Pldi23 (approved):

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -251,6 +251,20 @@ static void verifySucceededCleanly(Jenkins j, WorkflowRun run) throws Exception
         assertHasTimingAction(run.getExecution());
     }
 
+    /** A minimal check of resuming the job with no node checks, in case of a dirty shutdown.
+     * Currently dirty shutdowns have the potential to lose the head node ID if they happen at the moment of writing
+     * it to build.xml
+     */
+    static void verifyDirtyResumed(JenkinsRule rule, WorkflowRun run, String logStart) throws Exception {
```
jglick, line 258:
```suggestion
    private static void verifyDirtyResumed(JenkinsRule rule, WorkflowRun run, String logStart) throws Exception {
```

jglick (approved):
Looks OK—again, I am not deeply familiar with this. @dwnusbaum?

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
Riliane: I copied that signature off `verifySafelyResumed` as it's a much simplified replacement for that

jglick: Maybe so, I am just reviewing and seeing a newly introduced method referred to only within the same class, so it should be `private` because it can. Friendlier to IDEs, human readers, etc. to not have to look around for other references.

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -251,6 +251,20 @@ static void verifySucceededCleanly(Jenkins j, WorkflowRun run) throws Exception
...
+        assert run.isBuilding();
```
dwnusbaum, line 259: I know you copied this from other tests, but IMO it is very strange and potentially confusing to use Java `assert` statements in tests:

```suggestion
        Assert.assertTrue(run.isBuilding());
```

Also though, it looks like this assertion is potentially timing sensitive (maybe the build finishes very quickly). Maybe better to just delete it and not check anything before `rule.waitForCompletion`.

dwnusbaum (approved):
Everything looks generally fine to me, but I would make the new method `private` like Jesse mentioned.

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
Riliane: shall we do the same for `verifySafelyResumed`?

jglick: If you like.

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
dwnusbaum: > because it always does so on a waiting step like sleep/semaphore (every pipeline restart seems to be meant to happen on a sleep), and it can't successfully finish faster than the allotted sleep time

Unless we use something like the `semaphore` step to _guarantee_ that the build is not allowed to finish before we check `run.isBuilding`, I think there is a potential timing issue, even if it is very unlikely.

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -251,9 +251,22 @@ static void verifySucceededCleanly(Jenkins j, WorkflowRun run) throws Exception
...
     }
 
+    /** A minimal check of resuming the job with no node checks, in case of a dirty shutdown.
+     * Currently dirty shutdowns have the potential to lose the head node ID if they happen at the moment of writing
+     * it to build.xml
+     */
+    private static void verifyDirtyResumed(JenkinsRule rule, WorkflowRun run, String logStart) throws Exception {
+        assertHasTimingAction(run.getExecution());
+        rule.waitForCompletion(run);
+        Assert.assertEquals(Result.SUCCESS, run.getResult());
```
jglick, line 261: BTW there is `JenkinsRule.assertBuildStatus` for this purpose, though it hardly matters.

jglick (approved):

PR #570, opened 2022-07-21
[JENKINS-49707] Implement `BodyExecution.cancel(Throwable)`

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/124.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -11,9 +11,7 @@ import com.cloudbees.groovy.cps.sandbox.SandboxInvoker;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.FutureCallback;
 import hudson.model.Action;
-import hudson.model.Result;
 import hudson.util.Iterators;
-import jenkins.model.CauseOfInterruption;
 import org.jenkinsci.plugins.workflow.actions.BodyInvocationAction;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
@@ -22,14 +20,12 @@ import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext;
 import org.jenkinsci.plugins.workflow.steps.BodyExecution;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
-import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
 import net.jcip.annotations.GuardedBy;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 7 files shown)


jglick:
Cannot reproduce test failure locally; looks like a flake (investigating).

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -243,13 +239,12 @@ public Collection<StepExecution> getCurrentExecutions() {
...
 
     @Override
-    public boolean cancel(final CauseOfInterruption... causes) {
+    public boolean cancel(Throwable error) {
         // 'stopped' and 'thread' are updated atomically
         CpsThread t;
         synchronized (this) {
             if (isDone())  return false;   // already complete
-            // TODO should perhaps rather override cancel(Throwable) and make this overload just delegate to that one
-            stopped = new FlowInterruptedException(Result.ABORTED, causes);
```
dwnusbaum, line 252: With these changes, any code that currently calls `cancel(Throwable)` with an exception that is not a `FlowInterruptedException` will no longer result in a `FlowInterruptedException` at runtime, so behavior would change. Did you check the PCT? I am not sure how many callers of `cancel(Throwable)` exist.

jglick: > I am not sure how many callers of `cancel(Throwable)` [that do not pass `FlowInterruptedException`] exist.

I could not find any in common Pipeline plugins. There seem to be a few plugins still improperly overriding `stop` (not updated like e.g. https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/49) that I could locate in my local Maven index:

* `DatabaseConnectionStep`
* `ExwsExecution`
* `GitLabBuildsStep`
* `GitLabCommitStatusStep`

> behavior would change

Indeed it would, though probably for the better—the precise reason for the failure of the body would be thrown up naturally.

dwnusbaum: Ok, that seems fine. I was just worried that if there was widespread usage then this might cause a bunch of test failures because of things like build results changing from `ABORTED` to `FAILURE`.

dwnusbaum (approved):

jglick:
> looks like a flake (investigating)

Did not manage to track down the cause. Seems that the same thread which ran `WorkflowRun.finish` also somehow triggered `CpsFlowExecution.logTimings`, which is called only from `CpsVmExecutorService.tearDown` and thus should only ever be called from the CPS VM thread. No hypothesis there.

`testFullyDurableSurvivesDirtyRestart` seems like a flake also. I do not understand this code well enough to analyze what the failure means.

src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
jglick: Checked `usage-in-plugins` with `-i -M` on

```
org.jenkinsci.plugins.workflow.steps.BodyExecution#cancel
```

resulting in

```
    "org/jenkinsci/plugins/workflow/steps/BodyExecution#cancel(Ljava/lang/Throwable;)Z": [
      "applitools-eyes",
      "browserstack-integration",
      "database",
      "external-workspace-manager",
      "gitlab-plugin",
      "performance-signature-dynatrace",
      "performance-signature-dynatracesaas",
      "pipeline-gitstatuswrapper",
      "pipeline-npm",
      "sauce-ondemand",
      "testingbot",
      "workflow-cps",
      "workflow-step-api"
    ]
```

so looks like the previously mentioned hits plus a few more obscure ones. Checking these quickly, they are all `StepExecution.stop` overrides. This idiom is deprecated but unlikely to lead to a behavioral change anyway, since if a block-scoped step is running another step in its body, `stop` will be called on the innermost.

dwnusbaum (approved):

car-roll (approved):

PR #582, opened 2022-08-23
feat: Switch to ionicons api

NotMyFault:
The change proposed makes use of the centralized ionicons API to remove the need to ship symbols on a per-plugin basis, that are provided by the library.

cc @jenkinsci/workflow-cps-plugin-developers


basil (approved):

car-roll (approved):
LGTM

PR #583, opened 2022-08-25
Accept run not loading at all as a kind of run failure in testDurableAgainstCleanRestartFailsWithDirtyShutdown

Riliane:
`testDurableAgainstCleanRestartFailsWithDirtyShutdown` occasionally throws an NPE and flakes, because the build.xml file of the run ends up empty after being interrupted by a dirty restart, as the writes for performance-optimized durability are not atomic. That's still a kind of run failure and a plausible consequence of a dirty restart, so I propose to accept this outcome and skip all the remaining assertions.

- [n/a] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -63,8 +63,6 @@ import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import org.junit.Assume;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  * Tests implementations designed to verify handling of the flow durability levels and persistence of pipeline state.
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
@@ -45,11 +45,15 @@ import org.junit.Test;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
-import org.jvnet.hudson.test.*;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileOutputStream;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -585,6 +585,7 @@ public void evaluate() throws Throwable {
             @Override
             public void evaluate() throws Throwable {
                 WorkflowRun run = story.j.jenkins.getItemByFullName("durableAgainstClean", WorkflowJob.class).getLastBuild();
+                if (run == null) { return; } //there is a small chance due to non atomic write that build.xml will be empty and the run won't load at all
```
jtnord, line 588: does the same check need to be done elsewhere where the `PERFORMANCE_OPTIMIZED` is used?

e.g. `testDurableAgainstCleanRestartFailsWithBogusStorageFile`?

Riliane: I shall look into it. On a first glance, maybe yes for `testDurableAgainstCleanRestartFailsWithBogusStorageFile` as it also utilizes a dirty shutdown and also expects a failure. I will look for its flake logs to see if the same issue has actually ever arisen

Riliane: I don't see any NPE flakes in that one. My guess would be that while we mess with the `flowNodeStore.xml` after starting our run to corrupt and fail it, it might be giving the extra time before shutdown that the build.xml write needs to be finished?..

jtnord (approved):

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
Riliane: It also seems not necessary when the test is based around a clean shutdown, or when there is a `cpsFlow.pause(true)` to force persistence (looks like this write should also happen with the pause). In one test there's a `Thread.sleep` to wait and let the run definitely start. I see no other logs of such a NPE - looks like in all other tests something else (pause, more file editing, sleeps) gives it time to write properly.

jtnord: thanks for checking 👍

jmdesprez (approved):

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
Riliane: @jtnord I've found an old failure log like that for `testDurableAgainstCleanRestartFailsWithBogusStorageFile` (guess it's more rare). Will add the check, then

jtnord (approved):

src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
dwnusbaum: > we would shut Jenkins down cleanly, ruin the flow node file while it's still down, and then try to boot it back up - but I don't think RestartableJenkinsRule allows for that?

If you are able to use `JenkinsSessionRule`, then you can write code that uses `JenkinsSessionRule.getHome` between two calls to `then` to modify files in `JENKINS_HOME` while Jenkins is not running, but yeah you cannot do that with `RestartableJenkinsRule`. Normally you can switch to `JenkinsSessionRule` easily, but here it is nontrivial because `JenkinsSessionRule` has no equivalent to `addStepWithDirtyShutdown`, and given all of the problems with the tests that use `addStepWithDirtyShutdown` today, it probably doesn't make sense to add an equivalent method to `JenkinsSessionRule`.

dwnusbaum (approved):

PR #597, opened 2022-10-19
[SECURITY-2824]

dwnusbaum:
A few PRs were merged to the `master` branch after https://github.com/jenkinsci/workflow-cps-plugin/commit/1af77ffcc773ffac4eb49956168b9778dca9a3b6 had been staged, so we cannot push the security fix directly and need to merge it into `master` instead.

I am marking this as `internal` for now to avoid creating a CD release since the advisory has only just gone out and we do not want users to see another update right away.


dwnusbaum:
CI build failure is caused by https://github.com/jenkins-infra/helpdesk/issues/3189.

jglick (approved):

PR #599, opened 2022-10-25
[JENKINS-69917] - Snippetizer/index.jelly javascript un-inlined.

Artmorse:
**The issue**
You can find all the issue details [here](https://issues.jenkins.io/browse/JENKINS-69917).

**My updates**
I've moved the `onclick` call from the `Snippetizer/index.jelly` file to the `handle-prototype.js` one.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
@@ -1,0 +1,28 @@
+
+function handlePrototype() {
+    buildFormTree(document.forms.config);
+    // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
+    var json = Object.toJSON(JSON.parse(document.forms.config.elements.json.value).prototype);
+    if (!json) {
+        return; // just a separator
+    }
+    const url = '/jenkins/pipeline-syntax/generateSnippet' // ${rootURL}/${it.GENERATE_URL}
+    new Ajax.Request(url, {
+        method: 'POST',
+        parameters: {json: json},
+        onSuccess: function(r) {
+            document.getElementById('prototypeText').value = r.responseText;
+        }
+    });
+}
+
+
+document.addEventListener('DOMContentLoaded', (event) => {
+
+    const generatePipelineScript = document.getElementById("generatePipelineScript");
+    generatePipelineScript.onclick = (_) => {
+        handlePrototype();
+        return false;
+    }
+
+});
--- a/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -67,27 +67,10 @@ THE SOFTWARE.
                     <f:dropdownListBlock title="${%— Advanced/Deprecated —}"/>
                     <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(true)}"/>
                 </f:dropdownList>
-                <j:set var="id" value="${h.generateId()}"/>
                 <f:block>
-                    <input type="button" value="${%Generate Pipeline Script}" onclick="handlePrototype_${id}(); return false" class="submit-button primary"/>
-                    <f:textarea id="prototypeText_${id}" readonly="true" style="margin-top: 10px" />
-                    <script>
-                        function handlePrototype_${id}() {
-                            buildFormTree(document.forms.config);
-                            // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
-                            var json = Object.toJSON(JSON.parse(document.forms.config.elements.json.value).prototype);
-                            if (!json) {
-                                return; // just a separator
-                            }
-                            new Ajax.Request('${rootURL}/${it.GENERATE_URL}', {
-                                method: 'POST',
-                                parameters: {json: json},
-                                onSuccess: function(r) {
-                                    document.getElementById('prototypeText_${id}').value = r.responseText;
-                                }
-                            });
-                        }
-                    </script>
+                    <input type="button" id="generatePipelineScript" value="${%Generate Pipeline Script}" class="submit-button primary"/>
+                    <f:textarea id="prototypeText" readonly="true" style="margin-top: 10px" />
+                    <st:adjunct includes="org.jenkinsci.plugins.workflow.cps.Snippetizer.handle-prototype"/>
                 </f:block>
                 <f:section title="${%Global Variables}"/>
                 <f:block>


src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -0,0 +1,28 @@
+
+function handlePrototype() {
+    buildFormTree(document.forms.config);
+    // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
+    var json = Object.toJSON(JSON.parse(document.forms.config.elements.json.value).prototype);
+    if (!json) {
+        return; // just a separator
+    }
+    const url = '/jenkins/pipeline-syntax/generateSnippet' // ${rootURL}/${it.GENERATE_URL}
```
jglick, line 9: This will not work if the context path is set to anything other than `/jenkins`.

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -67,27 +67,10 @@ THE SOFTWARE.
                     <f:dropdownListBlock title="${%— Advanced/Deprecated —}"/>
                     <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(true)}"/>
                 </f:dropdownList>
-                <j:set var="id" value="${h.generateId()}"/>
```
jglick, line 70: Loses uniquification of the textarea. OTOH https://github.com/jenkinsci/workflow-cps-plugin/blob/6467f4f1b7842ce7d0c3ed7ad01eb62b5098b724/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly#L78 already worked only with the singleton https://github.com/jenkinsci/workflow-cps-plugin/blob/6467f4f1b7842ce7d0c3ed7ad01eb62b5098b724/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly#L31 so I guess it does not matter.

jglick (changes requested):
Needs to be tested with snippet generator running in various contexts (root, job, multibranch project) and with different values for the Jenkins context path.

Artmorse:
Yes, I'll find some time to have a look to it this week!

jglick:
The page is accessible from e.g.
* `/pipeline-syntax/` (Jenkins root, e.g. https://ci.jenkins.io/pipeline-syntax/)
* that URL appended to a multibranch folder (do not have a publicly accessible live example for you)
* that URL appended to a job (including one branch project under a multibranch folder)

And you need to check when the context root is empty (like in ci.jenkins.io) as well as when it has a value (like `/jenkins` prefix used when you `mvn hpi:run` this plugin).

src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
jglick: I do not often work in JS, but there are various ways; IIRC the preferred technique is to define `data` attributes on HTML elements and look those up from JS. Search core sources for examples.

Artmorse:
I've some issues with `mvn hpi:run` so it's not tested yet.
image

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -0,0 +1,28 @@
...
+    // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
+    var json = Object.toJSON(JSON.parse(document.forms.config.elements.json.value).prototype);
+    if (!json) {
+        return; // just a separator
+    }
+    new Ajax.Request(url, {
+        method: 'POST',
+        parameters: {json: json},
+        onSuccess: function(r) {
+            document.getElementById('prototypeText').value = r.responseText;
```
jglick, line 13: also like https://github.com/jenkinsci/workflow-cps-plugin/pull/599#discussion_r1004453734

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -67,27 +67,11 @@ THE SOFTWARE.
...
-                                parameters: {json: json},
-                                onSuccess: function(r) {
-                                    document.getElementById('prototypeText_${id}').value = r.responseText;
-                                }
-                            });
-                        }
-                    </script>
+                    <input type="button" id="generatePipelineScript" value="${%Generate Pipeline Script}"
+                           class="submit-button primary" data-url="${rootURL}/${it.GENERATE_URL}"/>
+                    <f:textarea id="prototypeText" readonly="true" style="margin-top: 10px" />
```
jglick, line 73:
```suggestion
                    <f:textarea id="prototypeText" readonly="true" style="margin-top: 10px"/>
```
while we are here

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -0,0 +1,28 @@
+
```
jglick, line 1:
```suggestion
```

jglick (approved):
Looks right. Tested successfully?

jglick:
@Artmorse do you plan to resolve the merge conflict with #702 and retest?

Artmorse:
@jglick thanks for pinging me, I'll look at it this week!

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -67,36 +67,13 @@ THE SOFTWARE.
...
-                                        document.getElementById('prototypeText_${id}').value = responseText;
-                                    });
-                                }
-                            });
-                        }
-                    </script>
+                    <input type="button" id="generatePipelineScript" value="${%Generate Pipeline Script}"
+                           class="submit-button primary"
+                           data-url="${rootURL}/${it.GENERATE_URL}"
+                           data-crumb="${h.getCrumb(request)}"/>
```
Artmorse, line 74: I've added the jenkins crumb to avoid 403.
That's the easiest way, I found, to pass it to the javascript script.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -0,0 +1,37 @@
+function handlePrototype(url, crumb) {
+    buildFormTree(document.forms.config);
+    // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
+    // TODO simplify when Prototype.js is removed
+    const json = Object.toJSON ? Object.toJSON(JSON.parse(document.forms.config.elements.json.value).prototype) : JSON.stringify(JSON.parse(document.forms.config.elements.json.value).prototype);
+    if (!json) {
+        return; // just a separator
+    }
+
+    const xhr = new XMLHttpRequest();
```
jglick, line 10: Is there a reason this is using a different API than the script previously (at least as of #702) inside `index.jelly`, using `fetch`?

jglick (commented):
I do not think I am qualified to review this. Someone else?

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
Artmorse: It wasn't working anymore locally using `fetch` so I moved with that other HTTP library.

jglick: > It wasn't working anymore locally using `fetch`

But why would moving the code from inline `<script>` to an adjunct cause that? It apparently works in, e.g., https://github.com/jenkinsci/credentials-plugin/pull/443 so what went wrong here?

Artmorse: After some investigations, I've moved back to `fetch`!
see https://github.com/jenkinsci/workflow-cps-plugin/pull/599/commits/196f96ca5c3bc698e5a9c8d47e2e6ed4287dc701

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
jglick: Could also just add a `CrumbExclusion` for https://github.com/jenkinsci/workflow-cps-plugin/blob/4e2b31c20833a35b08a00e9f5b8c3a042b9820df/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java#L487 since it takes `POST` only to handle larger quantities of input but it is in fact read-only so not a CSRF threat. Anyway, this is the straightforward translation of the original.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -0,0 +1,43 @@
...
+        });
+}
+
+
+document.addEventListener('DOMContentLoaded', () => {
+
+    const generatePipelineScript = document.getElementById("generatePipelineScript");
+    const url = generatePipelineScript.getAttribute("data-url");
+    const crumb = generatePipelineScript.getAttribute("data-crumb");
+    generatePipelineScript.onclick = (_) => {
```
jglick, line 38: Using a `class` and `Behaviour.specify` is more conventional I think, but this should be fine as well.

jglick (approved):
Seems to work.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -67,36 +67,13 @@ THE SOFTWARE.
                     <f:dropdownListBlock title="${%— Advanced/Deprecated —}"/>
                     <local:listSteps quasiDescriptors="${it.getQuasiDescriptors(true)}"/>
                 </f:dropdownList>
-                <j:set var="id" value="${h.generateId()}"/>
```
jglick, line 70: Note that removing `generateId` and reverting to a hard-coded id loses the ability for this code to exist in multiple places in a page. Not an issue in this case since it exists only in one place in a top-level page. The general approach is to use behaviors and look up elements via relative DOM path.

PR #543, opened 2022-05-24
[JENKINS-63766] Work around JDK-8231454

basil:
### Background

See [JENKINS-63766](https://issues.jenkins.io/browse/JENKINS-63766) and upstream bug [JDK-8231454](https://bugs.openjdk.java.net/browse/JDK-8231454). Users have reported gradual exhaustion of the metaspace when running on Java 11 with Pipeline jobs that define custom classes.

### Steps to reproduce

Create a new Pipeline job with

```java
class asdf {
  void run () {}
}

new asdf().run()
```

and run it under Java 11 with `-verbose:class`. After the job finishes, run `jcmd ${PID} GC.run`

### Expected results

**Note:** These are the _actual_ results when running under Java 8 or Java 17.

The `WorkflowScript` and `asdf` classes get unloaded.

### Actual results

The `WorkflowScript` and `asdf` classes never get unloaded on Java 11.

### Evaluation

As described in the upstream bug, a memory memory leak can occur when `java.beans.Introspector#getBeanInfo` is invoked against a class, even if `java.beans.Introspector#flushFromCaches` is later called on that class. In our case, `java.beans.Introspector#getBeanInfo` is invoked from the following stack trace:

```
getBeanInfo:196, Introspector (java.beans)
run:3328, MetaClassImpl$15 (groovy.lang)
doPrivileged:-1, AccessController (java.security)
addProperties:3326, MetaClassImpl (groovy.lang)
initialize:3303, MetaClassImpl (groovy.lang)
getMetaClassUnderLock:289, ClassInfo (org.codehaus.groovy.reflection)
getMetaClass:331, ClassInfo (org.codehaus.groovy.reflection)
getMetaClass:277, MetaClassRegistryImpl (org.codehaus.groovy.runtime.metaclass)
getMetaClass:905, InvokerHelper (org.codehaus.groovy.runtime)
createCallConstructorSite:86, CallSiteArray (org.codehaus.groovy.runtime.callsite)
defaultCallConstructor:59, CallSiteArray (org.codehaus.groovy.runtime.callsite)
callConstructor:238, AbstractCallSite (org.codehaus.groovy.runtime.callsite)
call:208, Checker$3 (org.kohsuke.groovy.sandbox.impl)
onNewInstance:42, GroovyInterceptor (org.kohsuke.groovy.sandbox)
onNewInstance:173, SandboxInterceptor (org.jenkinsci.plugins.scriptsecurity.sandbox.groovy)
[...]

PR #600, opened 2022-10-26
Workaround for JDK-8231454 on Java 11 is no longer necessary as of 11.0.17

basil:
The workaround introduced in #543, while harmless on Java 11.0.17, is no longer necessary (because the fix was backported to 11.0.17 and later). Therefore, skip the workaround on those releases. If accepted I will open another PR to apply the same copypasta to `script-security`. To test this change, I stepped through the code in the debugger and verified that we applied the workaround on 11.0.16 but not on 11.0.17 or 17.0.4.


jglick (approved):
Fine, or just delete the workaround and we can add a release notes entry reminding people to keep their Java version up to date.

basil:
I do not think deleting the workaround is appropriate at this time when adoption of 11.0.17 is so low. It may be appropriate a few months down the road once adoption of 11.0.17 increases.

car-roll (approved):

PR #581, opened 2022-08-23
Update webpack, babel, npm and node

NotMyFault:
Misc frontend updates, does dependabot take a nap? I see no babel PRs 👀

PR #604, opened 2022-10-31
Update a few frontend dependencies

NotMyFault:
Follow up to https://github.com/jenkinsci/workflow-cps-plugin/pull/581

Visual testing didn't reveal any regressions, additionally, I added an entry to the dependabot configuration, to handle future updates automatically.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "yarn"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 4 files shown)


jglick (approved):
Thanks!

.github/dependabot.yml
```diff
@@ -13,3 +13,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "yarn"
```
jglick, line 16: FYI not added in #382 https://github.com/jenkinsci/workflow-cps-plugin/blob/cee026aaaab559d970d2b5d444a6a69775196185/.github/dependabot.yml#L5-L7

PR #596, opened 2022-10-17
Robustness of `CpsFlowExecution.getNode`

jglick:
Same stack trace as in https://github.com/jenkinsci/workflow-job-plugin/pull/300. Somewhere the build is corrupt in this case; unclear how, but a `NullPointerException` is not appropriate.

PR #612, opened 2022-11-01
Merge groovy-cps into workflow-cps

dwnusbaum:
This PR merges the entire history of https://github.com/cloudbees/groovy-cps into this plugin. `groovy-cps` is only supported in the context of Jenkins, and should only be used by `workflow-cps`. Keeping the library and plugin separate complicates maintenance and testing.

The only interesting changes should be in https://github.com/jenkinsci/workflow-cps-plugin/pull/612/commits/306d1ce99ea2d66a7f4441c1dbcb4e36f60b904b (moves `workflow-cps` into `plugin` module), https://github.com/jenkinsci/workflow-cps-plugin/pull/612/commits/96735ee1c080442ed388aa7c6695fe00041cce04 (merges `groovy-cps` into the repo), and any newer commits. Older commits are just the existing history of `groovy-cps`.

**With this PR, `workflow-cps` can only be built using Java 11 or newer.** The plugin will still run against Java 8, but the `dgm-builder` module, which is only used when building `groovy-cps`, uses Java compiler APIs that were modified in Java 9 and so do not work with Java 8.

TODO:
* [x] Add `workflow-cps-parent` and `groovy-cps` to https://github.com/jenkins-infra/repository-permissions-updater
    * PR merged: https://github.com/jenkins-infra/repository-permissions-updater/pull/2917
* [x] Check whether we need to add a new PCT hook to be able handle to handle `workflow-cps` after this change
    * PR merged: https://github.com/jenkinsci/plugin-compat-tester/pull/383
* [x] Test PCT: https://github.com/jenkinsci/bom/pull/1553 
* [x] Double-check that the merge looks correct
* [x] File PR to groovy-cps to delete all existing content
    * https://github.com/cloudbees/groovy-cps/pull/124 
* [ ] Push old groovy-cps tags into this repo
* [ ] Archive groovy-cps


LICENSE.txt
```diff
@@ -0,0 +1,177 @@
+
```
dwnusbaum, line 1: TODO: Should be deleted.

lib/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps-parent</artifactId>
+      <version>${changelist}</version>
+  </parent>
+
+  <groupId>com.cloudbees</groupId>
```
dwnusbaum, line 11: Should we switch the `groupId` for `groovy-cps` to `org.jenkins-ci.plugins.workflow` going forward? I don't think it really matters either way.

LICENSE.txt
dwnusbaum: https://github.com/jenkinsci/workflow-cps-plugin/pull/612/commits/33e276c2c53e7178d7a81afb2db15aaf42393483

dwnusbaum:
The CI build is trying to publish an incremental build of `groovy-cps-dgm-builder`, which we don't really want to do. I'm not sure if there is an easy way to make https://github.com/jenkins-infra/pipeline-library/blob/5288902cac5871e540ae9146e15e47cb47e6d757/vars/infra.groovy#L325 aware of `maven.deploy.skip`.

dwnusbaum:
Something to note is that because this PR imports groovy-sandbox's history, the first release with this commit with jump from version 2803.x to ~3500.x, which might look unusual to Jenkins users.

lib/pom.xml
jglick: I would leave the `groupId` alone unless you are sure there would be no consequences to changing it.

jglick (approved):

lib/pom.xml
dwnusbaum: Ok, looks like we agree. I will leave it.

dwnusbaum:
See https://github.com/jenkinsci/plugin-compat-tester/pull/383 for associated PCT changes.

dwnusbaum:
Rebuilding. (FTR, re-running a specific failed check from the left-hand sidebar on the checks page seems to work to trigger a new build, but the "re-run all checks" button from the top-right corner of the checks page does not seem to do anything, and you can only use the former if some check failed.)

dwnusbaum:
@jglick Do you know of any other examples of multi-module projects which skip deployment of a module that are able to publish incrementals? https://github.com/jenkinsci/workflow-cps-plugin/pull/612/commits/1b01482f7077021d68c23b1c266e3b669ff83182 makes things work, but skipping installation for `groovy-cps-dgm-builder` makes working with `groovy-cps` awkward, for example it breaks the ability to run specific tests in IDEs by default (because you always need to run tests from the parent project and run at least the `package` phase).

I'm not sure what other options we have. Maybe we could make `groovy-cps-dgm-builder` always be `1.0-SNAPSHOT` so that https://github.com/jenkins-infra/pipeline-library/blob/c86c8a970564019c5708e9b10e0717c28866309d/vars/infra.groovy#L333 ignores it, and then we could remove `<maven.install.skip>true</...>`?

jglick:
> Do you know of any other examples of multi-module projects which skip deployment of a module that are able to publish incrementals?

`plugin-compat-tester` itself but I guess it does not suffer from the issue you describe.

Maybe activate the profile to skip installation when `set.changelist` is defined (same as `produce-incrementals`)?

car-roll (approved):

dwnusbaum:
I am planning on merging this once the CI build passes. Then I'll merge master into #596 and release a new version of the plugin.

I will work on merging master into existing open PRs, but I will probably not be able to finish them all today. I'll also refile a few things from groovy-cps here today so I can archive it.

PR #613, opened 2022-11-04
[JENKINS-46163] Add support for spread and spread-map operators

dwnusbaum:
See [JENKINS-46163](https://issues.jenkins.io/browse/JENKINS-46163). This PR is a copy of https://github.com/cloudbees/groovy-cps/pull/123, which was based on https://github.com/cloudbees/groovy-cps/pull/116.

We follow the same approach as regular Groovy for `SpreadMapExpression` (the expression gets turned into a special `SpreadMap` by [`InvokerHelper.spreadMap`](https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/runtime/InvokerHelper.java#L378), and then Map literals are constructed using [`InvokerHelper.createMap`](https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/runtime/InvokerHelper.java#L399) which detects this marker class and interpolates it).

We use a similar approach for `SpreadExpression`, but this is not the approach that Groovy itself uses. Some reasoning behind this choice is given in the Javadoc for `SpreadBlock`.

This PR also includes various improvements to the test harness (565b1cac7cf2ebf2dd251c081a495e746d0f091f) which have been copied from `groovy-sandbox`. I used them in the new tests and adapted existing tests to use the new assertion methods. The two commits can be reviewed independently.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/Builder.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/Builder.java
@@ -32,6 +32,8 @@ import com.cloudbees.groovy.cps.impl.PropertyAccessBlock;
 import com.cloudbees.groovy.cps.impl.ReturnBlock;
 import com.cloudbees.groovy.cps.impl.SequenceBlock;
 import com.cloudbees.groovy.cps.impl.SourceLocation;
+import com.cloudbees.groovy.cps.impl.SpreadBlock;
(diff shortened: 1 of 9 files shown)


jglick (approved):
:shrug:

car-roll (approved):

PR #614, opened 2022-11-04
Convert GStringImpl to String in the RHS of CPS-transformed method pointers

dwnusbaum:
Fixes https://github.com/cloudbees/groovy-cps/issues/93.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/MethodPointerBlock.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/MethodPointerBlock.java
@@ -7,6 +7,7 @@ import com.cloudbees.groovy.cps.Next;
 import com.cloudbees.groovy.cps.sandbox.CallSiteTag;
 import com.cloudbees.groovy.cps.sandbox.Invoker;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import groovy.lang.GString;
 import java.util.Collection;
 import java.util.Collections;
 import org.codehaus.groovy.runtime.InvokerHelper;
@@ -64,6 +65,9 @@ public class MethodPointerBlock implements CallSiteBlock {
          * Obtain a method pointer, which is really just a {@link MethodClosure}.
          */
         public Next done(Object methodName) {
+            if (methodName instanceof GString) {
+                methodName = methodName.toString();
+            }
             return k.receive(e.getInvoker().contextualize(MethodPointerBlock.this).methodPointer(lhs, (String)methodName));
         }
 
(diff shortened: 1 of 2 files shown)


lib/src/test/java/com/cloudbees/groovy/cps/sandbox/SandboxInvokerTest.java
```diff
@@ -671,4 +671,19 @@ public void sandboxInterceptsImplicitCastsArrayAssignment() throws Throwable {
                 "LinkedHashMap.asBoolean()");
     }
 
+    @Test public void dynamicMethodPointer() throws Throwable {
+        assertIntercept(
+                "def method3() {\n" +
+                "    true\n" +
+                "}\n" +
+                "def mp = this.&/method${1 + 2}/\n" +
```
dwnusbaum, line 679: This already worked in regular Groovy, in `groovy-sandbox`, and in `@NonCPS` methods with `groovy-cps`, so the fix only matters for CPS-transformed method pointers.

lib/src/main/java/com/cloudbees/groovy/cps/impl/MethodPointerBlock.java
```diff
@@ -64,6 +65,9 @@ public Next fixLhs(Object lhs) {
          * Obtain a method pointer, which is really just a {@link MethodClosure}.
          */
         public Next done(Object methodName) {
+            if (methodName instanceof GString) {
+                methodName = methodName.toString();
```
jglick, line 69: Can this call something outside the sandbox?

dwnusbaum: No, this is the same as code like this in `DSL.java`: https://github.com/jenkinsci/workflow-cps-plugin/blob/a8fc49e2f96fbb9aa92860b6fa053c70734f0452/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java#L601-L602

It can call closures embedded in the `GString` (simple expressions have already been fully evaluated by this point), but any closures it calls will have been CPS and sandbox-transformed.

dwnusbaum: Also though, the behavior of `GString`-embedded closures in this context is totally broken (regardless of this patch), even worse than described [here](https://www.jenkins.io/doc/book/pipeline/cps-method-mismatches/#PipelineCPSmethodmismatches-ClosuresinsideGString) (somehow the argument to the closure is passed as an extra argument to `new GStringImpl`). Maybe we should update `visitGString` to just fail immediately if it contains a `ClosureExpression`.

That said, this PR is not really important, it just looked like it might be easy to fix.

jglick (approved):
Sounds OK.

car-roll (approved):

PR #622, opened 2022-11-07
[JENKINS-69905] Backport SECURITY-2824 for Jenkins 2.332.x users

dwnusbaum:
See [JENKINS-69905](https://issues.jenkins.io/browse/JENKINS-69905) and https://github.com/jenkinsci/workflow-cps-plugin/commit/1af77ffcc773ffac4eb49956168b9778dca9a3b6. Because `script-security` only requires Jenkins 2.332.1, users running Jenkins 2.332.x who install `script-security` and `workflow-cps` from the plugin manager will end up with incompatible versions (`script-security` will have the SECURITY-2824 fix, but `workflow-cps` will not). There is not currently an easy way to directly control the version of `script-security` offered to 2.332.x users (so that we could offer them the most recent version without the SECURITY-2824 fix), so instead we decided to just backport the SECURITY-2824 changes to the last version of `workflow-cps` that supports 2.332.x.

In retrospect, we should have bumped the minimum supported Jenkins version in `script-security` to 2.346.1 as part of the fix for SECURITY-2824 to avoid this kind of issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,7 +66,7 @@
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <jenkins.version>2.332.1</jenkins.version>
         <no-test-jar>false</no-test-jar>
-        <groovy-cps.version>1.32</groovy-cps.version>
+        <groovy-cps.version>1.34</groovy-cps.version>
         <node.version>12.19.0</node.version>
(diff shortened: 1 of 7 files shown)


jtnord (approved):
looks reasonable have not tested or done a diff comparison.

PR #627, opened 2022-11-13
Reduce reflection in `CpsFlowExecution`

basil:
Jenkins core currently specifies `--add-opens java.base/java.io=ALL-UNNAMED`, which is technical debt in the sense that it enables deprecated behavior. This is only needed for `workflow-cps`'s `cleanUpObjectStreamClassCaches` method, and that method is in turn only needed for old JVMs that do not have the fix for [JDK-8277072](https://bugs.openjdk.org/browse/JDK-8277072). This PR makes the reflection-based logic conditional on running on these older JVMs, so that when running on a newer JVM the illegal reflective access is not attempted. When this PR is widely adopted, we can remove the `--add-opens java.base/java.io=ALL-UNNAMED` directive from Jenkins core and clean up this technical debt.

To test this change I removed `--add-opens java.base/java.io=ALL-UNNAMED` and verified that this exception was logged even when running on Java 11.0.17:

```
java.lang.reflect.InaccessibleObjectException: Unable to make field static final java.io.ClassCache java.io.ObjectStreamClass$Caches.localDescs accessible: module java.base does not "opens java.io" to unnamed module @2cb2b9bd
        at java.base/java.lang.reflect.AccessibleObject.throwInaccessibleObjectException(AccessibleObject.java:387)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:363)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:311)
        at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:180)
        at java.base/java.lang.reflect.Field.setAccessible(Field.java:174)
        at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpObjectStreamClassCaches(CpsFlowExecution.java:1489)
        at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpLoader(CpsFlowExecution.java:1346)
        at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpHeap(CpsFlowExecution.java:1311)
        at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:464)
        at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.access$400(CpsThreadGroup.java:95)
[...]

PR #635, opened 2022-12-03
Require Java 11 or newer

basil:


jtnord (approved):

lib/src/test/java/com/cloudbees/groovy/cps/CpsTransformerTest.java
```diff
@@ -525,7 +527,7 @@ public void regexpOperator() throws Throwable {
     @Test
     public void arrayPassedToMethod() throws Throwable {
         assertEvaluate(4, "def m(x) {x.size()}; def a = [1, 2]; a.size() + m(a)"); // control case
-        assertEvaluate(4, "def m(x) {x.size()}; def a = [1, 2].toArray(); a.length + m(Arrays.asList(a))"); // workaround #1
+        assertEvaluate(4, "def m(x) {x.size()}; def a = [1, 2].toArray(); a.length + m(List.of(a))"); // workaround #1
```
dwnusbaum, line 530: For some reason which is not immediately apparent to me, this causes a `VerifyError` when running `groovy-cps`'s tests using Java 17. The only thing that seems special about this case compared to all of the other replacements is that this replacement is inside of a Groovy script.

<details>
<summary>Stack trace</summary>

```
java.lang.VerifyError: (class: java_util_List$of$4, method: callStatic signature: (Ljava/lang/Class;[Ljava/lang/Object;)Ljava/lang/Object;) Illegal type in constant pool
        at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
        at java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:3373)
        at java.base/java.lang.Class.getConstructor0(Class.java:3578)
        at java.base/java.lang.Class.getConstructor(Class.java:2271)
        at org.codehaus.groovy.reflection.ClassLoaderForClassArtifacts.defineClassAndGetConstructor(ClassLoaderForClassArtifacts.java:82)
        at org.codehaus.groovy.runtime.callsite.CallSiteGenerator.compileStaticMethod(CallSiteGenerator.java:262)
        at org.codehaus.groovy.reflection.CachedMethod.createStaticMetaMethodSite(CachedMethod.java:295)
        at org.codehaus.groovy.runtime.callsite.StaticMetaMethodSite.createStaticMetaMethodSite(StaticMetaMethodSite.java:114)
        at groovy.lang.MetaClassImpl.createStaticSite(MetaClassImpl.java:3421)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.createCallStaticSite(CallSiteArray.java:76)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.createCallSite(CallSiteArray.java:161)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:116)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:128)
        at Script2.run(Script2.groovy:1)
        at groovy.lang.GroovyShell.evaluate(GroovyShell.java:574)
        at groovy.lang.GroovyShell.evaluate(GroovyShell.java:612)
        at groovy.lang.GroovyShell.evaluate(GroovyShell.java:583)
        at com.cloudbees.groovy.cps.AbstractGroovyCpsTest.eval(AbstractGroovyCpsTest.java:93)
        ... 46 more

```
</details>

jglick: Possibly some issue with varargs. Just revert it.

dwnusbaum: No worries, I rolled it into https://github.com/jenkinsci/workflow-cps-plugin/pull/640. I only mentioned it here since it seems somewhat unusual.

jimklimov:
Looking at https://stackoverflow.com/a/27107189/4715872 and https://github.com/jenkinsci/workflow-cps-plugin/pull/627#issuecomment-1352869571 it seems the change all over the place to `List.of()` was toxic, causing problems if some of the legit items were `null` and worked with singletons or `asList` before.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
```diff
@@ -102,7 +102,7 @@ public static <E> Iterator<E> iterator(Set<E> set) {
 
     public static <E> Iterator<E> iterator(E[] array) {
         // TODO as above
-        return new Itr<>(Arrays.asList(array));
+        return new Itr<>(List.of(array));
```
jglick, line 105: This evidently caused a regression.

jglick:
@basil are you using some automated refactoring tool for these PRs? If so, please disable the `List.of` refactoring as it is not safe. https://github.com/jenkinsci/jenkins/pull/6700#discussion_r982620526 at least

basil:
Sorry for the inconvenience and thank you for fixing this. (Actually I would have jumped on this right away, but you tend to wake up a few hours earlier than me on the East Coast.) No I am not using an automated refactoring tool. The automated tools don't make this recommendation because it is not provably safe. However it is safe most of the time, and I tend to rely on a combination of quick code inspection and automated test coverage to ascertain with relatively high probability whether the change is safe or not. Lack of test coverage means that I've been wrong a handful of times, like in this case and in core, but altogether this heuristic works more often than not.

jglick:
> I would have jumped on this right away

Oh no problem, I just got alerted to it my morning time via Gitter. Anyway it is ultimately up to maintainers to determine whether a given PR is safe; in the future we will inspect `List.of` calls with more care.

PR #640, opened 2022-12-12
[JENKINS-70108] CastBlock must contextualize invoker to avoid issues in mixed-trust scenarios

dwnusbaum:
See [JENKINS-70108](https://issues.jenkins.io/browse/JENKINS-70108).

https://github.com/jenkinsci/workflow-cps-plugin/commit/819cc9be597121a2bf88b7669323aa6caa502f72 introduced a new `Block` for casts, `CastBlock`, which did not implement `CallSiteBlock` and did not contextualize its `Invoker`. In cases where untrusted code calls trusted code that performs a cast, this could result in trusted code calling `SandboxInvoker.cast` instead of `DefaultInvoker.cast`. In practical terms, this means that casts inside of global Pipeline libraries were sometimes incorrectly blocked by the sandbox.

This PR makes `CastBlock` implement `CallSiteBlock` so that it can contextualize `Invoker` correctly before performing the cast.

TODO:
* [x] Does `ContinuationGroup.castToBoolean` also need changes?
    * I am pretty sure that it technically does, but the current code would only block casts if the type being cast implements an `asBoolean` method that is not allowed by the sandbox, so IDK if it matters in practice.
    * See https://github.com/jenkinsci/workflow-cps-plugin/pull/640#discussion_r1047399933 and surrounding ignored code and comment.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CastBlock.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CastBlock.java
@@ -4,12 +4,13 @@ import com.cloudbees.groovy.cps.Block;
 import com.cloudbees.groovy.cps.Continuation;
 import com.cloudbees.groovy.cps.Env;
 import com.cloudbees.groovy.cps.Next;
-import java.util.Collections;
+import com.cloudbees.groovy.cps.sandbox.CallSiteTag;
+import java.util.Collection;
 import java.util.List;
 import org.codehaus.groovy.runtime.InvokerInvocationException;
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
 
-public class CastBlock implements Block {
+public class CastBlock extends CallSiteBlockSupport {
     private final SourceLocation loc;
     private final Block valueExp;
     private final Class<?> type;
@@ -17,7 +18,8 @@ public class CastBlock implements Block {
     private final boolean coerce;
     private final boolean strict;
 
-    public CastBlock(SourceLocation loc, Block valueExp, Class<?> type, boolean ignoreAutoboxing, boolean coerce, boolean strict) {
+    public CastBlock(SourceLocation loc, Collection<CallSiteTag> tags, Block valueExp, Class<?> type, boolean ignoreAutoboxing, boolean coerce, boolean strict) {
+        super(tags);
         this.loc = loc;
         this.valueExp = valueExp;
         this.type = type;
@@ -42,7 +44,7 @@ public class CastBlock implements Block {
 
         public Next cast(Object value) {
             try {
-                return k.receive(e.getInvoker().cast(value, type, ignoreAutoboxing, coerce, strict));
+                return k.receive(e.getInvoker().contextualize(CastBlock.this).cast(value, type, ignoreAutoboxing, coerce, strict));
             } catch (CpsCallableInvocation inv) {
                 // Implementations of asType and other methods used by the Groovy stdlib should be @NonCPS, but we
                 // just log a warning and invoke the callable anyway to maintain the existing behavior.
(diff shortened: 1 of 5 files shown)


lib/src/test/java/com/cloudbees/groovy/cps/sandbox/SandboxInvokerTest.java
```diff
@@ -696,4 +696,16 @@ public void sandboxInterceptsImplicitCastsArrayAssignment() throws Throwable {
...
+        TrustedCpsCompiler trusted = new TrustedCpsCompiler();
+        trusted.setUp();
+        getBinding().setVariable("trusted", trusted.getCsh().parse("def foo() { File f = ['secret.key'] }"));
+
+        assertIntercept(
+            "trusted.foo()", // Untrusted script
+            new File("secret.key"),
+            "Script1.super(Script1).setBinding(Binding)",
+            "Script1.trusted",
+            "Script1.foo()");
```
dwnusbaum, line 709: Previously, we would also intercept `new File(String)`, which was wrong because that is in trusted code that should not be subject to the sandbox.

lib/src/test/java/com/cloudbees/groovy/cps/CpsTransformerTest.java
```diff
@@ -527,7 +527,7 @@ public void regexpOperator() throws Throwable {
     @Test
     public void arrayPassedToMethod() throws Throwable {
         assertEvaluate(4, "def m(x) {x.size()}; def a = [1, 2]; a.size() + m(a)"); // control case
-        assertEvaluate(4, "def m(x) {x.size()}; def a = [1, 2].toArray(); a.length + m(List.of(a))"); // workaround #1
+        assertEvaluate(4, "def m(x) {x.size()}; def a = [1, 2].toArray(); a.length + m(Arrays.asList(a))"); // workaround #1
```
dwnusbaum, line 530: See https://github.com/jenkinsci/workflow-cps-plugin/pull/635#discussion_r1046343921.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CastBlock.java
```diff
@@ -4,20 +4,22 @@
...
 import com.cloudbees.groovy.cps.Next;
-import java.util.Collections;
+import com.cloudbees.groovy.cps.sandbox.CallSiteTag;
+import java.util.Collection;
 import java.util.List;
 import org.codehaus.groovy.runtime.InvokerInvocationException;
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
 
-public class CastBlock implements Block {
+public class CastBlock extends CallSiteBlockSupport {
```
dwnusbaum, line 13: I verified that this does not cause problems deserializing `CastBlock`s serialized before this change.

lib/src/test/java/com/cloudbees/groovy/cps/sandbox/SandboxInvokerTest.java
```diff
@@ -696,4 +696,44 @@ public void sandboxInterceptsImplicitCastsArrayAssignment() throws Throwable {
...
+    }
+
+    /*
+     * All blocks that perform boolean casts internally using ContinuationGroup.castToBoolean incorrectly intercept
+     * calls performed by the cast when the cast is in trusted code that is called by untrusted code. For boolean
+     * casts, this is not that interesting, since it only causes problems in practice if you cast a type that
+     * implements an asBoolean method that would not be allowed by the sandbox.
+     * I see two obvious ways to fix this:
+     * 1. Add a CallSiteBlock parameter to ContinuationGroup.castToBoolean, and use it to contextualize the invoker
+     *    used in that method. All Blocks that use the method would need to be updated to implement CallSiteBlock.
```
dwnusbaum, line 719: This seems like the simplest option. Let me know if you think it's worth fixing. There are eight `Block`s that would need to implement `CallSiteBlock` if we make this change.

jglick (approved):
Fine to the extent I understand it.

PR #642, opened 2022-12-15
Replacing `Arrays.asList` with `List.of` can introduce `NullPointerException`

jglick:
https://github.com/jenkinsci/workflow-cps-plugin/pull/635#discussion_r1049599451 https://github.com/jenkinsci/workflow-cps-plugin/pull/627#issuecomment-1352869571 @jimklimov

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
@@ -175,7 +175,7 @@ public class ArgumentsActionImpl extends ArgumentsAction {
             this.isUnmodifiedBySanitization = false;
             return NotStoredReason.OVERSIZE_VALUE;
         }
-        List inputList = List.of(objects);
+        List<Object> inputList = Arrays.asList(objects);
         Object sanitized = sanitizeListAndRecordMutation(inputList, variables);
         if (sanitized == inputList) { // Works because if not mutated, we return original input instance
             return objects;
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHackTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHackTest.java
@@ -256,4 +256,18 @@ public class IteratorHackTest {
             r.assertBuildStatusSuccess(r.waitForCompletion(b));
         });
     }
+
+    @Issue("https://github.com/jenkinsci/workflow-cps-plugin/pull/627#issuecomment-1352869571")
+    @Test public void arrayIterator() throws Exception {
+        rr.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class);
+            p.setDefinition(new CpsFlowDefinition(
+                    "Object[] arr = [1, null, 2]\n" +
+                    "for (def elt : arr) {\n" +
+                    "  echo(/iterating on $elt/)\n" +
+                    "}\n", false));
+            rr.j.assertLogContains("iterating on null", rr.j.buildAndAssertSuccess(p));
+        });
+    }
+
 }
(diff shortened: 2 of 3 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
```diff
@@ -175,7 +175,7 @@ Object sanitizeArrayAndRecordMutation(@NonNull Object[] objects, @CheckForNull E
             this.isUnmodifiedBySanitization = false;
             return NotStoredReason.OVERSIZE_VALUE;
         }
-        List inputList = List.of(objects);
+        List<Object> inputList = Arrays.asList(objects);
```
jglick, line 178: Also found this from a usages search that did not look obviously safe.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHackTest.java
```diff
@@ -256,4 +256,18 @@ public class IteratorHackTest {
...
+    @Issue("https://github.com/jenkinsci/workflow-cps-plugin/pull/627#issuecomment-1352869571")
+    @Test public void arrayIterator() throws Exception {
+        rr.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class);
+            p.setDefinition(new CpsFlowDefinition(
+                    "Object[] arr = [1, null, 2]\n" +
+                    "for (def elt : arr) {\n" +
+                    "  echo(/iterating on $elt/)\n" +
+                    "}\n", false));
+            rr.j.assertLogContains("iterating on null", rr.j.buildAndAssertSuccess(p));
```
jglick, line 269: reproduced as

```
java.lang.AssertionError: 
unexpected build status; build log was:
------
Started
[Pipeline] Start of Pipeline
[Pipeline] End of Pipeline
	at java.base/java.util.Objects.requireNonNull(Objects.java:221)
	at java.base/java.util.ImmutableCollections$ListN.<init>(ImmutableCollections.java:432)
	at java.base/java.util.List.of(List.java:1040)
	at org.jenkinsci.plugins.workflow.cps.persistence.IteratorHack.iterator(IteratorHack.java:106)
	at …
	at WorkflowScript.run(WorkflowScript:2)
	at …
Finished: FAILURE

------

Expected: is <SUCCESS>
     but: was <FAILURE>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.jvnet.hudson.test.JenkinsRule.assertBuildStatus(JenkinsRule.java:1438)
	at org.jvnet.hudson.test.JenkinsRule.assertBuildStatus(JenkinsRule.java:1444)
	at org.jvnet.hudson.test.JenkinsRule.buildAndAssertStatus(JenkinsRule.java:1491)
	at org.jvnet.hudson.test.JenkinsRule.buildAndAssertSuccess(JenkinsRule.java:1477)
	at org.jenkinsci.plugins.workflow.cps.persistence.IteratorHackTest.lambda$arrayIterator$4(IteratorHackTest.java:269)
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jtnord: [`@NonNull Object[] objects](https://github.com/jenkinsci/workflow-cps-plugin/pull/642/files#diff-72a51c4c67a1c8eaed43038451cf5b39f7221a5c486a7e8464b72481af91d0c2R173)`  would imply otherwise. either that annotation is wrong or the caller is violating this.

jtnord (approved):
approved but  with a comment so disabling auto merge so the comment can be seen and responded to (even if that is dismissed) and then a merge

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java
jglick: https://github.com/jenkinsci/workflow-cps-plugin/blob/7c908a0a2d9b16b68a0a6f010f183cc8c34b145e/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/actions/ArgumentsActionImpl.java#L173 refers to the array, which is not null. But `List.of` fails also when an _element_ is null (which the annotation does not mention).

In practice I doubt there could be null elements, because arrays would be of `Describable` subtypes which would require real values, but I would rather not find out. Safely handling inputs with null elements would require additional work.

PR #654, opened 2023-02-01
Dark, but not Dark (System), theme support

basil:
Fixes https://github.com/jenkinsci/dark-theme-plugin/issues/127. Can be merged and released as-is without any regressions, but depends on https://github.com/jenkinsci/dark-theme-plugin/pull/319 for the fix to actually take effect. I tested this by running in both the dark and light theme and verifying that the background was adjusted appropriately.

PR #653, opened 2023-02-01
Upgrade ACE Editor from 1.2.2 to 1.15.0

basil:
Upgrade [ACE Editor](https://ace.c9.io/) from 1.2.2 (released on October 28, 2015) to 1.15.0 (released on January 25, 2023).

### Implementation

Followed the example in the [Webpack demo here](https://github.com/ajaxorg/ace/blob/159aa70d551530f2866fb0006fad37bd75e60dda/demo/webpack/demo.js).

### Testing done

- Create a new Pipeline job and try samples
- Autocomplete Pipeline-specific snippets
- Introduce a syntax error and verify validation fails
- Fix a syntax error and verify validation succeeds
- Resize the editor

PR #658, opened 2023-02-09
[JENKINS-70591] Revert recent ACE Editor changes

jglick:
Reverting #653, #654, and #656 until a regression is tracked down and fixed.


basil (changes requested):
> Until

https://github.com/jenkinsci/workflow-cps-plugin/pull/657 ?

basil:
Was it fair to propose this revert just one day after the original regression was introduced when [JENKINS-70591](https://issues.jenkins.io/browse/JENKINS-70591) has remained open for 8 months?

PR #667, opened 2023-03-02
Call `GroovyClassLoader.close` when build completes

jglick:
When considering loading libraries from JAR files in https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/55 it occurred to me that a completed build might leave classpath JAR files open at least until garbage collection, which could be a problem particularly on Windows. Might also affect Grape users. Safest to close these deterministically. Untested beyond the many functional tests in this plugin.


jglick:
Note that this is a special loader as of #120. Should not matter for purposes of this PR I guess.

dwnusbaum (approved):

jglick:
Does not seem to have sufficed; even with this fix, I am getting file lock errors running tests on Windows (without a final `System.gc()`), and on Linux I can see some JAR files in `lsof`. A heap dump shows no root references for `JarFile`s, and anyway they have set `closeRequested`. Going to check if https://file-leak-detector.kohsuke.org/ helps.

jglick:
https://github.com/jenkinsci/lib-file-leak-detector/pull/132 :eyes:

PR #668, opened 2023-03-03
Better `CodeSource.location` for Groovy loaded from JARs

jglick:
Needs more comprehensive testing.


jglick:
I suspect that Groovy may have used the fake URL intentionally, to avoid problems with JAR URL caching. https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/57#issuecomment-1454156835

PR #669, opened 2023-03-03
Flush overall Pipeline log stream when shutting down Jenkins

dwnusbaum:
Right now, when Jenkins shuts down gracefully, we attempt to pause Pipeline execution and save all related state information. However, we do not flush the listener for the overall log, which means that for `LogStorage` implementations like `FileLogStorage` that write the log and log metadata to separate files, and write metadata synchronously while buffering log writes, the metadata and log file may be out of sync after a Jenkins restart.

In particular, when performing a safe restart, we write ["Pausing (Preparing for shutdown)"](https://github.com/jenkinsci/workflow-cps-plugin/blob/43ba38b62bb731c7dffccbc6e975732719f708a2/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java#L308) to the build log and pause the Pipeline. For `FileLogStorage`, if the last log line was written by a step, then we will immediately write the current offset in the log to the metadata file and then flush the metadata file. We will also write the pause message to the log file, but that write will be buffered, and may be lost. On restart, subsequent writes will begin at the log offset that we expected to start with "Pausing (Preparing for shutdown)". For `FileLogStorage`, this is fine, because the first lines after the restart are guaranteed to come from the overall log, but for more complex implementations of `LogStorage` that store additional metadata (e.g. line lengths), this desynchronization can cause problems.

TODO:
* [x] Can this be reproduced in a test?
    * I am not entirely sure why, but so far I have been unable to reproduce the issue in a test (e.g. `JenkinsSessionRule` + `doQuietDown`). I can only reproduce the issue while restarting an actual Jenkins instance using `/safeRestart`.
    * It can be reproduced using `RealJenkinsRule`, so I added a new test
 
### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -45,6 +45,7 @@
         <node.version>18.13.0</node.version>
         <npm.version>8.19.3</npm.version>
         <yarn.version>1.22.19</yarn.version>
+        <jenkins-test-harness.version>1926.v9b_323a_4da_421</jenkins-test-harness.version> <!-- Temporary, required to diagnose RJR issues due to https://github.com/jenkinsci/jenkins-test-harness/pull/544#issuecomment-1456707299 -->
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -45,7 +45,7 @@
         <node.version>18.13.0</node.version>
         <npm.version>8.19.3</npm.version>
         <yarn.version>1.22.19</yarn.version>
-        <jenkins-test-harness.version>1926.v9b_323a_4da_421</jenkins-test-harness.version> <!-- Temporary, required to diagnose RJR issues due to https://github.com/jenkinsci/jenkins-test-harness/pull/544#issuecomment-1456707299 -->
+        <jenkins.version>2.391</jenkins.version> <!-- TODO: Trying to diagnose errors with RealJenkinsRule in tests -->
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -45,7 +45,6 @@
         <node.version>18.13.0</node.version>
         <npm.version>8.19.3</npm.version>
         <yarn.version>1.22.19</yarn.version>
-        <jenkins.version>2.391</jenkins.version> <!-- TODO: Trying to diagnose errors with RealJenkinsRule in tests -->
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1694,6 +1694,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                                 LOGGER.log(Level.WARNING, null, t);
                             }
                         });
+                        cpsExec.getOwner().getListener().getLogger().close();
                     }
                 } catch (Exception ex) {
                     LOGGER.log(Level.WARNING, "Error persisting Pipeline execution at shutdown: "+((CpsFlowExecution) execution).owner, ex);
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionRJRTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionRJRTest.java
@@ -1,8 +1,10 @@
 package org.jenkinsci.plugins.workflow.cps;
 
+import hudson.Functions;
 import java.util.logging.Level;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.junit.Assume;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.JenkinsRule;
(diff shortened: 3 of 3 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1694,6 +1694,7 @@ public void pause(final boolean v) throws IOException {
                                 LOGGER.log(Level.WARNING, null, t);
                             }
                         });
+                        cpsExec.getOwner().getListener().getLogger().close();
```
dwnusbaum, line 1697: We could also just `flush`, but I'm not really sure that it matters either way.
```suggestion
                        cpsExec.getOwner().getListener().getLogger().flush();
```

jglick:
If `JenkinsSessionRule` does not work, try `RealJenkinsRule`.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick, line 1697: Is `getOwner()` guaranteed to be non-null? `getListener()`? In principle SpotBugs should know but sometimes it seems to miss these things.

jglick (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: I think `CpsFlowExecution.owner` is guaranteed to be non-null. It's set [here](https://github.com/jenkinsci/workflow-cps-plugin/blob/7b84f995ad630b07b7ef193e31a87dd9db85770f/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L406) and [here](https://github.com/jenkinsci/workflow-cps-plugin/blob/7b84f995ad630b07b7ef193e31a87dd9db85770f/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L721) (called [here](https://github.com/jenkinsci/workflow-job-plugin/blob/e6d865025906842fac129cab52ad00722eaeb124/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L718)).

`getListener` may return `WorkflowRun.NULL_LISTENER`, which is harmless to close, or throw an exception, but this code already handles exceptions, and I made this be the final operation in the `try` block so that other more important cleanup still runs even if this throws.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionRJRTest.java
```diff
@@ -21,4 +22,26 @@ private static void doesItSmoke(JenkinsRule r) throws Exception {
...
+        r.waitForMessage("Sleeping for 10 sec", b);
+        r.jenkins.doQuietDown();
+    }
+
+    private static void flushLogsOnShutdownPostRestart(JenkinsRule r) throws Exception {
+        WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+        WorkflowRun b = p.getLastBuild();
+        r.assertBuildStatusSuccess(r.waitForCompletion(b));
+        r.assertLogContains("Resuming build at ", b);
+        r.assertLogContains("Pausing (Preparing for shutdown)", b);
```
dwnusbaum, line 45: The first assertion here passes with or without the fix, since that message is printed after the restart. However, the second assertion is printed right before Jenkins shuts down, and it is missing from the log if you revert the fix.

dwnusbaum (commented):
> If JenkinsSessionRule does not work, try RealJenkinsRule.

Good call, `RealJenkinsRule` is indeed able to reproduce the issue. I guess JVM reuse across the restart is why this can't be reproduced using `JenkinsSessionRule` or something? The static `FileLogStorage.openStorages` field actually ends up returning the same object across the restart? EDIT: Yeah that is exactly the problem, with `JenkinsSessionRule` we reuse the instance and so we cannot lose anything from before the restart.

jglick (approved):

dwnusbaum:
Build failure is because `program.dat` couldn't be loaded on Windows after the restart: `java.io.FileNotFoundException: C:\Jenkins\workspace\ugins_workflow-cps-plugin_PR-669\plugin\target\tmp\j h14028995992276197625\jobs\p\builds\1\program.dat (The system cannot find the file specified)` Going to try a few rebuilds to see if it is consistent or not.

plugin/pom.xml
```diff
@@ -45,6 +45,7 @@
         <node.version>18.13.0</node.version>
         <npm.version>8.19.3</npm.version>
         <yarn.version>1.22.19</yarn.version>
+        <jenkins-test-harness.version>1926.v9b_323a_4da_421</jenkins-test-harness.version> <!-- Temporary, required to diagnose RJR issues due to https://github.com/jenkinsci/jenkins-test-harness/pull/544#issuecomment-1456707299 -->
```
jglick, line 48: https://github.com/jenkinsci/jenkins-test-harness/pull/544#issuecomment-1456707299

dwnusbaum:
I am merging this now because I would like to get a release out sooner than later, but I'll try to keep an eye on Dependabot PRs and remove the test assumption once we pick up https://github.com/jenkinsci/jenkins-test-harness/pull/559.

jglick:
> once we pick up [jenkinsci/jenkins-test-harness#559](https://github.com/jenkinsci/jenkins-test-harness/pull/559)

You can simply override `jenkins-test-harness.version` with a TODO comment asking to delete the override once this gets picked up via `plugin-pom`.

PR #673, opened 2023-03-06
More efficient `JarURLConnection` implementation

jglick:
https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/57#issuecomment-1457112652

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
@@ -14,6 +14,14 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.File;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URLConnection;
+import java.net.URLStreamHandler;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.codehaus.groovy.control.CompilationFailedException;
 import org.codehaus.groovy.control.CompilationUnit;
 import org.codehaus.groovy.control.CompilerConfiguration;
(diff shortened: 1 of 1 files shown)


dwnusbaum (approved):
Is it worth adding a regression test for whatever behavior you are trying to prevent?

jglick:
https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/57 effectively does that—without https://github.com/jglick/pipeline-groovy-lib-plugin/commit/b7bd7741e05e751584afd78b1dc6f47dd06d4129 you get Windows failures in one test: https://github.com/jenkinsci/pipeline-groovy-lib-plugin/runs/11808228071 which I could reliably reproduce locally on Windows 10, and also on Linux if I used `file-leak-detector` to assert that the lib JAR was not open at the end of the test. Theoretically it would be possible to write a `JenkinsRule` test here that fails (on Windows) without the hack but it would be pretty artificial since until https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/57 there would be no scenario (that I know of) in which Groovy sources are loaded from a JAR file.

https://github.com/openjdk/jdk/pull/12871 also fixes the leak, though I think this override is still potentially helpful in that it bypasses the need to ever open the JAR file to begin with in two cases used by Groovy (checking for a content encoding and checking for a last modification time).

jocwurs:
@jglick yes we are grapbing a java library which contains log4j2. Still do not fully understand what is going wrong here.

jglick:
@jocwurs looks like this is a known bug LOG4J2-3047, which has already been fixed as https://github.com/apache/logging-log4j2/pull/1164. (BTW hi @jvz!) It may suffice to `@Grab` the fixed release explicitly before the library which pulls in a broken version.

jocwurs:
@jglick thanks for the hint to this log4j bug. The library was using another log4j version than expected. It took the log4j lib coming from the analysis-model-api-plugin. And this version contains an older log4j (2.19.0).
BY shadowing log4j 2.20.0 I am now able to use the latest version of workflow-cps.
@uhafner maybe it is a good idea to update log4j used by analysis-model-api-plugin.

PR #689, opened 2023-03-27
Support PCT for multi-module Maven projects (II)

jglick:
Subsumes #685 but should allow https://github.com/jenkins-infra/repository-permissions-updater/pull/3210 to be reverted, cleaning up technical debt from #612. Local testing seems fine using both a clean working copy and a clean local repo:

```bash
mvn clean test -Dtest=CpsDefaultGroovyMethodsTest
```

PR #685, opened 2023-03-25
Support PCT for multi-module Maven projects

basil:
For the forthcoming support for multi-module Maven projects in PCT, we are planning on running at the root of each Maven multi-module project:

```
$ mvn […] clean process-test-classes  # compile test classes against original production classes
$ mvn -Djenkins.version=${LATEST_WEEKLY} -DoverrideWar=${MEGAWAR} […] hpi:resolve-test-dependencies hpi:test-hpl surefire:test  # run tests against new versions
```

The idea is to use the original production classes for the compatibility test, compiling the test classes against them for the most realistic compatibility test. This is how PCT already works today for flat Maven projects, so we are extending this to multi-module Maven projects. When run in the root of `workflow-cps`, this fails in `lib/` because compiling the test classes depends on having the main sources available, and that in turn fails with:

```
[INFO] --- exec:3.1.0:exec (default) @ groovy-cps ---
Error: Invalid or corrupt jarfile /home/basil/src/jenkinsci/workflow-cps-plugin/dgm-builder/target/classes
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:404)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:166)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:1000)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:947)
    at org.codehaus.mojo.exec.ExecMojo.execute (ExecMojo.java:471)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:342)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:330)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:213)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:175)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:76)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:163)
[...]

PR #690, opened 2023-03-28
Restore deployment of `dgm-builder`

jglick:
Reverting #689 beyond #685. Issue reproducible outside PCT as

```bash
mvn -Pquick-build package -am -pl plugin
```

since 7fc22238d80bd17031c0df7b4b94ec399bac1f20 makes Maven think this need not be built. I think there is a way to build & deploy `dgm-builder` normally without needing an überjar (the plain JAR is tiny), but it would take some more time to prepare those changes and it would be easier to experiment with that once the dust has settled on PCT.

Likely needs https://github.com/jenkins-infra/repository-permissions-updater/pull/3219 to build.

### DIFF
--- a/dgm-builder/pom.xml
+++ b/dgm-builder/pom.xml
@@ -21,11 +21,6 @@
         </license>
     </licenses>
 
-    <properties>
-        <maven.install.skip>true</maven.install.skip>
-        <maven.deploy.skip>true</maven.deploy.skip>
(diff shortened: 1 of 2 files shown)


jglick:
https://github.com/jenkinsci/workflow-cps-plugin/releases/tag/3653.v07ea_433c90b_4

PR #710, opened 2023-05-30
Backports to `PersistenceProblemsTest` from `CpsPersistenceTest` edits

jglick:
Tried to capture noteworthy edits I could find in the forked code deleted in https://github.com/jenkinsci/workflow-job-plugin/pull/361.


jglick:
@car-roll this was actually just an aside; #709 is more important.

PR #718, opened 2023-05-31
Drop support for ancient web browsers

basil:
Core dropped support for older browsers some time ago, but this plugin was never updated to match. See inline review for a detailed explanation of changes. To test this PR I did a full build and then successfully edited Pipeline configuration in the UI.

### DIFF
--- a/plugin/.eslintrc.js
+++ b/plugin/.eslintrc.js
@@ -5,31 +5,16 @@ module.exports = {
     root: true,
     env: {
         browser: true,
-        es6: true
+        es2022: true
     },
     // Uses eslint default ruleset
     extends: "eslint:recommended",
     parserOptions: {
-        ecmaVersion: 2018,
+        ecmaVersion: 2022,
         sourceType: "module"
     },
     rules: {
-    },
-    globals: {
-        Atomics: "readonly",
-        SharedArrayBuffer: "readonly",
-
-        '__dirname': false,
-
-        // Allow jest globals used in tests
-        jest: false,
-        expect: false,
-        it: false,
-        describe: false,
-        beforeEach: false,
-        afterEach: false,
-        beforeAll: false,
-        afterAll: false,
+        curly: "error",
     }
   }; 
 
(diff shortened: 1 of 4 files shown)


plugin/.eslintrc.js
```diff
@@ -5,31 +5,16 @@ module.exports = {
...
     env: {
         browser: true,
-        es6: true
+        es2022: true
     },
     // Uses eslint default ruleset
     extends: "eslint:recommended",
     parserOptions: {
-        ecmaVersion: 2018,
+        ecmaVersion: 2022,
```
basil, line 13: Matching https://github.com/jenkinsci/jenkins/blob/d2bc58d83f51a144c6ceb59a5336a3444386621b/war/.eslintrc.js#L11.

plugin/package.json
```diff
@@ -43,11 +43,10 @@
   "dependencies": {
     "ace-builds": "^1.15.0",
     "jquery": "^3.6.1",
-    "jquery-ui": "^1.13.2",
-    "raf": "^3.4.1"
```
basil, line 47: This was only needed before to provide `window.requestAnimationFrame` support to browsers we no longer support: https://caniuse.com/requestanimationframe

plugin/package.json
```diff
@@ -43,11 +43,10 @@
...
+    "jquery-ui": "^1.13.2"
   },
   "browserslist": [
     "defaults",
-    "IE 11"
+    "not IE 11"
```
basil, line 50: Matching https://github.com/jenkinsci/jenkins/blob/d2bc58d83f51a144c6ceb59a5336a3444386621b/war/package.json#L63. See also https://www.jenkins.io/doc/administration/requirements/web-browsers/

plugin/src/main/js/workflow-editor.js
```diff
@@ -1,6 +1,4 @@
 import $ from 'jquery';
-// Polyfill for window.requestAnimationFrame
-import 'raf/polyfill';
```
basil, line 3: This was only needed before to provide `window.requestAnimationFrame` support to browsers we no longer support: https://caniuse.com/requestanimationframe

jglick (approved):

PR #752, opened 2023-07-19
Update terminology

zbynek:
Just updates wording of help -- no testing done. There are some places where outdated terms are used as variable/private class names but those are probably better kept for consistency 🤷

### DIFF
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
@@ -2,6 +2,6 @@
     Specify where to obtain a source code repository containing your Groovy script.
     It will be checked out somewhere on the Jenkins controller and used to load your Pipeline script.
     (If you wish to use other files from the same repository during your Pipeline,
-    you will need to check them out separately on some node;
+    you will need to check them out separately on some agent;
     this checkout cannot be reused.)
 </div>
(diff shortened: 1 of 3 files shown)


plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
```diff
@@ -1,7 +1,7 @@
 <div>
     Specify where to obtain a source code repository containing your Groovy script.
-    It will be checked out somewhere on the Jenkins master and used to load your Pipeline script.
+    It will be checked out somewhere on the Jenkins controller and used to load your Pipeline script.
     (If you wish to use other files from the same repository during your Pipeline,
-    you will need to check them out separately on some slave;
+    you will need to check them out separately on some node;
```
jtnord, line 5: should be Node.

```suggestion
    you will need to check them out separately on some agent;
```

jtnord (commented):
Thanks @zbynek

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition/help-scm.html
zbynek: here I thought you need to do the separate checkout even if your workspace is in the built-in node on the controller machine 🤷 But that's a discouraged usecase anyway.

jtnord (approved):

PR #756, opened 2023-07-27
Remove unnecessary workarounds

basil:
Drops support for OpenJDK versions older than 11.0.17 on the 11 LTS line and 17.0.4 on the 17 LTS line, both now about a year old. The impact when running without these workarounds on older OpenJDK versions is that you may be subject to memory leaks. It seems worth it to simplify this hard-to-maintain reflection-based code and drop support for these older runtime environments. See also https://github.com/jenkinsci/jenkins/pull/8193.


jglick (approved):

jglick:
I think https://github.com/jenkinsci/script-security-plugin/pull/420 can also be cleaned up now; will file it if you were not already planning to do so.

basil:
Yup, that was next on my list: https://github.com/jenkinsci/script-security-plugin/pull/517

PR #757, opened 2023-07-28
Assertion error from `CpsThread.getStackTrace`

jglick:
Have seen a test flake in PCT:

```
java.lang.AssertionError
	at org.jenkinsci.plugins.workflow.cps.CpsThread.getStackTrace(CpsThread.java:330)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadDump$ThreadInfo.<init>(CpsThreadDump.java:68)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadDump.from(CpsThreadDump.java:150)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.getThreadDump(CpsThreadGroup.java:515)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.getThreadDump(CpsFlowExecution.java:1090)
	at org.jenkinsci.plugins.workflow.steps.GeneralNonBlockingStepExecutionTest.getStatus(GeneralNonBlockingStepExecutionTest.java:79)
```

https://github.com/jenkinsci/workflow-step-api-plugin/blob/6ecacd8c04aa1326eab48553730541b0abc9c1f9/src/test/java/org/jenkinsci/plugins/workflow/steps/GeneralNonBlockingStepExecutionTest.java#L69 looks inherently flaky, but anyway #368 seems to have introduced an unnecessary assertion error: this is a diagnostic method so it should behave conservatively.

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
@@ -327,8 +327,11 @@ public final class CpsThread implements Serializable {
     }
 
     public List<StackTraceElement> getStackTrace() {
-        assert program != null; // Otherwise this CpsThread is not even part of the CpsThreadGroup, so how is it being accessed?
-        return program.getStackTrace();
+        Continuable p = program;
+        if (p == null) {
+            return List.of(new StackTraceElement("not", "running", null, -1));
+        }
+        return p.getStackTrace();
     }
 
     private static final Logger LOGGER = Logger.getLogger(CpsThread.class.getName());


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThread.java
```diff
@@ -327,8 +327,11 @@ public void stop(Throwable t) {
     }
 
     public List<StackTraceElement> getStackTrace() {
-        assert program != null; // Otherwise this CpsThread is not even part of the CpsThreadGroup, so how is it being accessed?
```
dwnusbaum, line 330: Looks like the answer is that this can happen if the thread group is modified concurrently since `CpsFlowExecution.getThreadDump` does not use `runInCpsVmThread` and the lower-level methods do not do anything to synchronize the thread dump creation operation with respect to thread group modifications.

dwnusbaum (approved):

PR #768, opened 2023-08-17
Switch from `RUN_SCRIPTS` to `ADMINISTER`

jglick:
A user reported not seeing the option for **Serialized program state** in a virtual thread dump. They are running `3713.vd671d4321509` (prior to #755). I suspect that they may have had `ADMINISTER` without `RUN_SCRIPTS`. Since this permission is deprecated, I think we just need to check the new permission.


car-roll (approved):

dwnusbaum (approved):

daniel-beck (approved):
🤷 Makes sense after 2.222.x.

Although I'm confused how someone can have ADMINISTER without RUN_SCRIPTS, that should be very unusual and cause for review of their setup.

PR #769, opened 2023-08-19
Support system themes for Pipeline editor

timja:
Amends https://github.com/jenkinsci/workflow-cps-plugin/pull/654

see PR description in https://github.com/jenkinsci/theme-manager-plugin/pull/171 for details on the change

this has no UI changes other than dark system theme is now supported as well as dark theme

### Testing done

Tested with:
- No theme set
- Default theme explicitly set
- Dark
- Dark (system) - dark
- Dark (system) - light
- Live switching between dark and light on the system theme and seeing it automatically adapting

```[tasklist]
### Submitter checklist
```


plugin/pom.xml
```diff
@@ -191,17 +191,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
-            <groupId>io.jenkins.plugins</groupId>
-            <artifactId>theme-manager</artifactId>
```
timja, line 196: only needed as a JavaScript API now so not needed to be in pom. We could leave it in bumping the version to ensure that the plugins work together but I don't think needed

plugin/package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "webpack --config webpack.dev.js",
     "prod": "webpack --config webpack.prod.js",
-    "start": "yarn dev -- --watch",
+    "start": "yarn dev --watch",
```
timja, line 11: deprecation warning in my build console that yarn hasn't needed this since yarn v1.0

jglick (approved):
Does not appear to break anything when run on `master` without `theme-manager`.

PR #780, opened 2023-08-31
[JENKINS-71692] Attempt to stop step executions if the CPS VM thread dies with a fatal error

dwnusbaum:
See [JENKINS-71692](https://issues.jenkins.io/browse/JENKINS-71692).

If there is a fatal error in the CPS VM thread, currently running steps should be given a chance to stop before we kill the build, otherwise async steps will just keep running in the background. Things can get very messy since the execution and run associated with the `StepExecution` will already be dead, and some steps like `node` and `stage` will just run forever if we do not stop them.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -81,7 +81,7 @@ class CpsVmExecutorService extends InterceptingExecutorService {
             }
         }
         cpsThreadGroup.getExecution().croak(t);
-        // cpsThreadGroup.run() must not execute again. We shut it down after stopping steps and completing the build
+        // cpsThreadGroup.run() must not execute again. We shut this executor service down after stopping steps and completing the build rather than before
         // to avoid RejectedExecutionExceptions inside of StepExecution.stop above as the steps try to call
         // StepContext.onFailure which eventually submits a task to this executor service to trigger CpsThreadGroup.run.
         shutdown();
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -11,8 +11,11 @@ import java.util.concurrent.ExecutorService;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
+import hudson.model.TaskListener;
 import jenkins.model.Jenkins;
 import jenkins.util.InterceptingExecutorService;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
 /**
  * {@link ExecutorService} for running CPS VM.
@@ -53,8 +56,34 @@ class CpsVmExecutorService extends InterceptingExecutorService {
      * That makes it worth reporting.
      */
     private void reportProblem(Throwable t) {
+        if (isShutdown()) {
+            // We probably already got here once with the actual root cause.
+            LOGGER.log(Level.FINE, t, () -> "Unexpected exception in CPS VM thread: " + cpsThreadGroup.getExecution());
+            return;
+        }
         LOGGER.log(Level.WARNING, "Unexpected exception in CPS VM thread: " + cpsThreadGroup.getExecution(), t);
+        try {
+            // Give steps a chance to clean up.
+            for (CpsThread thread : cpsThreadGroup.getThreads()) {
+                StepExecution se = thread.getStep();
+                if (se != null) {
+                    try {
+                        se.stop(t);
+                        TaskListener listener = se.getContext().get(TaskListener.class);
+                        FlowNode node = se.getContext().get(FlowNode.class);
+                        if (listener != null && node != null) {
+                            listener.getLogger().println("Terminating " + node.getDisplayFunctionName() + " (id: " + node.getId() + ")");
+                        }
+                    } catch (Exception e) {
+                        t.addSuppressed(e);
+                    }
+                }
+            }
+        } catch (Throwable t2) {
+            t.addSuppressed(t2);
+        }
         cpsThreadGroup.getExecution().croak(t);
+        shutdown(); // cpsThreadGroup.run() must not execute again.
     }
 
     @Override
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
@@ -62,25 +62,21 @@ class CpsVmExecutorService extends InterceptingExecutorService {
             return;
         }
         LOGGER.log(Level.WARNING, "Unexpected exception in CPS VM thread: " + cpsThreadGroup.getExecution(), t);
-        try {
-            // Give steps a chance to clean up.
-            for (CpsThread thread : cpsThreadGroup.getThreads()) {
-                StepExecution se = thread.getStep();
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -53,8 +56,34 @@ public void run() {
...
+                            listener.getLogger().println("Terminating " + node.getDisplayFunctionName() + " (id: " + node.getId() + ")");
+                        }
+                    } catch (Exception e) {
+                        t.addSuppressed(e);
+                    }
+                }
+            }
+        } catch (Throwable t2) {
+            t.addSuppressed(t2);
+        }
```
dwnusbaum, line 84: I think if we switch the inner catch to `Throwable` this should be unreachable and could be deleted.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -664,4 +667,47 @@ public boolean isAllowed(String groovyResourceUrl) {
...
+                "          semaphore('croak')\n" +
+                "          break label\n" + // CallEnv.getBreakAddress throws an IllegalStateException when this runs.
+                "        }\n" +
+                "      }\n" +
+                "    }\n" +
+                "  },\n" +
+                "  'keepsRunning': {\n" +
+                "    semaphore('other')\n" +
+                "    echo 'kept running!'\n" +
+                "  }\n" +
```
dwnusbaum, line 689: Not necessary to reproduce the main problem with `node`, but this shows how steps could still "run" even after `CpsFlowExecution.croak`.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
dwnusbaum: Simplified in [e76ebb6](https://github.com/jenkinsci/workflow-cps-plugin/pull/780/commits/e76ebb65eb3178d6a21599ce23d72ac616dec9d8).

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -664,4 +670,117 @@ public boolean isAllowed(String groovyResourceUrl) {
...
+            // It's difficult to add meaningful non-flaky test assertions here.
+            // Even before the associated fix, 'kept running' wasn't printed, because although CpsThreadGroup.run
+            // did execute again, it hit an IllegalStateException in SandboxContinuable, which resulted in a second
+            // call to CpsFlowExecution.croak.
+            assertTrue(((CpsFlowExecution) b.getExecution()).programPromise.get(1, TimeUnit.SECONDS).runner.isShutdown());
+        });
+    }
+
+    @Issue("JENKINS-71692")
+    @Test public void stepsAreStoppedWhenBodyExecutionCallbackThrows() throws Throwable {
```
dwnusbaum, line 717: Mimics the case described in [this comment](https://issues.jenkins.io/browse/JENKINS-70267?focusedId=434633&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-434633), although the callback was also fixed upstream, see https://github.com/jenkinsci/gitlab-plugin/pull/1410.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -53,8 +56,30 @@ public void run() {
      * That makes it worth reporting.
      */
     private void reportProblem(Throwable t) {
+        if (isShutdown()) {
```
dwnusbaum, line 59: Or maybe better to check this instead:
```suggestion
        if (cpsThreadGroup.getExecution().isComplete()) {
```

car-roll: Not sure about the micro specifics between the two, but using `getExecution` at least looks more readable to me.

car-roll (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
dwnusbaum: See [b9080e8](https://github.com/jenkinsci/workflow-cps-plugin/pull/780/commits/b9080e88143ad3eb6cdb6e57ad5fdd91abab9ce4).

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -53,8 +56,35 @@ public void run() {
...
+                    if (listener != null && node != null) {
+                        listener.getLogger().println("Terminating " + node.getDisplayFunctionName() + " (id: " + node.getId() + ")");
+                    }
+                } catch (Throwable e) {
+                    t.addSuppressed(e);
+                }
+            }
+        }
         cpsThreadGroup.getExecution().croak(t);
+        // cpsThreadGroup.run() must not execute again. We shut it down after stopping steps and completing the build
```
dwnusbaum, line 84:
```suggestion
        // cpsThreadGroup.run() must not execute again. We shut this executor service down after stopping steps and completing the build
```

car-roll (approved):

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
dwnusbaum: Yes I also tested changes along those lines, but to start with I just wanted to show that the new fix is somewhat generic and covers various kinds of fatal VM thread errors. I will add the special case now though.

dwnusbaum: Updated to match `TailCall` behavior in [a27f32f](https://github.com/jenkinsci/workflow-cps-plugin/pull/780/commits/a27f32fd7a35f3e847e10a429cc7c15c723df26f).

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -664,4 +670,117 @@ public boolean isAllowed(String groovyResourceUrl) {
...
+        sessions.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition(
+                "parallel(\n" +
+                "  'willCroak': {\n" +
+                "    node {\n" +
+                "      label: {\n" + // Defines a label on a BlockStatement, which is erroneously ignored by CpsTransformer.
+                "        while (true) {\n" +
+                "          semaphore('croak')\n" +
+                "          break label\n" + // CallEnv.getBreakAddress throws an IllegalStateException when this runs.
```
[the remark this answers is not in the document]
dwnusbaum: Yeah, let me see if we can use `CpsFlowExecution.runInCpsVmThread` to exercise the same general behavior in a way that we will never fix.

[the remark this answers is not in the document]
dwnusbaum: Test updated to use `runInCpsVmThread` in [0354b48](https://github.com/jenkinsci/workflow-cps-plugin/pull/780/commits/0354b48597a469e20fd027392a10cdaec0c721a2), and I manually checked that without the fix the updated test triggers the same behavior as with the label bug.

jglick (approved):

PR #788, opened 2023-09-11
Be quieter handling `CpsFlowExecution.owner == null` in `suspendAll`

jglick:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/669#discussion_r1125056205. I did in fact observe

```
WARNING	o.j.p.w.cps.CpsFlowExecution#suspendAll: Error persisting Pipeline execution at shutdown: null
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.suspendAll(CpsFlowExecution.java:1674)
	at …
```

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1638,25 +1638,23 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     @Restricted(DoNotUse.class)
     @Terminator(attains = FlowExecutionList.EXECUTIONS_SUSPENDED)
     public static void suspendAll() {
-        CpsFlowExecution exec = null;
         try (Timeout t = Timeout.limit(3, TimeUnit.MINUTES)) { // TODO some complicated sequence of calls to Futures could allow all of them to run in parallel
             LOGGER.fine("starting to suspend all executions");
             for (FlowExecution execution : FlowExecutionList.get()) {
-                try {
-                    if (execution instanceof CpsFlowExecution) {
-                        CpsFlowExecution cpsExec = (CpsFlowExecution)execution;
+                if (execution instanceof CpsFlowExecution) {
+                    CpsFlowExecution cpsExec = (CpsFlowExecution) execution;
+                    try {
                         cpsExec.checkAndAbortNonresumableBuild();
 
                         LOGGER.log(Level.FINE, "waiting to suspend {0}", execution);
-                        exec = (CpsFlowExecution) execution;
                         // Like waitForSuspension but with a timeout:
-                        if (exec.programPromise != null) {
+                        if (cpsExec.programPromise != null) {
                             LOGGER.log(Level.FINER, "Waiting for Pipeline to go to sleep for shutdown: "+execution);
                             try {
-                                exec.programPromise.get(1, TimeUnit.MINUTES).scheduleRun().get(1, TimeUnit.MINUTES);
+                                cpsExec.programPromise.get(1, TimeUnit.MINUTES).scheduleRun().get(1, TimeUnit.MINUTES);
                                 LOGGER.log(Level.FINER, " Pipeline went to sleep OK: "+execution);
                             } catch (InterruptedException | TimeoutException ex) {
-                                LOGGER.log(Level.WARNING, "Error waiting for Pipeline to suspend: "+exec, ex);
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1638,25 +1638,23 @@ public void pause(final boolean v) throws IOException {
     @Restricted(DoNotUse.class)
     @Terminator(attains = FlowExecutionList.EXECUTIONS_SUSPENDED)
     public static void suspendAll() {
-        CpsFlowExecution exec = null;
```
jglick, line 1641: Outermost `catch` clause seemed redundant. (`Timeout.close` does not throw exceptions, so it was not for that.)

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1638,25 +1638,23 @@ public void pause(final boolean v) throws IOException {
...
+                if (execution instanceof CpsFlowExecution) {
+                    CpsFlowExecution cpsExec = (CpsFlowExecution) execution;
+                    try {
                         cpsExec.checkAndAbortNonresumableBuild();
 
                         LOGGER.log(Level.FINE, "waiting to suspend {0}", execution);
-                        exec = (CpsFlowExecution) execution;
                         // Like waitForSuspension but with a timeout:
-                        if (exec.programPromise != null) {
+                        if (cpsExec.programPromise != null) {
```
jglick, line 1651: simplifying

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1671,15 +1669,15 @@ public static void suspendAll() {
                                 }
                             });
                         }
-                        cpsExec.getOwner().getListener().getLogger().close();
+                        if (cpsExec.owner != null) {
```
jglick, line 1672: main fix

dwnusbaum: Any idea how this would be reachable? Corruption during resumption or something? Perhaps https://github.com/jenkinsci/workflow-api-plugin/pull/304 has made it possible? If I understand right, in this case `FlowExecutionList.runningTasks` contains a `FlowExecutionOwner` which successfully returns a `FlowExecution` from `.get()`, but for which `FlowExecution.owner` on the returned object is null, which seems problematic and unexpected.

dwnusbaum (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: I am afraid I do not know—just saw this in a log.

jglick: I suppose `[Workflow]Run.reload` deserializes `execution`, but then `WorkflowRun.onLoad` fails before calling `getExecution`, or `getExecution` fails before calling `fetchedExecution.onLoad(new Owner(this))`, or `unmarshal` fails partway through, etc. Presumably the build was badly corrupted somehow. The point of this PR is just to avoid unnecessary stack traces after that.

dwnusbaum: For what it's worth though I saw another case of a `CpsFlowExecution` with a `null` `owner` recently, which is why I am wondering if something has changed things:

```
java.lang.IllegalStateException: List of flow heads unset for CpsFlowExecution[null]
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.getCurrentHeads(CpsFlowExecution.java:1018)
        ... insignificant, just a user loading some build page that accessed the heads for an execution ...
```

I looked at the XML for that execution on disk and `head` was non-null and pointed to a `FlowEndNode` that did exist in `workflow/`, so it wasn't obvious what might have gone wrong.

jtnord (approved):
(how) does this relate to https://github.com/jenkinsci/workflow-cps-plugin/pull/797 ?

jglick:
@jtnord I believe it is unrelated.

dwnusbaum (approved):

PR #797, opened 2023-10-03
`RejectedExecutionException` from `CpsFlowExecution.getCurrentExecutions`

jglick:
Caught and logged as a warning from `FlowExecutionList.StepExecutionIteratorImpl` when trying to call `StepExecution.applyAll` at a certain point during shutdown, for example from `DurableTaskStep$AgentReconnectionListener.check` if an agent is disconnected. In fact that listener does check `Jenkins.get().isTerminating()` to avoid noise, but unfortunately this flag is set by `Jenkins.cleanUp` _after_ running `@Terminator`s (and before all other cleanup steps).


jtnord (commented):
Code looks fine - 4 failing tests on windows?  timout > 180sbut other PRs seem to be OK.

jglick:
Windows CI is just super slow and prone to timeouts.

jtnord (approved):

PR #805, opened 2023-10-18
refactor: Java Security Ultimate Scan 2023

caytec:
Use this link to re-run the recipe: https://app.moderne.io/recipes/builder/MtR7dypUE?organizationId=SmVua2lucyBDSQ%3D%3D

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -557,7 +557,7 @@ public final class CpsThreadGroup implements Serializable {
     @CpsVmThreadOnly
     public void saveProgram(File f) throws IOException {
         File dir = f.getParentFile();
-        File tmpFile = File.createTempFile("atomic",null, dir);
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -557,7 +557,7 @@ void saveProgram() throws IOException {
     @CpsVmThreadOnly
     public void saveProgram(File f) throws IOException {
         File dir = f.getParentFile();
-        File tmpFile = File.createTempFile("atomic",null, dir);
+        File tmpFile = Files.createTempFile(dir.toPath(), "atomic", null).toFile();
```
jtnord, line 560: This is not equivalent and will create a file with more restrictive permissions than desired.
As the file is moved later on (line 580) it will retain this restrictive permissions.
Thus I do not believe this to be a correct change.

jglick: Also, what problem does this PR purport to solve, other than simply applying some automated refactoring rule? The title implies something about “security” but if there is a vulnerability, it should be reported through proper channels, never as a public PR.

PR #807, opened 2023-10-19
Migrate `FlowNode` storage to `BulkFlowNodeStorage` upon execution completion to improve read performance

dwnusbaum:
`SimpleXStreamFlowNode` storage optimizes write performance (important when using `MAX_SURVIVABILITY`) at the cost of read performance (in particular network file systems do not deal well with having to read hundreds of tiny little `FlowNode` XML files). For complex Pipeline builds running on controllers using network file systems, loading flow nodes after a Jenkins restart can take a significant amount of time (I observed a case where it was taking 15+ seconds), which can impact performance in various ways, most obviously when opening visualizations like the "Pipeline steps" view.

This PR explores the possibility of migrating all completed executions to `BulkFlowNodeStorage` to increase read performance. We don't have to worry about write performance for completed executions, so at least in theory there are no downsides after the migration is complete.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -535,7 +535,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             return;
         }
         if (storage.delegate instanceof SimpleXStreamFlowNodeStorage) {
-            LOGGER.log(Level.FINER, () -> "Migrating " + this + " to BulkFlowNodeStorage");
+            LOGGER.log(Level.FINE, () -> "Migrating " + this + " to BulkFlowNodeStorage");
             String newStorageDir = (this.storageDir != null) ? this.storageDir + "-completed" : "workflow-completed";
             try {
                 FlowNodeStorage newStorage = new BulkFlowNodeStorage(this, new File(this.owner.getRootDir(), newStorageDir));
@@ -547,6 +547,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                     newStorage.storeNode(node, true);
                 }
                 newStorage.flush();
+                LOGGER.log(Level.FINE, () -> "Copied nodes to " + newStorageDir);
                 File oldStorageDir = getStorageDir();
                 this.storageDir = newStorageDir;
                 // TODO: A more conservative option could be to instead keep using the current storage until after
@@ -555,11 +556,12 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 this.storage = new TimingFlowNodeStorage(newStorage);
                 try {
                     Util.deleteRecursive(oldStorageDir);
+                    LOGGER.log(Level.FINE, () -> "Deleted " + oldStorageDir);
                 } catch (IOException e) {
                     LOGGER.log(Level.FINE, e, () -> "Unable to delete unused flow node storage directory " + oldStorageDir + " for " + this);
                 }
             } catch (Exception e) {
-                LOGGER.log(Level.FINE, e, () -> "Unable to migrate " + this + " to BulkFlowNodeStorage");
+                LOGGER.log(Level.WARNING, e, () -> "Unable to migrate " + this + " to BulkFlowNodeStorage");
             }
         }
     }
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -550,15 +550,17 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 LOGGER.log(Level.FINE, () -> "Copied nodes to " + newStorageDir);
                 File oldStorageDir = getStorageDir();
                 this.storageDir = newStorageDir;
-                // TODO: A more conservative option could be to instead keep using the current storage until after
-                // a restart. In that case we'd have to defer deletion of the old storage dir and handle it in
-                // initializeStorage.
-                this.storage = new TimingFlowNodeStorage(newStorage);
+                storage.readWriteLock.writeLock().lock();
                 try {
-                    Util.deleteRecursive(oldStorageDir);
-                    LOGGER.log(Level.FINE, () -> "Deleted " + oldStorageDir);
-                } catch (IOException e) {
-                    LOGGER.log(Level.FINE, e, () -> "Unable to delete unused flow node storage directory " + oldStorageDir + " for " + this);
+                    storage.delegate = newStorage;
+                    try {
+                        Util.deleteRecursive(oldStorageDir);
+                        LOGGER.log(Level.FINE, () -> "Deleted " + oldStorageDir);
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINE, e, () -> "Unable to delete unused flow node storage directory " + oldStorageDir + " for " + this);
+                    }
+                } finally {
+                    storage.readWriteLock.writeLock().unlock();
                 }
             } catch (Exception e) {
                 LOGGER.log(Level.WARNING, e, () -> "Unable to migrate " + this + " to BulkFlowNodeStorage");
@@ -1906,7 +1908,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     static final ThreadLocal<CpsFlowExecution> PROGRAM_STATE_SERIALIZATION = new ThreadLocal<>();
 
     class TimingFlowNodeStorage extends FlowNodeStorage {
-        final FlowNodeStorage delegate;
+        FlowNodeStorage delegate;
         private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
 
         TimingFlowNodeStorage(FlowNodeStorage delegate) {
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -28,8 +28,6 @@ import com.cloudbees.groovy.cps.Continuable;
 import com.cloudbees.groovy.cps.Env;
 import com.cloudbees.groovy.cps.Envs;
 import com.cloudbees.groovy.cps.Outcome;
-import com.cloudbees.groovy.cps.impl.ConstantBlock;
-import com.cloudbees.groovy.cps.impl.ThrowBlock;
 import com.cloudbees.groovy.cps.sandbox.Invoker;
 import com.cloudbees.jenkins.support.api.Component;
 import com.cloudbees.jenkins.support.api.Container;
@@ -109,6 +107,7 @@ import hudson.AbortException;
 import hudson.BulkChange;
 import hudson.Extension;
 import hudson.Util;
+import hudson.Functions;
 import hudson.init.Terminator;
 import hudson.model.Item;
 import hudson.model.Job;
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -506,13 +508,53 @@ public FlowExecutionOwner getOwner() {
...
+                    newStorage.storeNode(node, true);
+                }
+                newStorage.flush();
+                File oldStorageDir = getStorageDir();
+                synchronized (this) {
+                    this.storageDir = newStorageDir;
+                    // TODO: A more conservative option could be to instead keep using the current storage until after
+                    // a restart. In that case we'd have to defer deletion of the old storage dir and handle it in
+                    // initializeStorage.
+                    this.storage = new TimingFlowNodeStorage(newStorage);
```
dwnusbaum, line 545: Not sure if this is safe or not.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -506,13 +508,53 @@ public FlowExecutionOwner getOwner() {
...
+                DepthFirstScanner scanner = new DepthFirstScanner();
+                scanner.setup(flowEndNode);
+                // The hope is that by doing this right when the execution completes, most of the nodes will already be
+                // cached in memory, saving us the cost of having to read them all from disk.
+                for (FlowNode node : scanner) {
+                    newStorage.storeNode(node, true);
+                }
+                newStorage.flush();
+                File oldStorageDir = getStorageDir();
+                synchronized (this) {
```
dwnusbaum, line 540: Hmm, just noticed that `onProgramEnd` is already `synchronized`. The worst-case scan performance where many nodes have to be read from disk might be a problem. I guess we could trigger the migration here but run it asynchronously, and then give up on swapping `this.storage` below.

I would really like though for anything that only loads completed executions to only ever see the migrated storage.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -506,13 +508,53 @@ public FlowExecutionOwner getOwner() {
...
         return new TimingFlowNodeStorage(wrappedStorage);
     }
 
+    /**
+     * Called when the build completes to migrate from {@link SimpleXStreamFlowNodeStorage} to
+     * {@link BulkFlowNodeStorage} to improve read performance for completed builds.
+     */
+    private void optimizeStorage(FlowNode flowEndNode) {
+        if (storage.delegate instanceof SimpleXStreamFlowNodeStorage) {
+            String newStorageDir = (this.storageDir != null) ? this.storageDir + "-completed" : "workflow-completed";
```
dwnusbaum, line 528: Similar to what we do with `workflow-fallback` in `createPlaceholderNodes`, but of course other approaches could be used.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -506,13 +508,53 @@ public FlowExecutionOwner getOwner() {
...
+            try {
+                FlowNodeStorage newStorage = new BulkFlowNodeStorage(this, new File(this.owner.getRootDir(), newStorageDir));
+                DepthFirstScanner scanner = new DepthFirstScanner();
+                scanner.setup(flowEndNode);
+                // The hope is that by doing this right when the execution completes, most of the nodes will already be
+                // cached in memory, saving us the cost of having to read them all from disk.
```
dwnusbaum, line 534: It would be simpler to do the migration when first loading the execution, but then we guarantee the worst-case of having to read _all_ the individual `FlowNode` XML files just to write them again, which I think in practice would mean that you would only see performance benefits from this PR when navigating to a build after two restarts in row.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -153,6 +154,7 @@ private static InputStepExecution getInputStepExecution(WorkflowRun run, String
     final  static String DEFAULT_JOBNAME = "testJob";
 
     /** Simulates something happening badly during final shutdown, which may cause build to not appear done. */
+    @Ignore("TODO: test needs to be adapted")
```
dwnusbaum, line 157: The test tries to delete a specific `FlowNode` XML file, but with the migration that file no longer exists after the build completes.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick, line 545: Does this require a `build.xml` save? Or is that done by something else anyway?

jglick (commented):
Looks right. May be possible to write a `JenkinsSessionRule` test verifying accesses to flow nodes after build completion in the same session, and then in the next session, asserting somehow that the optimized storage is in use.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: The two places that call `onProgramEnd`  (`CpsFlowExecution.croak` and `CpsThreadGroup.run`) both call `saveOwner` right after the call, so I think things will be ok, but I will check in more detail later.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
```diff
@@ -818,4 +827,37 @@ public boolean takesImplicitBlockArgument() {
...
+            List<FlowNode> nodes = new DepthFirstScanner().allNodes(b.getExecution());
+            assertThat(nodes.stream().map(FlowNode::getDisplayFunctionName).collect(Collectors.toList()), equalTo(
+                    List.of("End of Pipeline", "echo", "semaphore", "echo", "Start of Pipeline")));
+        });
+        sessions.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            FlowNodeStorage storage = ((TimingFlowNodeStorage) ((CpsFlowExecution) b.getExecution()).getStorage()).delegate;
+            assertThat(storage, instanceOf(BulkFlowNodeStorage.class));
+            assertThat(((CpsFlowExecution) b.getExecution()).getStorageDir().toString(), endsWith("/workflow-completed"));
```
jglick, line 856: Could also verify that it has just one file; and that `/workflow` is gone.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: One change I will make is to expand the outer catch clause in `optimizeStorage` to catch `Exception` in case `DepthFirstScanner` can throw an NPE or similar runtime exceptions in unusual cases.

Other than that, I think things should be ok unless we ever see a `Throwable` that is not an `Exception` in `onProgramEnd` (e.g. `StackOverflowError`, `OutOfMemoryError`), but that is the same with or without this PR. I think we could perhaps catch `Throwable` or adapt existing callers to use `try`/`finally` to be more robust against thrown `Error`s but I am not sure that it would make a difference in practice.

dwnusbaum:
As far as performance, I did some very basic local testing (M1 Mac, 32GB RAM, SSD) by running a Pipeline that just loops over `withEnv[] { sleep 0 }` thousands of times, waiting for the build to complete, then restarting Jenkins, opening the script console and timing how long it takes to use `DepthFirstScanner` to scan all nodes. `BulkFlowNodeStorage` was between 3 and 6 times faster in my trials (roughly 0.5 seconds vs 2.5 seconds to scan all nodes).

I also added some `FINE` logging to `optimizeStorage` just to see how long the migration took in the best case where the build runs to completion the flow node cache never being invalidated, and with around 20k flow nodes in a build there was at most a 1 second gap between starting migration and finishing deletion of `workflow/` (standard logger doesn't include milliseconds so that's all the granularity I have for now). In the worst case, where you restart right before the final step completes, after the restart `optimizeStorage` has to actually load all the nodes from disk, which makes it substantially slower (roughly 2.5 seconds for 20k flow nodes for me).

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: I think there is some potential for race conditions here. Methods like `CpsFlowExecution.getNode` are not `synchronized`, so something could call `getNode` while `storage` still points to `SimpleXStreamFlowNodeStorage`, and then if that node is not in the cache, it will be read from disk, but in the meantime `optimizeStorage` could be  deleting `workflow/` after swapping `this.storage`, leading to `getNode` temporarily returning `null`, or the `workflow/` deletion failing, etc.

I can think of a few options:
1. Only delete `workflow/` in `initializeStorage` if `this.storageDir` is `workflow-completed`. Not ideal if the build never actually gets looked at again, because `workflow/` will contain redundant data, but simple.
2. Use `Cleaner` in `optimizeStorage` to register an action to delete the old storage directory once the old `this.storage.delegate` is phantom reachable instead of doing it immediately. Should work fine in the happy path, but if Jenkins crashes or something `workflow/` might not get cleaned up.
3. Add some kind of locking around `this.storage`. Has potential for bugs and general performance issues. Perhaps a `ReadWriteLock` would perform well enough.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: I think we keep this kind of approach. I marked the new method as `synchronized` just for clarity.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: The third option seems most straightforward.

dwnusbaum: Yes I thought so as well, but it got quite messy. See 4b89fed3df71d639e4f7e3959d9ff26b1f488d1c. It is awkward because although many methods that consume `storage` probably cannot ever run concurrently with `onProgramEnd` and so do not need any locking, it is not immediately obvious and so I am not really comfortable with skipping the locks selectively. Adding a new lock also introduces the possibility of deadlock if something attempts to acquire the monitor for `CpsFlowExecution` while holding `storageLock` and the build completes concurrently (nothing does this today as far as I can see).

The `Cleaner` approach is quite a bit simpler in terms of code changes, for what that's worth. Perhaps we could combine that approach with deletion of `workflow/` in `initializeStorage` to guarantee deletion in most cases without having to introduce the new lock and change all associated code.

dwnusbaum: Hmm, maybe actually I can reuse the lock inside of `TimingFlowNodeStorage` instead to simplify things.

dwnusbaum: Yes I think [62119b8](https://github.com/jenkinsci/workflow-cps-plugin/pull/807/commits/62119b8e009ee491f64f008e56ea85bbd9c743fd) is safe and it is far simpler than introducing another lock. I wish I had noticed it yesterday...

dwnusbaum:
One thing I noticed while looking into all of this is that [`BulkFlowNodeStorage.getOrLoadNodes`](https://github.com/jenkinsci/workflow-support-plugin/blob/f4d2878f75e3d7e9ed495405b6b49a62cb0386d9/src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java#L84) has no form of internal synchronization, and may be called by indirectly by `TimingFlowNodeStorage` methods that only acquire a read lock, so I think in theory after a restart many threads could all load the same file concurrently. As far as I can tell the only problem would be wasted effort though.

jglick (approved):

basil:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertNotNull(Assert.java:713)
	at org.junit.Assert.assertNotNull(Assert.java:723)
	at com.cloudbees.jenkins.support.actions.SupportAbstractItemActionTest.generatePipelineBundleDefaultsAndCheckContent(SupportAbstractItemActionTest.java:119)
```

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at com.cloudbees.jenkins.support.impl.AbstractItemComponentTest.addContentsFromPipeline(AbstractItemComponentTest.java:167)
```

jglick:
Looks like we just need assertion tweaks in https://github.com/jenkinsci/support-core-plugin/blob/fb93051e5ad707a27ec61ebdbca23b8e66f76358/src/test/java/com/cloudbees/jenkins/support/impl/AbstractItemComponentTest.java#L171 etc.; @dwnusbaum are you on it already?

dwnusbaum:
Yes

dwnusbaum:
https://github.com/jenkinsci/support-core-plugin/pull/494

PR #812, opened 2023-11-03
Test coverage for `evaluate`ing classes across restarts

jglick:
An unsuccessful attempt to reproduce a bug observed in a build that failed to resume after a restart with a message like `Script3: 8: unable to resolve class SomeHelper` where there were multiple helper classes loaded using the `evaluate(readTrusted(SomeHelper.groovy))` idiom. (Tried moving the `semaphore` to various places; test still passes.) Anyway, it appears there was no prior test coverage for `evaluate` particularly loading new types and those types being persisted across controller restarts, so it is useful to add some.

### DIFF
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
@@ -692,6 +692,26 @@ public class CpsFlowExecutionTest {
         }
     }
 
+    @Test public void evaluateAfterRestart() throws Throwable {
+        sessions.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition(
+                "def x = evaluate('class X {X() {}; def m1() {/OK/}}; new X()')\n" +
+                "def y = evaluate('class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()')\n" +
+                "semaphore('wait')\n" +
+                "y.x = x\n" +
+                "echo(/received ${y.m2()}/)\n", true));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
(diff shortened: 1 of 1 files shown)


jglick:
I was curious how this worked, actually;

```diff
diff --git plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
index d32acba6..f027c9bb 100644
--- plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
+++ plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
@@ -707,6 +707,7 @@ public class CpsFlowExecutionTest {
         sessions.then(r -> {
             WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
             WorkflowRun b = p.getLastBuild();
+            FileUtils.copyFile(new File(b.getRootDir(), "build.xml"), System.out);
             SemaphoreStep.success("wait/1", null);
             r.assertLogContains("received really OK", r.assertBuildStatus(Result.SUCCESS, r.waitForCompletion(b)));
         });
```

shows that these scripts _are_ persisted naturally:

```xml
…
    <script>def x = evaluate(&apos;class X {X() {}; def m1() {/OK/}}; new X()&apos;)
def y = evaluate(&apos;class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()&apos;)
semaphore(&apos;wait&apos;)
y.x = x
echo(/received ${y.m2()}/)
</script>
    <loadedScripts class="map">
      <entry>
        <string>Script1</string>
        <string>class X {X() {}; def m1() {/OK/}}; new X()</string>
      </entry>
      <entry>
        <string>Script2</string>
        <string>class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()</string>
      </entry>
    </loadedScripts>
…
```

jtnord (approved):

PR #821, opened 2023-12-04
Suppress ESLint warning `no-unused-vars`

jglick:
This code introduced in #769 is breaking the local build. ~Not sure why it did not at the time; maybe there is some dependency not being pinned properly?~ As per https://github.com/jenkinsci/plugin-pom/issues/874 this is not caught by CI.

### DIFF
--- a/plugin/src/main/js/workflow-editor.js
+++ b/plugin/src/main/js/workflow-editor.js
@@ -54,6 +54,7 @@ $(function() {
                         setTheme(editor);
 
                         if (window.isSystemRespectingTheme) {
+                            // eslint-disable-next-line no-unused-vars
                             window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', event => {
                                 setTheme(editor)
                             });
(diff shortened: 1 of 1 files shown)


plugin/src/main/js/workflow-editor.js
```diff
@@ -54,6 +54,7 @@ $(function() {
                         setTheme(editor);
 
                         if (window.isSystemRespectingTheme) {
+                            // eslint-disable-next-line no-unused-vars
                             window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', event => {
```
basil, line 58: Or just clean up now. Maybe the unused variable should just be removed?

```suggestion
                            window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', () => {
```

jglick: Sure, I was just unsure whether `addEventListener` would accept a no-arg function. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback seems so.

timja (approved):

PR #538, opened 2022-05-17
Remove `default-allowlist` entries for Groovy source files related to Declarative

dwnusbaum:
In 76a7681702f42d65f77bbaa5463f146876ea62db, we added an allowlist mechanism for loading arbitrary Groovy source files from the classpath. This included a `default-allowlist` file with entries related to various plugins that relied on the previous behavior (typically to implement some kind of DSL). Now that the security fix is out, we can update Declarative to use the new API and remove a lot of the entries in `default-allowlist`.

We will also need to update `docker-workflow` because of `/org/jenkinsci/plugins/docker/workflow/declarative/AbstractDockerPipelineScript.groovy`, but `kubernetes` and `amazon-ecs` will be fixed indirectly via Declarative without requiring any code changes.

Even once the downstream fixes are merged and released, we should wait a while to merge and release this to avoid unnecessary upgrade complexity for users, so I plan on leaving this in draft state for a few weeks even after downstream changes get released.

Downstream PRs:
* https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/529 (released as [2.2097.v33db_b_de764b_e](https://github.com/jenkinsci/pipeline-model-definition-plugin/releases/tag/2.2097.v33db_b_de764b_e))
* https://github.com/jenkinsci/docker-workflow-plugin/pull/256 (released as [1.29](https://github.com/jenkinsci/docker-workflow-plugin/releases/tag/docker-workflow-1.29))

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #828, opened 2024-01-09
Revert "Merge pull request #538 from dwnusbaum/post-SECURITY-359"

dwnusbaum:
Reverts #538.

See https://github.com/jenkinsci/workflow-cps-plugin/pull/538#issuecomment-1883042255.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
@@ -1,4 +1,34 @@
 # This list is ordered from most popular to least popular plugin to minimize performance impact.
+# pipeline-model-definition
+/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/AnyScript.groovy
+/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
+/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/NoneScript.groovy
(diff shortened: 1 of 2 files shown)


plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
```diff
@@ -1,4 +1,34 @@
...
+/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TriggeredByConditionalScript.groovy
+# pipeline-model-extensions
+/org/jenkinsci/plugins/pipeline/modeldefinition/agent/CheckoutScript.groovy
+# docker-workflow
+/org/jenkinsci/plugins/docker/workflow/Docker.groovy
+/org/jenkinsci/plugins/docker/workflow/declarative/AbstractDockerPipelineScript.groovy
+/org/jenkinsci/plugins/docker/workflow/declarative/DockerPipelineFromDockerfileScript.groovy
+/org/jenkinsci/plugins/docker/workflow/declarative/DockerPipelineScript.groovy
+# kubernetes
+/org/csanchez/jenkins/plugins/kubernetes/pipeline/KubernetesDeclarativeAgentScript.groovy
```
dwnusbaum, line 29: FWIW although this was the only reported problem, and the stack trace at least shows that `ModelInterpreter.groovy` was allowed correctly after the change, almost all of the files here that end with `Script.groovy` all use the same mechanism, and I don't really have any hypothesis as to why some of them would be affected but not others, so it seems safest to just revert the whole change. `WithScriptDescriptor` lives in `pipeline-model-extensions`, so theoretically I would expect scripts in `pipeline-model-definition` and the other plugins to have the same problems as the script in `kubernetes`.

jtnord (approved):

jglick (approved):
Seems best as a hotfix while investigations proceed.

Vlatombe (approved):

PR #829, opened 2024-01-09
Cache `GroovySourceFileAllowlist` responses and remove entries from the default allowlist that are unrelated to `WithScriptDescriptor.WithScriptAllowlist`

dwnusbaum:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/538#issuecomment-1883442531.

`GroovySourceFileAllowlist` can be consulted many times, but the allowed files are effectively static and just depend on which plugins are installed. Once a file has been allowed once, we don't need to check all of the allowlists again.

### Testing done

```[tasklist]
### Submitter checklist
```


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
```diff
@@ -152,6 +148,22 @@ private static boolean endsWithIgnoreCase(String value, String suffix) {
         }
     }
 
+    @Extension
+    public static class AllowedGroovyResourcesCache {
+        private final Map<String, Boolean> cache = new ConcurrentHashMap<>();
```
dwnusbaum, line 153: At most there would be around 50 entries in the map if you used every single plugin that has Groovy source files, so I don't think we need anything fancy.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
```diff
@@ -1,10 +1,8 @@
 # This list is ordered from most popular to least popular plugin to minimize performance impact.
 # pipeline-model-definition
-/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/AnyScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/NoneScript.groovy
-/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AbstractChangelogConditionalScript.groovy
```
dwnusbaum, line 7: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/f921b4e72c73e4b60564b8f6610f35bf8bbe9720/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeLogConditional.java#L100-L101

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
```diff
@@ -18,11 +16,7 @@
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/NotConditionalScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TagConditionalScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TriggeredByConditionalScript.groovy
-# pipeline-model-extensions
-/org/jenkinsci/plugins/pipeline/modeldefinition/agent/CheckoutScript.groovy
```
dwnusbaum, line 22: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/f921b4e72c73e4b60564b8f6610f35bf8bbe9720/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java#L110-L111

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
```diff
@@ -1,10 +1,8 @@
 # This list is ordered from most popular to least popular plugin to minimize performance impact.
 # pipeline-model-definition
-/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```
dwnusbaum, line 3: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/f921b4e72c73e4b60564b8f6610f35bf8bbe9720/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java#L77-L78

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
```diff
@@ -18,11 +16,7 @@
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/NotConditionalScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TagConditionalScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TriggeredByConditionalScript.groovy
-# pipeline-model-extensions
-/org/jenkinsci/plugins/pipeline/modeldefinition/agent/CheckoutScript.groovy
 # docker-workflow
-/org/jenkinsci/plugins/docker/workflow/Docker.groovy
-/org/jenkinsci/plugins/docker/workflow/declarative/AbstractDockerPipelineScript.groovy
```
dwnusbaum, line 25: https://github.com/jenkinsci/docker-workflow-plugin/blob/dc3cf1327f1b8c9967a56aa0e1a1f3f0ff79b57c/src/main/java/org/jenkinsci/plugins/docker/workflow/declarative/AbstractDockerAgent.java#L133-L134

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist/default-allowlist
```diff
@@ -18,11 +16,7 @@
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/NotConditionalScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TagConditionalScript.groovy
 /org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TriggeredByConditionalScript.groovy
-# pipeline-model-extensions
-/org/jenkinsci/plugins/pipeline/modeldefinition/agent/CheckoutScript.groovy
 # docker-workflow
-/org/jenkinsci/plugins/docker/workflow/Docker.groovy
```
dwnusbaum, line 24: https://github.com/jenkinsci/docker-workflow-plugin/blob/dc3cf1327f1b8c9967a56aa0e1a1f3f0ff79b57c/src/main/java/org/jenkinsci/plugins/docker/workflow/DockerDSL.java#L58-L59

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
dwnusbaum: Actually IDK if that's right, let me see if we get here with Groovy's speculative class loading as well.

jglick (approved):
Seems safe enough IIUC.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
```diff
@@ -152,6 +148,22 @@ private static boolean endsWithIgnoreCase(String value, String suffix) {
...
+
+        public boolean isAllowed(String groovySourceFileUrl) {
+            return cache.computeIfAbsent(groovySourceFileUrl, url -> {
+                for (GroovySourceFileAllowlist allowlist : GroovySourceFileAllowlist.all()) {
+                    if (allowlist.isAllowed(url)) {
+                        return true;
+                    }
+                }
+                return false;
```
jglick, line 162: I do not think we want to cache negative results? Unclear to me if this would cause problems with dynamic plugin installation with https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/686.

jglick (changes requested):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
jglick: Admittedly only if some weird pipeline attempted to load a plugin-supplied source file directly, and _before_ that plugin was loaded. Anyway, we do not expect allowlists to reject anything under normal operation so it should not be necessary to optimize for that case right?

dwnusbaum: > Admittedly only if some weird pipeline attempted to load a plugin-supplied source file directly, and before that plugin was loaded.

Yeah, I played around with this case when I filed the PR, and I think you would have to be doing something quite unusual in your Pipeline to run into problems due to caching negative results. 

> Anyway, we do not expect allowlists to reject anything under normal operation so it should not be necessary to optimize for that case right?

Sure, fair enough.

dwnusbaum: See [777deff](https://github.com/jenkinsci/workflow-cps-plugin/pull/829/commits/777deff649dbd81b45f63bbdd4d341bb7a473824).

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
```diff
@@ -152,6 +148,27 @@ private static boolean endsWithIgnoreCase(String value, String suffix) {
         }
     }
 
+    @Extension
+    @SuppressFBWarnings(value = "NP_BOOLEAN_RETURN_NULL", justification = "intentionally not caching negative results")
```
dwnusbaum, line 152: Not sure why, but this annotation actually does have to be on the class. If it's on the `isAllowed` itself, SpotBugs still complains.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
```diff
@@ -152,6 +148,27 @@ private static boolean endsWithIgnoreCase(String value, String suffix) {
...
+    public static class AllowedGroovyResourcesCache {
```
jglick, line 153: This could just be folded into `GroovySourceFileAllowlist`; there is no real reason to have a separate extension object for it.

jglick (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GroovySourceFileAllowlist.java
dwnusbaum: Yes, but then it would have to be a static field, and it holds state, which I think we prefer to avoid in static fields.

jglick: Oh, I missed that `GroovySourceFileAllowlist` is `abstract` and not itself an `@Extension`. Never mind.

PR #830, opened 2024-01-10
Forward compatibility with Guice 7

basil:
Without dropping compatibility with Guice 6 (currently delivered by Jenkins core), add support for Guice 7.

### DIFF
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
@@ -1,7 +1,7 @@
 package org.jenkinsci.plugins.workflow.cps.steps;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
-import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java
```diff
@@ -1,7 +1,7 @@
 package org.jenkinsci.plugins.workflow.cps.steps;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
-import javax.inject.Inject;
+import jakarta.inject.Inject;
```
jglick, line 4: FYI https://github.com/jenkinsci/workflow-cps-plugin/blob/dee06329dfb53029be095128cf5098a64287dc96/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java#L39-L40 is a weird old idiom that is completely gratuitous as you can either use `Jenkins.get` or `JenkinsRule.jenkins`.

jglick (approved):

PR #833, opened 2024-01-15
Suppress warnings for spotbugs 4.8.3

MarkEWaite:
## Suppress warnings for spotbugs 4.8.3

Will be needed with parent pom 4.77.

### Testing done

Confirmed that the spotbugs warnings are visible when using the 4.77-SNAPSHOT plugin pom before this change.  With this change, the spotbugs warnings are no longer visible.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/NotBlock.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/NotBlock.java
@@ -34,7 +34,7 @@ public class NotBlock implements Block {
         return new ContinuationImpl(e, k).then(b, e, cast);
     }
 
-    @SuppressFBWarnings(value = "SIC_INNER_SHOULD_BE_STATIC", justification = "Common Jenkins pattern that inner classses are not static.")
+    @SuppressFBWarnings(value = "SIC_INNER_SHOULD_BE_STATIC", justification = "Too late to fix compatibly.")
     class ContinuationImpl extends ContinuationGroup {
         final Continuation k;
         final Env e;
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/NotBlock.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/NotBlock.java
@@ -34,6 +34,7 @@ public class NotBlock implements Block {
         return new ContinuationImpl(e, k).then(b, e, cast);
     }
 
+    @SuppressFBWarnings(value = "SIC_INNER_SHOULD_BE_STATIC", justification = "Common Jenkins pattern that inner classses are not static.")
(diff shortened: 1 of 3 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/impl/NotBlock.java
```diff
@@ -34,6 +34,7 @@ public Next receive(Object o) {
         return new ContinuationImpl(e, k).then(b, e, cast);
     }
 
+    @SuppressFBWarnings(value = "SIC_INNER_SHOULD_BE_STATIC", justification = "Common Jenkins pattern that inner classses are not static.")
```
jglick, line 37: No,

> This class is an inner class, but does not use its embedded reference to the object which created it.  This reference makes the instances of the class larger, and may keep the reference to the creator object alive longer than necessary.  If possible, the class should be made static.

so that is really a bug…but not one that we can fix compatibly now (since builds may be running across the upgrade). So
```suggestion
    @SuppressFBWarnings(value = "SIC_INNER_SHOULD_BE_STATIC", justification = "Too late to fix compatibly.")
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -267,6 +267,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
      *
      * @see #runInCpsVmThread(FutureCallback)
      */
+    @SuppressFBWarnings(value = "PA_PUBLIC_PRIMITIVE_ATTRIBUTE", justification = "Preserve API compatibility.")
```
jglick, line 270: This is not part of any API. Could be at least package-scoped and https://github.com/jenkinsci/workflow-cps-plugin/blob/305192405bc0d5f8923aa75226e97a700aef4c5c/plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java#L150 just deleted or something.
```suggestion
    @SuppressFBWarnings(value = "PA_PUBLIC_PRIMITIVE_ATTRIBUTE", justification = "TODO clean up")
```

jglick (commented):
https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.77 is already out so I guess this should just switch to it?

MarkEWaite:
> https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.77 is already out so I guess this should just switch to it?

Done in db9c7812dc11584b4bac54a9f9afb22088564e7c

jglick (approved):
:bow:

PR #861, opened 2024-02-27
Fix snippet generator when a custom crumb header is configured

lamaral:
The CSRF crumb header can be configured on Jenkins via the java property `hudson.security.csrf.requestfield`.

When an instance overrides the header value with something different than `Jenkins-crumb`, the snippet generator would still use the hard-coded value, causing a 403 to be returned and no snippet to be displayed

### Testing done
On a Jenkins version 2.440.1 on a Debian 12 machine, Jenkins was started with `-Dhudson.security.csrf.requestfield=X-Jenkins-crumb`. 

Then on the https://jenkins.exemple.com/job/example/pipeline-syntax/ page, the `Generate Pipeline Script` button is clicked. 

Without this patch, the AJAX calls gets a 503, as the sent header is `Jenkins-crumb`. With the patch applied (I built a custom version of the plugin locally with Maven), the header is sent as configured by the Java option and the request is successful. 

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
@@ -1,4 +1,4 @@
-function handlePrototype(url, crumb) {
+function handlePrototype(url) {
     buildFormTree(document.forms.config);
     // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
     // TODO simplify when Prototype.js is removed
@@ -7,13 +7,11 @@ function handlePrototype(url, crumb) {
         return; // just a separator
     }
 
-    const headers = new Headers();
-    headers.append("Content-Type", "application/x-www-form-urlencoded");
-    headers.append("Jenkins-Crumb", crumb);
-
     fetch(url, {
         method: "POST",
-        headers: headers,
+        headers: crumb.wrap({
+            "Content-Type": "application/x-www-form-urlencoded"
+        }),
         body: "json=" + encodeURIComponent(json),
 
     })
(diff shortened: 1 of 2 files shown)


jglick (approved):
I think this was a regression in #599 as the previous inlined script used a standard function `crumb.wrap` which uses

```html
<head … data-crumb-header="${h.getCrumbRequestField()}" data-crumb-value="${h.getCrumb(request)}" …>
```

Is that still accessible in this scope? If so, that might be a simpler fix, though this looks fine as well.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
```diff
@@ -71,7 +71,8 @@ THE SOFTWARE.
                     <input type="button" id="generatePipelineScript" value="${%Generate Pipeline Script}"
                            class="submit-button primary"
                            data-url="${rootURL}/${it.GENERATE_URL}"
-                           data-crumb="${h.getCrumb(request)}"/>
+                           data-crumb-header="${h.getCrumbRequestField()}"
+                           data-crumb-value="${h.getCrumb(request)}"/>
```
jtnord, line 75: these should already be set in the `head` element produced by the [`layout`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/layout/layout.jelly#L104-L107) tag.
Do we need to set them here - or can we use what comes from the layout and use `document.head.getAttribute(...)`

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -1,4 +1,4 @@
-function handlePrototype(url, crumb) {
+function handlePrototype(url, crumb_header, crumb_value) {
```
jtnord, line 1: if we have access to the document we should be able to load the crumbs from `document.head`?

Additionally (and I am completely not sure of this) but I think core should handle this for us - we just need to call [`crumb.wrap`](https://www.jenkins.io/blog/2023/05/12/removing-prototype-from-jenkins/#ajax-requests)

e.g. https://github.com/search?q=org%3Ajenkinsci+crumb.wrap&type=code

lamaral: crumb.wrap works in this context. I will clean the rest and update the commit.

jtnord (approved):
LGTM, thankyou

PR #865, opened 2024-03-06
Reparse `CpsFlowExecution.loadedScripts` in original order

jglick:
Finally tracked down the error noted in #812. Could reproduce by

<details>
<summary>Test patch</summary>

```diff
diff --git plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
index 69c2399d..00d050b4 100644
--- plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -150,6 +150,7 @@ import org.codehaus.groovy.GroovyBugError;
 import org.jboss.marshalling.reflect.SerializableClassRegistry;
 
 import static com.thoughtworks.xstream.io.ExtendedHierarchicalStreamWriterHelper.startNode;
+import java.util.Comparator;
 import static org.jenkinsci.plugins.workflow.cps.persistence.PersistenceContext.RUN;
 
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionList;
@@ -635,7 +636,10 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
 
             s = (CpsScript) shell.reparse("WorkflowScript",script);
 
-            for (Entry<String, String> e : loadedScripts.entrySet()) {
+            Map<String, String> ls = new TreeMap<>(Comparator.reverseOrder());
+            ls.putAll(loadedScripts);
+            System.err.println("TODO reparsing " + ls);
+            for (Entry<String, String> e : ls.entrySet()) {
                 shell.reparse(e.getKey(), e.getValue());
             }
         } catch (RuntimeException | Error x) {
```

</details>

`loadedScripts` was being stored as a `HashMap` with (semi-)random order, so if code in a later script referred to a static type name defined in an earlier script, it may or may not succeed in resolving the name.

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -82,7 +82,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -254,7 +253,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
      * Any additional scripts {@linkplain CpsGroovyShell#parse(GroovyCodeSource) parsed} afterward, keyed by
      * their FQCN.
      */
-    /*package*/ /*final*/ Map<String,String> loadedScripts = new HashMap<>();
+    /*package*/ /*final*/ Map<String,String> loadedScripts = new LinkedHashMap<>();
 
     private final boolean sandbox;
     private transient /*almost final*/ FlowExecutionOwner owner;
@@ -644,12 +643,6 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         }
 
         s.execution = this;
-        if (false) {
-            System.out.println("scriptName="+s.getClass().getName());
-            System.out.println(List.of(s.getClass().getInterfaces()));
-            System.out.println(List.of(s.getClass().getDeclaredFields()));
-            System.out.println(List.of(s.getClass().getDeclaredMethods()));
-        }
         return s;
     }
 
@@ -1877,7 +1870,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                     }
 
                     if (result.loadedScripts == null) {
-                        result.loadedScripts = new HashMap<>();   // field added later
+                        result.loadedScripts = new LinkedHashMap<>(); // field added later
                     }
                     result.liveTimings = result.timings == null ?
                         new ConcurrentHashMap<>() :
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
@@ -709,6 +709,7 @@ public class CpsFlowExecutionTest {
         sessions.then(r -> {
             WorkflowJob p = r.createProject(WorkflowJob.class, "p");
             p.setDefinition(new CpsFlowDefinition(
+                "evaluate('true')\n" + // just force this to be Script1
                 "def x = evaluate('class X {X() {}; def m1() {/OK/}}; new X()')\n" +
                 "def y = evaluate('class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()')\n" +
                 "semaphore('wait')\n" +
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecutionTest.java
@@ -720,6 +720,7 @@ public class CpsFlowExecutionTest {
         sessions.then(r -> {
             WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
             WorkflowRun b = p.getLastBuild();
+            FileUtils.copyFile(new File(b.getRootDir(), "build.xml"), System.out);
             SemaphoreStep.success("wait/1", null);
             r.assertLogContains("received really OK", r.assertBuildStatus(Result.SUCCESS, r.waitForCompletion(b)));
         });


jglick:
Found an even better way to reproduce. With patch, test passes with

```xml
<loadedScripts class="linked-hash-map">
  <entry>
    <string>Script1</string>
    <string>true</string>
  </entry>
  <entry>
    <string>Script2</string>
    <string>class X {X() {}; def m1() {/OK/}}; new X()</string>
  </entry>
  <entry>
    <string>Script3</string>
    <string>class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()</string>
  </entry>
</loadedScripts>
```

Without patch, test fails with

```xml
<loadedScripts class="map">
  <entry>
    <string>Script3</string>
    <string>class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()</string>
  </entry>
  <entry>
    <string>Script1</string>
    <string>true</string>
  </entry>
  <entry>
    <string>Script2</string>
    <string>class X {X() {}; def m1() {/OK/}}; new X()</string>
  </entry>
</loadedScripts>
```

```
Started
[Pipeline] Start of Pipeline
[Pipeline] semaphore
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
Script3: 1: unable to resolve class X 
 @ line 1, column 10.
   class Y {X x; def m2() {/really ${x.m1()}/}}; new Y()
            ^

1 error

	at org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:309)
	at org.codehaus.groovy.control.CompilationUnit.applyToSourceUnits(CompilationUnit.java:981)
	at org.codehaus.groovy.control.CompilationUnit.doPhaseOperation(CompilationUnit.java:626)
	at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:575)
	at groovy.lang.GroovyClassLoader.doParseClass(GroovyClassLoader.java:323)
	at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:293)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovySandbox$Scope.parse(GroovySandbox.java:163)
	at org.jenkinsci.plugins.workflow.cps.CpsGroovyShell.doParse(CpsGroovyShell.java:190)
	at org.jenkinsci.plugins.workflow.cps.CpsGroovyShell.reparse(CpsGroovyShell.java:175)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.parseScript(CpsFlowExecution.java:639)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.loadProgramAsync(CpsFlowExecution.java:856)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.onLoad(CpsFlowExecution.java:826)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getExecution(WorkflowRun.java:719)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.onLoad(WorkflowRun.java:577)
	at hudson.model.RunMap.retrieve(RunMap.java:233)
	at hudson.model.RunMap.retrieve(RunMap.java:61)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:650)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:632)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:530)
	at hudson.model.RunMap.getById(RunMap.java:213)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.run(WorkflowRun.java:968)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.get(WorkflowRun.java:980)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList.resume(FlowExecutionList.java:226)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$ItemListenerImpl.onLoaded(FlowExecutionList.java:217)
	at jenkins.model.Jenkins.<init>(Jenkins.java:1042)
	at hudson.model.Hudson.<init>(Hudson.java:86)
	at org.jvnet.hudson.test.JenkinsRule.newHudson(JenkinsRule.java:747)
	at org.jvnet.hudson.test.JenkinsRule.before(JenkinsRule.java:409)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:650)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
	at java.base/java.lang.Thread.run(Thread.java:829)
ERROR: Failed to load program
Finished: FAILURE
```

rsandell (approved):

jglick:
Oh one more note: this bug is specifically about defining types inside blocks of code loaded via `evaluate` (or probably also `load`). Libraries would not suffer from this problem because `src/**/*.groovy` are arranged in a package hierarchy, so the `GroovyClassLoader` knows how to find a fully-qualified type name on disk.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -254,7 +253,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
      * Any additional scripts {@linkplain CpsGroovyShell#parse(GroovyCodeSource) parsed} afterward, keyed by
      * their FQCN.
      */
-    /*package*/ /*final*/ Map<String,String> loadedScripts = new HashMap<>();
+    /*package*/ /*final*/ Map<String,String> loadedScripts = new LinkedHashMap<>();
```
Vlatombe, line 256: Maybe even

```suggestion
    /*package*/ /*final*/ LinkedHashMap<String,String> loadedScripts = new LinkedHashMap<>();
```

since it is intended to retain insertion order.

Vlatombe (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Unfortunately no, because previously serialized `build.xml`s would include a `HashMap`.

rsandell: My first thought would be a `TreeMap` but I guess this works just as well.

jglick: My first thought _was_ a `TreeMap`. However it would need to use a special `Comparator` since then for example `Script10` would come before `Script2`. `LinkedHashMap` seems simpler and safer; if the original `evaluate`s in the prior Jenkins session loaded scripts successfully in a certain order, after resumption we should try to load them in the same order.

PR #867, opened 2024-03-10
Move copy button and update generate button

NotMyFault:
The change proposed moves the "copy to clipboard" button above the output box, matching with other buttons' locations.
Additionally, I've changed the generation button from YUI to make use the Jenkins button layout:

Before:
Screenshot 2024-03-10 at 09 51 54

After:
Screenshot 2024-03-10 at 09 51 09


jtnord (commented):
aside from the inline comments the buttons are not vertically aligned - the copy button is quite a bit higher than the generate button.
This could be an issue in the Jenkins styles, or in the layout used here.

I am +0 on the moving of the button.  In one hand it means once you know the flow there is less need to move the mouse (or tab), on the other your eyes would normally click the button, then scan the text so your eyes would be at the bottom of the textarea.

jglick:
(merge conflicts)

PR #877, opened 2024-03-29
More efficient use of `GroovyCategorySupport`

jglick:
Follows up #124 / https://github.com/cloudbees/groovy-cps/pull/56 by calling `GroovyCategorySupport.use` only once per thread, rather than on every chunk.

Background: `CpsVmExecutorService` has a subtle design in that it is physically hosted on a fairly generic elastic thread pool; the CPS VM “thread” associated with a Pipeline build is virtual (does not correspond to any permanent Java `Thread`) and just makes use of some free physical thread whenever it has a chunk of work to run. Typically a chunk would involve as much Groovy logic as can run synchronously (milliseconds’ worth, you hope), up to the point where an asynchronous operation needs to take place, such as starting a step. (`parallel` is actually coöperative multitasking: step implementations can certainly run concurrently, but any interactions with the Groovy program are single-threaded. Thus the maximum number of physical threads potentially needed is the number of concurrently running builds, regardless of the number of branches in a single build.)

Now in order to support some Groovy hacks we do to make normal-looking language idioms transparently run in CPS mode (#124), we use `GroovyCategorySupport` which sets up some hooks in a dynamic scope (applicable to the current `Thread`), and its sole API is to take a `Closure` to run with those hooks in effect. Normally setting up and tearing down the hooks is expected to be very quick, but in a heavily loaded controller with complex Pipeline libraries it seems it can be slow and even a source of contention: pool threads from unrelated builds are waiting on the same static (global) lock. (CloudBees-internal reference: SECO-3635)

```
waiting to lock <0x…> (a java.lang.Class)
owned by "Running CpsFlowExecution[…]]" id=… (0x…)
	at org.codehaus.groovy.vmplugin.v7.IndyInterface.invalidateSwitchPoints(IndyInterface.java:124)
	at org.codehaus.groovy.vmplugin.v7.Java7.invalidateCallSites(Java7.java:70)
	at org.codehaus.groovy.runtime.GroovyCategorySupport$ThreadCategoryInfo.endScope(GroovyCategorySupport.java:109)
	at org.codehaus.groovy.runtime.GroovyCategorySupport$ThreadCategoryInfo.use(GroovyCategorySupport.java:138)
	at org.codehaus.groovy.runtime.GroovyCategorySupport.use(GroovyCategorySupport.java:275)
	at com.cloudbees.groovy.cps.Continuable.run0(Continuable.java:146)
```

The repeated setup and teardown of categories is actually entirely wasted work, since we set the same hooks for all builds, but Groovy (at least the v2 we run) does not offer an API to permanently install a hook. So to minimize the frequency with which we acquire this global lock, this PR passes Groovy a closure which runs any number of chunks of work from one or more builds in sequence, saving most of the overhead. Java still manages the thread pool as before, creating new threads when there are a lot of concurrent builds doing significant Groovy work, and discarding them when idle.

Note that we do not currently support categories from user code (https://github.com/cloudbees/groovy-cps/issues/16) and it was never clear how we could.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/Continuable.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/Continuable.java
@@ -23,6 +23,9 @@ import org.codehaus.groovy.runtime.GroovyCategorySupport;
  */
 public class Continuable implements Serializable {
 
+    /**
+     * Users of this library must pass (at least) these to {@link GroovyCategorySupport#use(List, Closure)} during all operations.
+     */
     @SuppressWarnings("rawtypes")
     public static final List<Class> categories = List.of(
         CpsDefaultGroovyMethods.class,
@@ -133,31 +136,21 @@ public class Continuable implements Serializable {
         return run0(new Outcome(null,arg)).wrapReplay();
     }
 
-    @Deprecated
-    public Outcome run0(final Outcome cn) {
-        return run0(cn, categories);
-    }
-
     /**
      * Resumes this program by either returning the value from {@link Continuable#suspend(Object)} or
      * throwing an exception
      */
-    public Outcome run0(final Outcome cn, List<Class> categories) {
-        return GroovyCategorySupport.use(categories, new Closure<>(null) {
-            @Override
-            public Outcome call() {
-                Next n = cn.resumeFrom(e,k);
+    public Outcome run0(final Outcome cn) {
+        Next n = cn.resumeFrom(e,k);
 
-                while(n.yield==null) {
-                    n = n.step();
-                }
+        while(n.yield==null) {
+            n = n.step();
+        }
 
-                e = n.e;
-                k = n.k;
+        e = n.e;
+        k = n.k;
 
-                return n.yield;
-            }
-        });
+        return n.yield;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/Continuable.java
```diff
@@ -133,31 +136,21 @@ public Object runByThrow(Throwable arg) throws InvocationTargetException {
...
-                }
-
-                e = n.e;
-                k = n.k;
-
-                return n.yield;
-            }
-        });
+    public Outcome run0(final Outcome cn) {
+        Next n = cn.resumeFrom(e,k);
```
jglick, line 144: (ignore WS)

dwnusbaum (approved):
FWIW, my main concerns with this kind of change were that it might introduce complex cross-script memory leaks and/or security issues. Looking at `GroovyCategorySupport.java` though, I don't really see any obvious problems, mainly given "Groovy (at least the v2 we run) does not offer an API to permanently install a hook" and https://github.com/cloudbees/groovy-cps/issues/16.

One thing to note though is that user-defined category methods and `use` do work in `@NonCPS` methods (`use` is not allowed in the sandbox by default), so it is at least possible for users to interact with user-defined categories in limited ways. Calls like https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/runtime/GroovyCategorySupport.java#L187 in the implementation could end up storing user-defined classes in the global `ClassInfo` cache, but this seems no different than just using the class directly in a script, and  `CpsFlowExecution.cleanUpGlobalClassValue` will clean all of that up when the Pipeline completes, so that seems fine.

jglick:
Right. The categories is just a list of (plugin-defined) classes, and even a `use` call in `@NonCPS` from a trusted library would run `endScope` as soon as it is done, so I do not see anything that could leak.

[This](https://github.com/jenkinsci/workflow-cps-plugin/pull/877/checks?check_run_id=23248853788) smells like a flake but I will recheck.

dwnusbaum (approved):

rsandell (approved):

PR #897, opened 2024-06-14
Add trilead-api as test dependency to avoid PCT failure

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -268,6 +268,12 @@
             <version>1.10.10</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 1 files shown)


jtnord:
I'm not sure this is correct the dep it should come from the subversion plugin but it doesn't which was a bug in the subversion plugin that you fixed :)

https://github.com/jenkinsci/subversion-plugin/releases/tag/1256.vee91953217b_6

jglick:
I am guessing this is obsolete, particularly after #922.

PR #907, opened 2024-07-17
[JENKINS-73470] Option to hide "Use Groovy Sandbox"

amuniz:
See [JENKINS-73470](https://issues.jenkins.io/browse/JENKINS-73470) for details.

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
@@ -89,6 +89,7 @@ public class CpsFlowDefinition extends FlowDefinition {
     @DataBoundConstructor
     public CpsFlowDefinition(String script, boolean sandbox) throws Descriptor.FormException {
         if (CPSConfiguration.get().isHideSandbox() && !sandbox && !Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
+            // this will end up in the /oops page until https://github.com/jenkinsci/jenkins/pull/9495 is picked up
             throw new Descriptor.FormException("Sandbox cannot be disabled. This Jenkins instance has been configured to not " +
                     "allow regular users to disable the sandbox in pipelines", "sandbox");
         }
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
@@ -34,8 +34,10 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.util.FormValidation;
 import hudson.util.StreamTaskListener;
+import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 import org.apache.commons.lang.StringUtils;
+import org.jenkinsci.plugins.workflow.cps.config.CPSConfiguration;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.flow.DurabilityHintProvider;
 import org.jenkinsci.plugins.workflow.flow.FlowDefinition;
@@ -43,6 +45,8 @@ import org.jenkinsci.plugins.workflow.flow.FlowDefinitionDescriptor;
 import org.jenkinsci.plugins.workflow.flow.FlowDurabilityHint;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.flow.GlobalDefaultFlowDurabilityLevel;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 
@@ -100,6 +104,12 @@ public class CpsFlowDefinition extends FlowDefinition {
 
     public boolean isSandbox() {
         return sandbox;
+    }
+
+    @Restricted(NoExternalUse.class) // stapler
+    public boolean shouldHideSandbox() {
+        // sandbox checkbox is shown to admins even if the global configuration says otherwise
+        return CPSConfiguration.get().isHideSandbox() && !Jenkins.get().hasPermission(Jenkins.ADMINISTER);
     }
 
     // Used only from Groovy tests.
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.AbortException;
 import hudson.Extension;
 import hudson.model.Action;
+import hudson.model.Descriptor;
 import hudson.model.Failure;
 import hudson.model.Item;
 import hudson.model.Job;
@@ -81,16 +82,21 @@ public class CpsFlowDefinition extends FlowDefinition {
      * @deprecated use {@link #CpsFlowDefinition(String, boolean)} instead
      */
     @Deprecated
-    public CpsFlowDefinition(String script) {
+    public CpsFlowDefinition(String script) throws Descriptor.FormException {
         this(script, false);
     }
 
     @DataBoundConstructor
-    public CpsFlowDefinition(String script, boolean sandbox) {
+    public CpsFlowDefinition(String script, boolean sandbox) throws Descriptor.FormException {
+        if (CPSConfiguration.get().isHideSandbox() && !sandbox && !Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
+            throw new Descriptor.FormException("Sandbox cannot be disabled. This Jenkins instance has been configured to not " +
+                    "allow regular users to disable the sandbox in pipelines", "sandbox");
+        }
         StaplerRequest req = Stapler.getCurrentRequest();
         this.script = sandbox ? script : ScriptApproval.get().configuring(script, GroovyLanguage.get(),
                 ApprovalContext.create().withCurrentUser().withItemAsKey(req != null ? req.findAncestorObject(Item.class) : null), req == null);
         this.sandbox = sandbox;
+
     }
 
     private Object readResolve() {
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.cps;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.AbortException;
 import hudson.Extension;
@@ -115,13 +116,6 @@ public class CpsFlowDefinition extends FlowDefinition {
         return sandbox;
     }
 
-    @Restricted(NoExternalUse.class) // stapler
-    public boolean shouldHideSandbox() {
-        // sandbox checkbox is shown to admins even if the global configuration says otherwise
-        // it's also shown when sandbox == false, so regular users can enable it
-        return CPSConfiguration.get().isHideSandbox() && !Jenkins.get().hasPermission(Jenkins.ADMINISTER) && sandbox;
-    }
-
     // Used only from Groovy tests.
     public CpsFlowExecution create(FlowExecutionOwner handle, Action... actions) throws IOException {
         return create(handle, StreamTaskListener.fromStderr(), List.of(actions));
@@ -196,6 +190,14 @@ public class CpsFlowDefinition extends FlowDefinition {
             }
             return CpsFlowDefinitionValidator.CheckStatus.SUCCESS.asJSON();
             // Approval requirements are managed by regular stapler form validation (via doCheckScript)
+        }
+
+        @Restricted(NoExternalUse.class) // stapler
+        public boolean shouldHideSandbox(@CheckForNull CpsFlowDefinition instance) {
+            // sandbox checkbox is shown to admins even if the global configuration says otherwise
+            // it's also shown when sandbox == false, so regular users can enable it
+            return CPSConfiguration.get().isHideSandbox() && !Jenkins.get().hasPermission(Jenkins.ADMINISTER)
+                    && (instance == null || instance.sandbox);
         }
 
     }
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
@@ -25,8 +25,10 @@
 package org.jenkinsci.plugins.workflow.cps;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.AbortException;
 import hudson.Extension;
 import hudson.model.Action;
+import hudson.model.Failure;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.Queue;
@@ -109,7 +111,8 @@ public class CpsFlowDefinition extends FlowDefinition {
     @Restricted(NoExternalUse.class) // stapler
     public boolean shouldHideSandbox() {
         // sandbox checkbox is shown to admins even if the global configuration says otherwise
-        return CPSConfiguration.get().isHideSandbox() && !Jenkins.get().hasPermission(Jenkins.ADMINISTER);
+        // it's also shown when sandbox == false, so regular users can enable it
+        return CPSConfiguration.get().isHideSandbox() && !Jenkins.get().hasPermission(Jenkins.ADMINISTER) && sandbox;
     }
 
     // Used only from Groovy tests.
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
@@ -1,0 +1,33 @@
+package org.jenkinsci.plugins.workflow.cps.config;
+
+import hudson.Extension;
+import hudson.ExtensionList;
+import jenkins.model.GlobalConfiguration;
+import org.jenkinsci.Symbol;
+
+@Extension
+@Symbol("cpsGlobalConfiguration")
+public class CPSConfiguration extends GlobalConfiguration {
+
+    /**
+     * Whether to show the sandbox checkbox in jobs to users without {@link jenkins.model.Jenkins.ADMINISTER}
+     */
+    private boolean hideSandbox;
+
+    public CPSConfiguration() {
+        load();
+    }
+
+    public boolean isHideSandbox() {
+        return hideSandbox;
+    }
+
+    public void setHideSandbox(boolean hideSandbox) {
+        this.hideSandbox = hideSandbox;
+        save();
+    }
+
+    public static CPSConfiguration get() {
+        return ExtensionList.lookupSingleton(CPSConfiguration.class);
+    }
+}
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
@@ -31,7 +31,7 @@ import jenkins.model.GlobalConfiguration;
 import jenkins.model.GlobalConfigurationCategory;
 import org.jenkinsci.Symbol;
 
-@Symbol("cpsGlobalConfiguration")
+@Symbol("cps")
 @Extension
 public class CPSConfiguration extends GlobalConfiguration {
 
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
@@ -1,3 +1,27 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2024, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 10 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -102,6 +106,12 @@ public boolean isSandbox() {
         return sandbox;
     }
 
+    @Restricted(NoExternalUse.class) // stapler
+    public boolean shouldHideSandbox() {
+        // sandbox checkbox is shown to admins even if the global configuration says otherwise
+        return CPSConfiguration.get().isHideSandbox() && !Jenkins.get().hasPermission(Jenkins.ADMINISTER);
```
jglick, line 112: If sandbox is already disabled, you should still show the checkbox; otherwise the user would have no way to turn it back on from the GUI.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
jglick, line 1: Probably makes sense to set the category to the security one so it appears in `/configureSecurity/` rather than `/configure`.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-showSandbox.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    <p>Controls whether the "Use Groovy Sandbox" is shown in pipeline jobs configuration to users without Overall/Administer permission.</p>
+    <p>Note that this is not a security configuration, as it does not prevent users to configure and run
+        pipelines in some other ways without using the sandbox.</p>
```
jglick, line 4: Enhancements under discusssion:
* Throw `AbortException` if an attempt is made to run a non-sandboxed pipeline.
* Throw something from the `@DataBoundConstructor` if `!sandbox`.

The first option may not work though, because we allow an admin to define a non-sandboxed pipeline (maybe some special admin/diagnostic job), and we would not know when the build starts who edited it.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
amuniz: Not sure. If the sandbox was disabled in the job it was probaby for a good reason, so enable it back probably requires an admin intervention (to approve some signature maybe?), and that point the admin can also enable it.

Also, it would generate a weird behaviour that users won't understand: you see the checkbox, enable it, save, next time you open the job configure page there is no checkbox... weird, at least.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
amuniz: As noted in the help text, this is not a security setting, as it can still be easily bypassed by using other interfaces to create jobs (CLI, HTTP, etc). This must be seen as an UX improvement under some specific circumstances, nothing more.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-showSandbox.html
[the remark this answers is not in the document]
amuniz: > Throw something from the @DataBoundConstructor if !sandbox.

It should also check if the current user is an admin (as they are allowed to disable the sandbox), but jobs can be created by other means which do not necesasrily have a user context (`job-dsl` for example, or CasC for items when using CloudBees CI). So this would not be a complete solution either. So given that there is not a complete solution, I decided to fix the minimum required to solve the UX issue that happens when using the UI.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
jglick: Not ideal, but seems better than having a job be broken in a way that the user cannot see and cannot fix without resort to REST or CLI.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
jglick: Perhaps, but it is clearly related to security, and there are other settings in this page that are not strictly about security defense: ports, markup formatter, etc.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-showSandbox.html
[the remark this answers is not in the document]
jglick: > jobs can be created by other means which do not necessarily have a user context (`job-dsl` for example)

In this case the person configuring the DSL job (or approving changes to its source code) is effectively a Jenkins admin whom we trust to decide whether to allow the sandbox to be disabled. Checking `hasPermission` would work uniformly, since `SYSTEM` would pass.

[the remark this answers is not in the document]
amuniz: I've been trying to block item creation/update by failing on the `@DataBoundConstructor` of `CpsFlowDefinition`. However UX is horrible, getting _Angry Jenkins_ (no matter what exception type I use - tried with `RuntimeException` and `Failure`, same result).

An alternative solution would be to unconditionally set `sandbox = true` when the right conditions are met (`CPSConfiguration.get().isHideSandbox() && !sandbox && !Jenkins.get().hasPermission(Jenkins.ADMINISTER)`). But then the sandbox would be _silently_ enabled when non-admin users save a job which had the sandbox disabled previously.

 Given all that, I think we should not open that can of worms and just hide the checkbox (current state of the PR).

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
amuniz: Done in https://github.com/jenkinsci/workflow-cps-plugin/pull/907/commits/5b9eccddefb3ae6c7ac2d01dec45aeb884a1d095

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
amuniz: Moved and help text updated to emphasize on the non-security aspect.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-showSandbox.html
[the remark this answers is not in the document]
jglick: (Discussed out of band: core might have seen a regression in handling of `Failure` as of https://github.com/jenkinsci/jenkins/pull/4389. https://github.com/jenkinsci/jenkins/blob/cc97ad8f330aff43f412a95235a5386194803b14/core/src/main/java/hudson/model/Failure.java#L36-L41 is supposed to render https://github.com/jenkinsci/jenkins/blob/cc97ad8f330aff43f412a95235a5386194803b14/core/src/main/resources/hudson/model/AbstractModelObject/error.jelly#L33-L41; that is the whole purpose of this class.)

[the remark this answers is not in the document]
amuniz: Reverting that particular change generates a lot of non-trivial conflicts. So I've _manually_ reverted the main chunks in there and the behaviour is still not the expected. The _oops_ page is rendered (including the stacktrace). `Failure` is not behaving different than any other unhandled exception. I'm not sure how this would have ever worked, as it rendered the `oops.jelly` since [it's creation](https://github.com/jenkinsci/jenkins/commit/6badbb3227701f93aeebf208cc47f8b1540617da).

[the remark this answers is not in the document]
jglick: `Failure` still works in general, including for non-admin users. Checked by trying to copy a nonexistent job as a user lacking `ADMINISTER`; rendered the polite error page with message as expected.
failure


```
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:1389)
	at hudson.model.Failure.<init>(Failure.java:56)
	at hudson.model.Failure.<init>(Failure.java:50)
	at hudson.model.ItemGroupMixIn.createTopLevelItem(ItemGroupMixIn.java:179)
	at jenkins.model.Jenkins.doCreateItem(Jenkins.java:4281)
	at jenkins.model.Jenkins.doCreateItem(Jenkins.java:344)
	at hudson.model.AllView.doCreateItem(AllView.java:98)
	at …
```

rendered in

```
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:1389)
	at hudson.model.Failure.generateResponse(Failure.java:68)
	at org.kohsuke.stapler.HttpResponseRenderer$Default.handleHttpResponse(HttpResponseRenderer.java:129)
	at org.kohsuke.stapler.HttpResponseRenderer$Default.generateResponse(HttpResponseRenderer.java:71)
	at org.kohsuke.stapler.Function.renderResponse(Function.java:162)
	at org.kohsuke.stapler.Function.bindAndInvokeAndServeResponse(Function.java:152)
	at org.kohsuke.stapler.MetaClass$11.doDispatch(MetaClass.java:644)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:61)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:804)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:942)
	at org.kohsuke.stapler.MetaClass$4.doDispatch(MetaClass.java:327)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:61)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:804)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:942)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:725)
	at org.kohsuke.stapler.Stapler.service(Stapler.java:253)
	at …
```

What was the user gesture you were testing, and what was its stack trace? Perhaps some intermediate method was catching `Failure` and rethrowing something else, or perhaps it was from an endpoint which somehow bypasses `HttpResponseRenderer`.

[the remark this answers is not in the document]
amuniz:
```diff
diff --git a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
index aeae1865..153b2352 100644
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
@@ -91,6 +91,7 @@ public class CpsFlowDefinition extends FlowDefinition {
         this.script = sandbox ? script : ScriptApproval.get().configuring(script, GroovyLanguage.get(),
                 ApprovalContext.create().withCurrentUser().withItemAsKey(req != null ? req.findAncestorObject(Item.class) : null), req == null);
         this.sandbox = sandbox;
+        throw new Failure("Cannot save!!");
     }
 
     private Object readResolve() {
```
And try to save a pipeline.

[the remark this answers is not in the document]
jglick: Ah, because it is wrapped in `Descriptor.newInstance` as you can see in the stack trace. I tried `Descriptor.FormException`, which is supposed to be for this purpose. (In fact we could even revert the code to hide the checkbox—this is arguably better UX.) Unfortunately right now it is getting wrapped by `RequestImpl.invokeConstructor` then `RequestImpl.TypePair.convertJSON`. Seems like a Stapler bug :eyes:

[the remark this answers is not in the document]
amuniz: Picked https://github.com/jenkinsci/stapler/pull/571 + https://github.com/jenkinsci/jenkins/pull/9495 up. It's working.

However, I just realized that we can't block the `create-job` CLI as it's not going through the data-bound constructor.

amuniz:
Build is failing with:
```
10:51:55  [INFO] Rule 2: org.apache.maven.enforcer.rules.version.RequireJavaVersion passed
10:51:55  [INFO] Restricted to JDK 11 yet org.jenkins-ci.main:websocket-spi:jar:2.470-rc35157.fa_9b_5f84f785:provided contains jenkins/websocket/Provider$Handler.class targeted to JDK 17
10:51:55  [INFO] Restricted to JDK 11 yet org.kohsuke.stapler:stapler-groovy:jar:1891.v4efde02f1509:provided contains org/kohsuke/stapler/jelly/groovy/GroovierJellyScript$AliasMetaMethod.class targeted to JDK 17
10:51:55  [INFO] Restricted to JDK 11 yet org.kohsuke.stapler:stapler-jelly:jar:1891.v4efde02f1509:provided contains org/kohsuke/stapler/framework/adjunct/Adjunct$Kind.class targeted to JDK 17
10:51:55  [INFO] Restricted to JDK 11 yet org.kohsuke.stapler:stapler:jar:1891.v4efde02f1509:provided contains org/apache/commons/fileupload/FileItem$1.class targeted to JDK 17
10:51:55  [INFO] Restricted to JDK 11 yet org.jenkins-ci.main:cli:jar:2.470-rc35157.fa_9b_5f84f785:provided contains hudson/cli/CLI$1.class targeted to JDK 17
10:51:55  [INFO] Restricted to JDK 11 yet org.jenkins-ci.main:jenkins-core:jar:2.470-rc35157.fa_9b_5f84f785:provided contains antlr/ANTLRException.class targeted to JDK 17
```
Coming from the incremental release of https://github.com/jenkinsci/jenkins/pull/9495 - are incrementals building without JDK 11 release compatibility?

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-showSandbox.html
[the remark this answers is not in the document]
jglick: Yes, both of those would go through `readResolve` as called by XStream. At least the CLI will treat certain exception types like `IllegalArgumentException` specially for purposes of error handling, if it matters.

jglick:
> Coming from the incremental release of jenkins

Which requires Java 17, so to specify such a `jenkins.version` you would need to switch the plugin to Java 17 using instructions defined in the parent POM docs (I hope).

amuniz:
How do you usually handle this type of changes? Wait until the change is integrated in a weekly and bump to the weekly here? Or wait until the change is in an LTS?

jglick:
Neither. Revert the `jenkins.version` bump, and skip the test unless the version in fact includes the fix. You are not using a new API. Someone with an older core will merely see an angry Jenkins rather than a polite message.

amuniz:
Ready for final review.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -353,11 +353,12 @@ public void cpsScriptSandboxHide() throws Exception {
...
                 jenkins.submit(config);
-                fail("Expected HTTP 400");
+                fail("Expected HTTP 500"); // should be 400 when https://github.com/jenkinsci/jenkins/pull/9495
             } catch (FailingHttpStatusCodeException e) {
                 // good, expected
-                assertThat(e.getStatusCode(), equalTo(400));
-                assertThat(e.getResponse().getContentAsString(StandardCharsets.UTF_8), containsStringIgnoringCase("Sandbox cannot be disabled"));
+                assertThat(e.getStatusCode(), equalTo(500)); // should be 400 when https://github.com/jenkinsci/jenkins/pull/9495
+                // uncomment when should be 400 when https://github.com/jenkinsci/jenkins/pull/9495
+                //assertThat(e.getResponse().getContentAsString(StandardCharsets.UTF_8), containsStringIgnoringCase("Sandbox cannot be disabled"));
```
jglick, line 361: You actually need to check the Jenkins version and switch test behavior accordingly. (Or skip the test entirely when the version is too low.) Otherwise `bom-weekly` PCT will complain.

jglick:
Core PR merged.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionTest.java
```diff
@@ -351,14 +352,20 @@ public void cpsScriptSandboxHide() throws Exception {
             // Get the last one, because previous ones might be from Lockable Resources during PCT.
             HtmlCheckBoxInput sandbox = (HtmlCheckBoxInput) sandboxes.get(sandboxes.size() - 1);
             assertFalse("Sandbox is disabled", sandbox.isChecked());
+            VersionNumber jenkinsVersion = new VersionNumber(Jenkins.VERSION);
+            int expectedStatus = 500;
+            if (jenkinsVersion.isNewerThanOrEqualTo(new VersionNumber("2.470"))) { // including https://github.com/jenkinsci/jenkins/pull/9495
```
jglick, line 357:
```suggestion
            if (jenkinsVersion.isNewerThanOrEqualTo(new VersionNumber("2.470"))) { // TODO pending https://github.com/jenkinsci/jenkins/pull/9495 in baseline
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -75,16 +82,22 @@ public class CpsFlowDefinition extends FlowDefinition {
...
      */
     @Deprecated
-    public CpsFlowDefinition(String script) {
+    public CpsFlowDefinition(String script) throws Descriptor.FormException {
         this(script, false);
     }
 
     @DataBoundConstructor
-    public CpsFlowDefinition(String script, boolean sandbox) {
+    public CpsFlowDefinition(String script, boolean sandbox) throws Descriptor.FormException {
```
jglick, line 90: Note: source-incompatible but binary-compatible.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+package org.jenkinsci.plugins.workflow.cps.config;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.ExtensionList;
+import jenkins.model.GlobalConfiguration;
+import jenkins.model.GlobalConfigurationCategory;
+import org.jenkinsci.Symbol;
+
+@Symbol("cpsGlobalConfiguration")
```
jglick, line 34:
```suggestion
@Symbol("cps")
```
or something like that. See Javadoc for `@Symbol`.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
```diff
@@ -28,8 +28,8 @@
   <f:entry title="${%Script}" field="script">
     <wfe:workflow-editor />
   </f:entry>
-  <f:entry field="sandbox">
-    <f:checkbox title="${%Use Groovy Sandbox}" default="true"/>
+  <f:entry field="sandbox" class="${instance.shouldHideSandbox() ? 'jenkins-hidden' : ''}">
+    <f:checkbox title="${%Use Groovy Sandbox}" default="true" />
```
jglick, line 32:
```suggestion
    <f:checkbox title="${%Use Groovy Sandbox}" default="true"/>
```
(minimize diff; the extra space is a relic of HTML)

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-hideSandbox.html
```diff
@@ -0,0 +1,6 @@
+<div>
+    <p>Controls whether the "Use Groovy Sandbox" is shown in pipeline jobs configuration page to users without Overall/Administer permission.</p>
+    <p>This can be used to get a better UX in highly secured environments where all pipelines are required to run in the sandbox (ie. running arbitrary code is never approved)</p>
+    <p>Note that this does not prevent users to configure and run pipelines with sandbox disabled if they create or update jobs by other means (like CLI or HTTP API).
```
jglick, line 4: So for now we are not touching `readResolve`, deliberately?

amuniz: Yes. I don't see how to differentiate from a regular call during start up and a call from a CLI/HTTP API call.

jglick: Deserialization during startup would run as `SYSTEM`, which of course has `ADMINISTER`; deserialization during a CLI/HTTP call would run as some user authentication.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -75,16 +82,22 @@ public class CpsFlowDefinition extends FlowDefinition {
...
+                    "allow regular users to disable the sandbox in pipelines", "sandbox");
+        }
         StaplerRequest req = Stapler.getCurrentRequest();
         this.script = sandbox ? script : ScriptApproval.get().configuring(script, GroovyLanguage.get(),
                 ApprovalContext.create().withCurrentUser().withItemAsKey(req != null ? req.findAncestorObject(Item.class) : null), req == null);
         this.sandbox = sandbox;
+
     }
 
     private Object readResolve() {
```
jglick, line 103: note :point_down:

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/config.jelly
```diff
@@ -28,7 +28,7 @@
   <f:entry title="${%Script}" field="script">
     <wfe:workflow-editor />
   </f:entry>
-  <f:entry field="sandbox">
+  <f:entry field="sandbox" class="${instance.shouldHideSandbox() ? 'jenkins-hidden' : ''}">
```
jglick, line 31: On a newly created project, `instance == null`, so this should be a method on the `descriptor` passing the `instance` as a `@CheckForNull` arg. Otherwise JEXL will silently convert the implicit NPE to a false and show the checkbox (I think).

amuniz: Good catch! Fixed.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/help-hideSandbox.html
amuniz: IMO, we should leave this as is and revisit it separately if/when needed/requested.

jglick (approved):

PR #909, opened 2024-07-24
Update to 2.426.x, simplify JSON parsing

basil:
Now that Prototype.js has been removed in 2.426.x, we can remove the no-longer needed `Object.toJSON` workaround.

### Testing done

`mvn clean verify -DskipTests`


plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/handle-prototype.js
```diff
@@ -1,8 +1,6 @@
 function handlePrototype(url) {
     buildFormTree(document.forms.config);
-    // TODO JSON.stringify fails in some circumstances: https://gist.github.com/jglick/70ec4b15c1f628fdf2e9 due to Array.prototype.toJSON
```
jglick, line 3: (Confirmed that the bug described in the Gist is no longer reproducible in script console of either Firefox or Chrome.)

jglick (approved):
Automated tests do not cover `Snippetizer` GUI (that I can recall), so did a quick sanity test manually.

PR #901, opened 2024-06-20
Enable Jenkins Security Scan

strangelookingnerd:
It is stated in the `JENKINS SECURITY POLICY` that

> The Jenkins project takes security seriously. We make every possible effort to ensure users can adequately secure their automation infrastructure. To that end, we work with Jenkins core and plugin developers, as well as security researchers, to fix security vulnerabilities in Jenkins in a timely manner, and to improve the security of Jenkins in general.

We as plugin developers should try our best to support this. One way of doing so is to enable the Jenkins Security Scan on this repository. For more details see the [documentation](https://www.jenkins.io/doc/developer/security/scan/).

This PR may create new issues in the Code scanning section of this project once it is merged. I'd be happy to try and help resolve those if necessary.

_Please note this PR was created semi-automatically. If you find any issue with it, don't hesitate to ping me._

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #915, opened 2024-08-05
Removing stock security scan based on Java 11

jglick:
Pending https://github.com/jenkinsci/workflow-cps-plugin/pull/901#discussion_r1704627521 which I do not really understand.


jtnord (approved):

jtnord:
@jglick Fine by me, but I removed the auto-merge as I was not sure if you wanted to actually merge this before you got / saw a reply for the referenced [comment](https://github.com/jenkinsci/workflow-cps-plugin/pull/901#discussion_r1704627521)

jglick:
Yes I think so; we know we will want to switch to a Java 17 baseline in the near future so it will just be confusing at that point to have a broken workflow.

PR #916, opened 2024-08-08
Improve support bundle data for running Pipelines, update support bundle component names and file paths, and update log warning for a particular type of error during step completion

dwnusbaum:
Relates to https://github.com/jenkinsci/workflow-cps-plugin/pull/892, https://github.com/jenkinsci/workflow-api-plugin/pull/347, and https://github.com/jenkinsci/workflow-api-plugin/pull/349.

Recently I have seen two cases where the CPS VM thread task queue is either very backed up or completely stuck for some builds, leading to various problems. It is currently difficult to detect this state without either running Groovy scripts that use reflection to access internal state of `SingleLaneExecutorService` or analyzing a heap dump. I think it is worth tracking both the amount of time that a build spends waiting for queued tasks to start as well as the current number of queued tasks.

Here is an example of `pipeline-thread-dump.txt` with the data from this PR:

```
Build: p #39
Started: 2024-08-12T20:08:44.701Z
Duration: PT7.651134S
Timings:
  classLoad	33ms
  flowNode	69ms
  parse	69ms
  run	177ms
  runQueue	417ms
  saveProgram	50ms
Active operations:
  run	1	7326ms
  runQueue	7	51535ms
Approximate graph size: 11
Thread #5
	at DSL.sleep(sleeping for another 52 sec)
	at WorkflowScript.run(WorkflowScript:3)
	at DSL.parallel(Native Method)
	at WorkflowScript.run(WorkflowScript:1)
Thread #6
	at WorkflowScript.run(WorkflowScript:5)
Thread #7
	at WorkflowScript.run(WorkflowScript:7)
Thread #8
	at WorkflowScript.run(WorkflowScript:9)


```

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -387,6 +387,10 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
          */
         run,
         /**
+         * Time spent waiting in the queue for {@link CpsVmExecutorService}.
+         */
+        runQueue,
+        /**
          * Saving the program state.
          * @see CpsThreadGroup#saveProgram(File)
          */
@@ -400,7 +404,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
 
     /** accumulated time in ns of a given {@link TimingKind#name}; {@link String} key for pretty XStream form */
     transient @NonNull Map<String, LongAdder> liveTimings = new ConcurrentHashMap<>();
-    /** XStream simplified form of {@link #liveTimings} */
+    /** Number of operations of a given {@link TimingKind#name} which are currently in progress. Never persisted. */
+    transient @NonNull Map<String, LongAdder> liveCounts = new ConcurrentHashMap<>();
+    /** *  XStream simplified form of {@link #liveTimings} */
     private Map<String, Long> timings;
 
     private @NonNull Set<String> internalCalls = ConcurrentHashMap.newKeySet();
@@ -431,9 +437,11 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         private Timing(TimingKind kind) {
             this.kind = kind;
             start = System.nanoTime();
+            liveCounts.computeIfAbsent(kind.name(), k -> new LongAdder()).increment();
         }
 
         @Override public void close() {
+            liveCounts.computeIfAbsent(kind.name(), k -> new LongAdder()).decrement();
             liveTimings.computeIfAbsent(kind.name(), k -> new LongAdder()).add(System.nanoTime() - start);
         }
     }
@@ -1879,6 +1887,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                             la.add(kv.getValue());
                             return la;
                         }));
+                    result.liveCounts = new ConcurrentHashMap<>();
                     return result;
                 } catch (Exception ex) {
                     LOGGER.log(Level.SEVERE, "Failed to even load the FlowExecution", ex);
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -437,7 +437,6 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         private Timing(TimingKind kind) {
             this.kind = kind;
             start = System.nanoTime();
-            liveIncompleteTimings.add(this);
         }
 
         TimingKind getKind() {
@@ -460,7 +459,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
      * @return something to {@link Timing#close} when finished
      */
     Timing time(TimingKind kind) {
-        return new Timing(kind);
+        var timing = new Timing(kind);
+        liveIncompleteTimings.add(timing);
+        return timing;
     }
 
     static final Logger TIMING_LOGGER = Logger.getLogger(CpsFlowExecution.class.getName() + ".timing");
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -406,7 +406,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
     transient @NonNull Map<String, LongAdder> liveTimings = new ConcurrentHashMap<>();
     /** instances of {@link Timing} which have not yet completed for reporting counts and durations in support bundles. Never persisted. */
     transient @NonNull Set<Timing> liveIncompleteTimings = ConcurrentHashMap.newKeySet();
-    /** *  XStream simplified form of {@link #liveTimings} */
+    /** XStream simplified form of {@link #liveTimings} */
     private Map<String, Long> timings;
 
     private @NonNull Set<String> internalCalls = ConcurrentHashMap.newKeySet();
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -38,6 +39,9 @@
  */
 class CpsVmExecutorService extends InterceptingExecutorService {
 
+    private static final String QUEUE_SIZE_METRIC = "jenkins.job.pipeline.cps-vm-thread.queue.size";
+    private static final String EXECUTION_DURATION_METRIC = "jenkins.job.pipeline.cps-vm-thread.queue.size";
```
dwnusbaum, line 43: Bad copy/paste when I created the constants:

```suggestion
    private static final String EXECUTION_DURATION_METRIC = "jenkins.job.pipeline.cps-vm-thread.execution.duration";
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -77,9 +81,11 @@ public Void call() {
 
     @Override
     protected Runnable wrap(final Runnable r) {
+        Metrics.metricRegistry().counter(QUEUE_SIZE_METRIC).inc();
         return () -> {
+            Metrics.metricRegistry().counter(QUEUE_SIZE_METRIC).dec();
```
dwnusbaum, line 86: I also considered using a `Timer` here instead of a `Counter` to be able to see how long tasks spend in the queue, but the problem is that `Timer`s only get counted when they are closed/stopped, and in the cases I have seen recently, the whole issue was that tasks were sitting in the queue for a very long time, so we need a metric that shows the problem while it is happening, not just after the blockage has been cleared.

This is also why I am not sure about trying to use `CpsFlowExecution.time` to track this data, but we could probably make it work with some API adjustments.

We could also consider using _both_ a `Counter` and a `Timer`.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java
```diff
@@ -77,9 +81,11 @@ public Void call() {
...
             ThreadContext context = setUp();
-            try {
+            try (var timer = Metrics.metricRegistry().timer(EXECUTION_DURATION_METRIC).time()) {
```
dwnusbaum, line 88: Here, I think the `Timer` is fine. We mainly want to see if tasks are on average taking a very long time, for example because of poor IO performance or something like that. I am also not sure though if the global metric offers significant benefit over the per-build timings, so perhaps this should be dropped.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -435,7 +435,9 @@ public void onFailure(Throwable t) {
                 }
             });
         } catch (IOException x) {
-            LOGGER.log(Level.FINE, null, x);
+            // TODO: If the problem is with the FlowNode and not the CpsFlowExecution, should we try to call
+            // CpsVmExecutorService.reportProblem or CpsFlowExecution.croak to kill the build right away?
+            LOGGER.log(Level.WARNING, "Unable to load FlowNode or CpsFlowExecution when completing " + this + ", which is likely to cause its execution to hang indefinitely", x);
```
dwnusbaum, line 440: We do not have any tests in this plugin which enter this `catch` block, and the only case where I ever saw it was while manually creating broken Pipelines while investigating https://github.com/jenkinsci/workflow-api-plugin/pull/349. I think for now it is good enough to just update the log message, and if we see it come up when investigating stuck Pipelines in the future, then it might be worth investigating further.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -400,6 +404,8 @@ enum TimingKind {
 
     /** accumulated time in ns of a given {@link TimingKind#name}; {@link String} key for pretty XStream form */
     transient @NonNull Map<String, LongAdder> liveTimings = new ConcurrentHashMap<>();
+    /** instances of {@link Timing} which have not yet completed for reporting counts and durations in support bundles. Never persisted. */
+    transient @NonNull Set<Timing> liveIncompleteTimings = ConcurrentHashMap.newKeySet();
```
dwnusbaum, line 408: I considered a few different approaches, like using `metrics` or refactoring `liveTimings`, but in the end this seemed like the best way to report useful information about ongoing operations while keeping things simple and similar to what we already do. I don't see any benefit to persisting this data since it only applies to actively running tasks - once the build completes it should be empty, barring any cleanup-related bugs.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
```diff
@@ -108,13 +118,47 @@ public CpsThreadDump getThreadDump() {
 
         @Override public void addContents(Container container) {
             container.add(new Content("nodes/master/pipeline-thread-dump.txt") {
```
dwnusbaum, line 120: Is it worth updating the display name and file name to be more generic, and/or moving this to a distinct class, or should we just keep it? Users may be used to the current display name, and there may be some CloudBees-internal automation which expects the current file name, but I am not sure.

jglick: Unclear to me how this related to the existing `pipeline-timings.txt`. Should we consolidate?

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
```diff
@@ -108,13 +118,47 @@ public CpsThreadDump getThreadDump() {
...
+                                pw.println();
+                                continue;
+                            }
+                            pw.println("Build: " + ownerExec);
+                            if (ownerExec instanceof Run<?, ?>) {
+                                var run = (Run<?, ?>) ownerExec;
+                                var started = Instant.ofEpochMilli(run.getStartTimeInMillis());
+                                pw.println("Started: " + started);
+                                var duration = Duration.between(started, Instant.now());
+                                pw.print("Duration: " + duration);
```
jglick, line 140: Not very legible; consider `Util.getTimeSpanString` (IIRC).

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
dwnusbaum: One is for completed builds, one is for running builds. If I was going to rename both of them, I would do this:
* "Timing data about recently completed Pipeline builds"/`pipeline-timings.txt` would become "Recently completed Pipeline builds"/`pipeline-recent-builds.txt`
* "Thread dumps of running Pipeline builds"/`pipeline-thread-dump.txt` would become "Running Pipeline builds"/`pipeline-running-builds.txt`

We could combine them into a single "Pipeline builds"/`pipelines.txt` component, but I think keeping running and completed builds seems clearest.

jglick: > I would do this

Seems reasonable. Anyway, this is fine as is, just trying to clarify.

jglick (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
dwnusbaum: I did some GitHub searches for usage of these exact strings and checked with CloudBees support about support bundle automation and didn't find anything problematic, so I will go ahead and update the names here.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
dwnusbaum: Yes, this is subjective (personally I like ISO 8601 in contexts like this). Mainly I did not use `getTimeSpanString` to avoid localization.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
dwnusbaum: See [e8af1ac](https://github.com/jenkinsci/workflow-cps-plugin/pull/916/commits/e8af1ac4ede9f3b07ae412e7747ee83ca77bfd46).

jglick (approved):
Clearer, thanks.

PR #866, opened 2024-03-06
Pause executions when terminating even when not quieting down

jglick:
I observed a large stream of errors from a controller shutting down. This was part of a Kubernetes restart so the pod would have gotten a termination signal yet Jenkins is not “quieting down”.

<details>
<summary>Redacted examples of errors</summary>

```
…	WARNING	o.j.p.w.cps.CpsVmExecutorService#reportProblem: Unexpected exception in CPS VM thread: CpsFlowExecution[Owner[…]]
org.jenkinsci.remoting.util.ExecutorServiceUtils$FatalRejectedExecutionException: Cannot execute the command org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$1@25af3d30. The executor service is shutting down
	at hudson.remoting.SingleLaneExecutorService.execute(SingleLaneExecutorService.java:116)
	at jenkins.util.InterceptingExecutorService.execute(InterceptingExecutorService.java:71)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.notifyNewHead(CpsThreadGroup.java:502)
	at org.jenkinsci.plugins.workflow.cps.FlowHead.setNewHead(FlowHead.java:158)
	at org.jenkinsci.plugins.workflow.cps.CpsBodyExecution.addBodyEndFlowNode(CpsBodyExecution.java:414)
	at org.jenkinsci.plugins.workflow.cps.CpsBodyExecution$FailureAdapter.receive(CpsBodyExecution.java:345)
	at com.cloudbees.groovy.cps.impl.ContinuationGroup.throwException(ContinuationGroup.java:211)
	at com.cloudbees.groovy.cps.impl.CastBlock$ContinuationImpl.cast(CastBlock.java:67)
	at jdk.internal.reflect.GeneratedMethodAccessor677.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at com.cloudbees.groovy.cps.impl.ContinuationPtr$ContinuationImpl.receive(ContinuationPtr.java:72)
	at com.cloudbees.groovy.cps.impl.CollectionLiteralBlock$ContinuationImpl.dispatch(CollectionLiteralBlock.java:55)
	at com.cloudbees.groovy.cps.impl.CollectionLiteralBlock$ContinuationImpl.item(CollectionLiteralBlock.java:45)
[...]

PR #923, opened 2024-08-24
Save `build.xml` during `CpsFlowExecution.suspendAll`

jglick:
I suspect this will fix a hard-to-reproduce bug in `FlowHead` deserialization. ~I have not managed to reproduce an error yet in a functional test, but~ I have found that `build.xml` including its list of heads is only getting saved rather incidentally from the asynchronous mode of `notifyListeners`, even though `FlowNode`s have gotten saved earlier and `program.dat` has been saved.

**Formerly**: If I insert some artificial delays into `notifyNewHead` and run a restart test involving `parallel`, sometimes `build.xml` is out of date after Jenkins shuts down, causing weird effects like the iota being off after restart, or block start nodes being omitted from the graph. **Now**: not sure where those artificial delays were or exactly how that was reproduced.


jglick:
Did not in fact fix the problem.

jglick:
The immediate problem was #925, but even with that addressed, there seem to be random problems finding stored `FlowNode`s, so something like this may still be necessary.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1660,7 +1660,17 @@ public static void suspendAll() {
...
+                            try {
+                                Queue.Executable exec = cpsExec.owner.getExecutable();
+                                if (exec instanceof Saveable) {
+                                    LOGGER.fine(() -> "saving " + exec);
+                                    ((Saveable) exec).save();
+                                }
+                            } catch (IOException x) {
+                                LOGGER.log(Level.WARNING, "failed to save " + cpsExec, x);
+                            }
+                        }
```
jglick, line 1673: or perhaps just
```suggestion
                        cpsExec.saveOwner();
                            try {
                                Queue.Executable exec = cpsExec.owner.getExecutable();
                                if (exec instanceof Saveable) {
                                    LOGGER.fine(() -> "saving " + exec);
                                    ((Saveable) exec).save();
                                }
                            } catch (IOException x) {
                                LOGGER.log(Level.WARNING, "failed to save " + cpsExec, x);
                            }
                        }
```
though OTOH does https://github.com/jenkinsci/workflow-cps-plugin/blob/dc9d3aefcc6a28e8b5efa430a23d5b04ae0d92d1/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L1675 not already do just this?

jglick:
I think this is complementary to #866: that suspends and saves `program.dat`, whereas this saves `build.xml` and `workflow/*.xml`.

Vlatombe (approved):

dwnusbaum (approved):

PR #929, opened 2024-08-26
False positives from `LoggingInvoker.isInternal`

jglick:
Amending #228. ([CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-40343))

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -2105,6 +2105,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             container.add(new Content("nodes/master/pipeline-internal-calls.txt") {
                 @Override public void writeTo(OutputStream outputStream) throws IOException {
                     PrintWriter pw = new PrintWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8));
+                    Set<String> observed = new HashSet<>();
                     for (Job<?, ?> job : Jenkins.get().getAllItems(Job.class)) {
                         // TODO as above
                         if (job instanceof Queue.FlyweightTask) {
@@ -2114,9 +2115,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                                 if (owner != null) {
                                     FlowExecution exec = owner.getOrNull();
                                     if (exec instanceof CpsFlowExecution) {
-                                        Set<String> calls = ((CpsFlowExecution) exec).getInternalCalls();
+                                        List<String> calls = ((CpsFlowExecution) exec).getInternalCalls().stream().filter(observed::add).collect(Collectors.toList());
                                         if (!calls.isEmpty()) {
-                                            pw.println("Internal calls for " + run + ":");
+                                            pw.println("Internal calls for " + run + " not previously noted:");
                                             for (String call : calls) {
                                                 pw.println("  " + call);
                                             }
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -75,7 +87,7 @@ private static boolean isInternal(Class<?> clazz) {
         // (simply checking whether the class loader can “see”, say, jenkins/model/Jenkins.class
         // would falsely mark third-party libs bundled in Jenkins plugins)
         String name = clazz.getName();
-        if (name.startsWith("com.cloudbees.groovy.cps.")) {
+        if (name.startsWith("com.cloudbees.groovy.cps.") || name.startsWith("org.jenkinsci.plugins.workflow.cps.persistence.IteratorHack$")) {
```
jglick, line 90: Only apparently reproducible with Declarative (asserted by downstream test).

Vlatombe (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -2114,9 +2115,9 @@ public void autopersist(@NonNull FlowNode n) throws IOException {
                                 if (owner != null) {
                                     FlowExecution exec = owner.getOrNull();
                                     if (exec instanceof CpsFlowExecution) {
-                                        Set<String> calls = ((CpsFlowExecution) exec).getInternalCalls();
+                                        List<String> calls = ((CpsFlowExecution) exec).getInternalCalls().stream().filter(observed::add).collect(Collectors.toList());
                                         if (!calls.isEmpty()) {
-                                            pw.println("Internal calls for " + run + ":");
+                                            pw.println("Internal calls for " + run + " not previously noted:");
```
dwnusbaum, line 2120: Have you found the data here to be useful when analyzing support bundles only in general, or was it actually the per-job data that was helpful? 

If it is mainly useful in general I think it would be clearer to just drop the per-job listing entirely and just show all unique internal calls from all jobs. The per-job data seems like it would be too inconsistent to be useful after this change.

If the per-job data is useful, I would consider printing all of the internal calls first with an associated ID number, then printing each job with the ID numbers of associated internal calls, or something like that, so that we can track per-job calls precisely if necessary while still reducing the size of the file.

dwnusbaum (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: I do not think we need an exhaustive list of jobs or anything. In a bundle I looked at, the false positives were aggravated by being reported from thousands of jobs, resulting in a 24Mib file. But if there is no clue where a call is being used, that would also be a problem.

:thinking: what about listing all calls and then just showing one example caller?

```
staticMethod jenkins.model.Jenkins get (somefolder/somejob #123)
```

That would be enough to help you track down unwanted calls, while keeping the file compact.

dwnusbaum: Sure, that seems fine. I think it would be nice to add a further note that other builds are not listed e.g. `staticMethod jenkins.model.Jenkins get (somefolder/somejob #123 and 4 other builds)` to make it clear that information has been omitted and to give an idea of how widespread the usage is.

jglick: > and 4 other builds

Thought of that, but this would perhaps be more misleading that helpful because we are only even inspecting the last build of each job.

dwnusbaum (approved):

PR #933, opened 2024-09-16
Prevent NPEs in `CpsFlowExecution.cleanUpGlobalClassValue` to make Pipeline cleanup more robust

dwnusbaum:
I ran into two unexpected NPEs in `CpsFlowExecution.cleanUpGlobalClassValue`. I have never seen them before, and I am not sure exactly what caused them. The user in question had a very complicated Pipeline library and a very active controller, which might be relevant. In both cases, I think we just want to ignore the unexpected `null`s and continue to attempt to clean up all other values. Here are the stack traces:

```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "o" is null
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.ensureObj(UnsafeFieldAccessorImpl.java:57)
	at java.base/jdk.internal.reflect.UnsafeQualifiedObjectFieldAccessorImpl.get(UnsafeQualifiedObjectFieldAccessorImpl.java:38)
	at java.base/java.lang.reflect.Field.get(Field.java:425)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpGlobalClassValue(CpsFlowExecution.java:1455)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpLoader(CpsFlowExecution.java:1417)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpHeap(CpsFlowExecution.java:1391)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:467)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:331)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:295)
	at org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$2.call(CpsVmExecutorService.java:97)
	at hudson.remoting.SingleLaneExecutorService$1.run(SingleLaneExecutorService.java:139)
	at jenkins.security.ImpersonatingExecutorService$1.run(ImpersonatingExecutorService.java:68)
	at jenkins.util.ErrorLoggingExecutorService.lambda$wrap$0(ErrorLoggingExecutorService.java:51)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

```
java.lang.NullPointerException: Cannot invoke "java.lang.Class.getClassLoader()" because "klazz" is null
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpGlobalClassValue(CpsFlowExecution.java:1468)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpLoader(CpsFlowExecution.java:1417)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.cleanUpHeap(CpsFlowExecution.java:1391)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:467)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:331)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:295)
	at org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$2.call(CpsVmExecutorService.java:97)
	at hudson.remoting.SingleLaneExecutorService$1.run(SingleLaneExecutorService.java:139)
	at jenkins.security.ImpersonatingExecutorService$1.run(ImpersonatingExecutorService.java:68)
	at jenkins.util.ErrorLoggingExecutorService.lambda$wrap$0(ErrorLoggingExecutorService.java:51)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

In this PR I am also cleaning up old obsolete code paths related to Groovy 2.4.7 and older for clarity. You can review each commit independently if desired.

I was unable to reproduce these issues in tests, and after further review of the upstream Groovy code I have no hypothesis as to how they occurred in the first place. It still seems worth making the minor changes here to avoid NPEs though.

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1437,7 +1437,6 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             if (encounteredClasses.add(clazz)) {
                 LOGGER.finer(() -> "found " + clazz.getName());
                 Introspector.flushFromCaches(clazz);
-                cleanUpGlobalClassSet(clazz);
                 cleanUpClassHelperCache(clazz);
                 cleanUpLoader(clazz.getClassLoader(), encounteredLoaders, encounteredClasses);
             }
@@ -1464,24 +1463,15 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         Class<?> entryC = Class.forName("org.codehaus.groovy.util.AbstractConcurrentMapBase$Entry");
         Method getValueM = entryC.getMethod("getValue");
         List<Class<?>> toRemove = new ArrayList<>(); // not sure if it is safe against ConcurrentModificationException or not
-        try {
-            Field classRefF = classInfoC.getDeclaredField("classRef"); // 2.4.8+
-            classRefF.setAccessible(true);
-            for (Object entry : entries) {
-                Object classInfo = getValueM.invoke(entry);
-                if (classInfo != null) {
-                    Class<?> clazz = ((WeakReference<Class<?>>) classRefF.get(classInfo)).get();
-                    if (clazz != null) {
-                        toRemove.add(clazz);
-                    }
+        Field classRefF = classInfoC.getDeclaredField("classRef"); // 2.4.8+
+        classRefF.setAccessible(true);
+        for (Object entry : entries) {
+            Object classInfo = getValueM.invoke(entry);
+            if (classInfo != null) {
+                Class<?> clazz = ((WeakReference<Class<?>>) classRefF.get(classInfo)).get();
+                if (clazz != null) {
+                    toRemove.add(clazz);
                 }
-            }
-        } catch (NoSuchFieldException x) {
-            Field klazzF = classInfoC.getDeclaredField("klazz"); // 2.4.7-
-            klazzF.setAccessible(true);
-            for (Object entry : entries) {
-                Object value = getValueM.invoke(entry);
-                toRemove.add((Class) klazzF.get(value));
             }
         }
         Iterator<Class<?>> it = toRemove.iterator();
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1496,38 +1491,6 @@ private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader) throws
...
     }
 
-    private static void cleanUpGlobalClassSet(@NonNull Class<?> clazz) throws Exception {
-        Class<?> classInfoC = Class.forName("org.codehaus.groovy.reflection.ClassInfo"); // or just ClassInfo.class, but unclear whether this will always be there
-        Field globalClassSetF = classInfoC.getDeclaredField("globalClassSet");
-        globalClassSetF.setAccessible(true);
-        Object globalClassSet = globalClassSetF.get(null);
-        try {
-            classInfoC.getDeclaredField("classRef");
-            return; // 2.4.8+, nothing to do here (classRef is weak anyway)
```
dwnusbaum, line 1506: The whole method was obsolete.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1464,19 +1463,15 @@ private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader) throws
...
-            Field klazzF = classInfoC.getDeclaredField("klazz"); // 2.4.7-
-            klazzF.setAccessible(true);
-            for (Object entry : entries) {
-                Object value = getValueM.invoke(entry);
-                toRemove.add((Class) klazzF.get(value));
+        Field classRefF = classInfoC.getDeclaredField("classRef"); // 2.4.8+
+        classRefF.setAccessible(true);
+        for (Object entry : entries) {
+            Object classInfo = getValueM.invoke(entry);
+            if (classInfo != null) {
```
dwnusbaum, line 1470: First NPE was that this was sometimes `null`. A bit weird given we've never seen this before, but [this method and its Javadoc](https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/reflection/ClassInfo.java#L175-L189) make me think that this case is not totally unusual.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1464,19 +1463,15 @@ private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader) throws
...
+                Class<?> clazz = ((WeakReference<Class<?>>) classRefF.get(classInfo)).get();
+                if (clazz != null) {
```
dwnusbaum, line 1472: Second NPE was that the weak reference had sometimes already expired. This one seems a bit stranger, since the whole point of all of this cleanup code is that in practice these classes do not get cleaned up promptly in most cases.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: Actually no, that method removes the entire entry, so I am not sure how we ran into a case where there was a non-null entry  with a null `ClassInfo` value. See also https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/reflection/GroovyClassValuePreJava7.java#L97 and https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/reflection/GroovyClassValuePreJava7.java#L42 compared to https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/reflection/ClassInfo.java#L112-L114.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: I could not reproduce this case, even with a relatively invasive test class clearing out caches and related references manually.

jglick (approved):

PR #935, opened 2024-09-23
Configure eslint for warnings-ng to report on CI

timja:
see https://github.com/jenkinsci/plugin-pom/issues/874

Depends on https://github.com/jenkins-infra/pipeline-library/pull/883 pipeline library to add:
https://github.com/jenkinsci/jenkins/blob/master/Jenkinsfile#L175C33-L179

and https://github.com/jenkinsci/plugin-pom/pull/1016

### Testing done

Introduced lint warnings locally with both CI and not CI.
Tested with PR of pipeline library showing that reporting is correct
Tested with SNAPSHOT of parent pom showing that new opt-in method works:

```
# -Dtest for a faster build, it still runs the frontend tests
mvn test -Dtest=org.jenkins_ci.plugins.workflow.workflow_cps.InjectedTest -Dmaven.test.failure.ignore
```

### Submitter checklist


plugin/package.json
```diff
@@ -10,8 +10,9 @@
     "prod": "webpack --config webpack.prod.js",
     "start": "yarn dev --watch",
     "lint:js": "eslint src/main/js --ext js",
+    "lint:js-ci": "eslint src/main/js -f checkstyle -o target/eslint-warnings.xml --ext js",
     "mvnbuild": "yarn prod",
-    "mvntest": "yarn lint:js"
+    "mvntest": "npx if-env CI=true && yarn lint:js-ci || yarn lint:js"
```
jglick, line 15: Does this work on Windows, or is this part of the build even run on Windows?

timja: didn't work on windows:

```
[2024-09-23T15:43:25.323Z] [INFO] $ npx if-env CI=true && yarn lint:js-ci || yarn lint:js
[2024-09-23T15:43:25.323Z] [INFO] 'npx' is not recognized as an internal or external command,
[2024-09-23T15:43:25.323Z] [INFO] operable program or batch file.
```

jglick: And yet the build passes?

I think it is fine to run validation tools like these only on Linux (they are not likely to be platform-dependent), just surprised that this does not cause a build failure.

timja: yes, the next parent pom version by default would cause it to fail with this PR: https://github.com/jenkinsci/plugin-pom/pull/1008

timja: It worked without `npx`

plugin/pom.xml
```diff
@@ -44,6 +44,10 @@
         <no-test-jar>false</no-test-jar>
         <node.version>18.15.0</node.version>
         <yarn.version>1.22.19</yarn.version>
+
+        <!-- Opt in to Jenkins handling eslint results -->
+        <maven.test.failure.ignore>false</maven.test.failure.ignore>
+        <testFailureIgnore>${maven.test.failure.ignore}</testFailureIgnore>
```
jglick, line 50: Should this be in `plugin-pom`?

Safer to configure this in `build/plugins/*/configuration` than to rely on setting a generic-sounding property.

timja: This is for the opting in to say that that we have configured test/lint properly for Jenkins build reporting. It is in `plugin-pom` here: https://github.com/jenkinsci/plugin-pom/pull/1016/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R149

We could prefix it, e.g. something like:
```suggestion
        <maven.test.failure.ignore>false</maven.test.failure.ignore>
        <frontend.testFailureIgnore>${maven.test.failure.ignore}</frontend.testFailureIgnore>
```

jglick: Yes I think prefixing it would be prudent. As written it looks like it is a magic property name interpreted by some unknown mojo. But https://github.com/eirslett/frontend-maven-plugin?tab=readme-ov-file#ignoring-failure seems to imply that regular `maven.test.failure.ignore` is honored? https://github.com/eirslett/frontend-maven-plugin/blob/4a501a10677716a0e6676b603e3b93bc6bf319f1/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java#L33-L40 so why do we even need to configure the plugin specially for this?

timja: see https://github.com/jenkinsci/plugin-pom/pull/1015#issuecomment-2368535480

timja:
@jglick any thoughts on bumping to parent pom 5.0? Requires updating the minimum `jenkins.version` to 2.479.

jglick (commented):
Out of date. Sorry, do not recall seeing this before.

plugin/pom.xml
```diff
@@ -44,6 +44,10 @@
         <no-test-jar>false</no-test-jar>
         <node.version>18.15.0</node.version>
         <yarn.version>1.22.19</yarn.version>
+
+        <!-- Opt in to Jenkins handling eslint results -->
+        <maven.test.failure.ignore>false</maven.test.failure.ignore>
```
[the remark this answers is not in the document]
timja: This is a default value for local development, pipeline library will set the correct value: https://github.com/jenkins-infra/pipeline-library/blob/050aae058e71a42a75f8d7c83c63f9ea9acfb65e/vars/buildPlugin.groovy#L103

plugin/.gitignore
```diff
@@ -0,0 +1 @@
+.yarn/
```
jglick, line 1: note https://github.com/jenkinsci/workflow-cps-plugin/blob/931d23bf875de82c313c8443ad4d481c42f62e5c/.gitignore#L3-L4

jglick (approved):

jglick:
> Tested with PR of pipeline library showing that reporting is correct

To be clear, the effect of this PR would be that
* before: lint warning ⇒ `FAILURE`, have to look at build log to find it
* after: lint warning ⇒ `UNSTABLE`, maybe shown in **Files**, still blocks automatic merge

right?

timja:
> > Tested with PR of pipeline library showing that reporting is correct
> 
> To be clear, the effect of this PR would be that
> 
> * before: lint warning ⇒ `FAILURE`, have to look at build log to find it
> * after: lint warning ⇒ `UNSTABLE`, maybe shown in **Files**, still blocks automatic merge
> 
> right?

correct.

image

image

image

jglick (approved):

PR #936, opened 2024-09-25
[JENKINS-73031] Print hint that a `static` `interface` method may be to blame

jglick:
https://issues.jenkins.io/browse/JENKINS-73031?focusedId=449518&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-449518

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
@@ -439,7 +439,7 @@ public final class CpsThreadGroup implements Serializable {
                         var msg = error.getMessage();
                         if (msg != null && msg.contains("Illegal type in constant pool")) {
                             try {
-                                execution.getOwner().getListener().getLogger().println("May be JENKINS-73031: static interface methods not yet supported in Groovy");
+                                execution.getOwner().getListener().getLogger().println("May be JENKINS-73031: calling static interface methods from Groovy is not currently supported by Jenkins");
                             } catch (IOException x) {
                                 LOGGER.log(Level.WARNING, null, x);
                             }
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -435,6 +435,16 @@ private boolean run() {
                     if (fn != null) {
                         t.head.get().addAction(new ErrorAction(error));
                     }
+                    if (error instanceof VerifyError) {
+                        var msg = error.getMessage();
+                        if (msg != null && msg.contains("Illegal type in constant pool")) {
+                            try {
+                                execution.getOwner().getListener().getLogger().println("May be JENKINS-73031: static interface methods not yet supported in Groovy");
```
dwnusbaum, line 442: I would maybe rephrase this to indicate that it is more of a Jenkins problem than a Groovy problem:

```suggestion
                                execution.getOwner().getListener().getLogger().println("May be JENKINS-73031: static interface methods are not currently supported");
```

dwnusbaum (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
jglick: Good point. Or what about
```suggestion
                                execution.getOwner().getListener().getLogger().println("May be JENKINS-73031: calling static interface methods from Groovy is not currently supported by Jenkins");
```
?

dwnusbaum: Yes that looks good to me.

dwnusbaum (approved):

PR #942, opened 2024-10-13
Try `forkCount`

jglick:
https://github.com/jenkins-infra/pipeline-library/pull/707 since CI on this plugin is pretty slow.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,8 @@ https://github.com/jenkins-infra/pipeline-library/
 
 */
 buildPlugin(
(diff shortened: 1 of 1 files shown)


jglick:
https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/master/620/ took 26m, of which (not counting queue time) 16m on Linux and 24m on Windows.

https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-942/1/ took 18m, of which 7½m on Linux and 12m on Windows.

So I think this is an improvement, roughly 2× once the agent is allocated.

dwnusbaum (approved):

PR #946, opened 2024-10-19
Update appearance of editor

janfaracik:
This PR sets out to update the appearance of the pipeline editor, updating it to the Jenkins variables (such as the color palette, fonts) for consistency and making it a little larger too.

**Before**

image

**After**

image

image

**Key changes**
* Use Jenkins color palette
* Increase height from `250px` to `350px`
* Use system monospace font
* Icons and animations updated

### Submitter checklist

### DIFF
--- a/plugin/src/main/less/workflow-editor.less
+++ b/plugin/src/main/less/workflow-editor.less
@@ -159,7 +159,7 @@
 .ace_editor .ace_storage,
 .ace_editor .ace_storage.ace_type,
 .ace_editor .ace_support.ace_type {
-  color: var(--pink);
+  color: var(--purple);
 }
 
 .ace_editor .ace_keyword.ace_operator {
--- a/plugin/src/main/less/workflow-editor.less
+++ b/plugin/src/main/less/workflow-editor.less
@@ -94,7 +94,7 @@
 
 .ace_editor {
   background: var(--input-color);
-  border: var(--input-border-width, 2px) solid var(--input-border);
+  border: var(--jenkins-border-width, 2px) solid var(--input-border);
   border-radius: var(--form-input-border-radius);
   font-family: var(--font-family-mono) !important;
   font-size: var(--font-size-monospace);
--- a/plugin/src/main/less/workflow-editor.less
+++ b/plugin/src/main/less/workflow-editor.less
@@ -3,7 +3,7 @@
   position: absolute;
   bottom: 0;
   right: 0;
-  cursor: se-resize;
+  cursor: row-resize;
 
   width: 12px;
   height: 12px;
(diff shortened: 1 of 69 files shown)


daniel-beck:
I expect that nobody will find the new link location in the description. Could it be made more visible?

jtnord:
where has the help icon gone?
image

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
```diff
@@ -1,4 +0,0 @@
-<p>
```
jtnord, line 1: whilst this was not the best help - it did at least prompt users where to look?

janfaracik: Restored it.

jtnord (approved):
Code change look one to me. Have not run the pr in any circumstances context (proprietary or other.)
ATH appears to be unaffected.

Will wait to give @daniel-beck a chance to rereview

daniel-beck:
Seems fine with the link restored.

The tiny resize handle seems like quite a step back though from the nonstandard resize line at the bottom though, hitting that is fairly difficult. It seems about a quarter of the size of a normal textarea's (e.g., cron) resize handle?

Additionally there's a 1px area in the border where I can click on the control and get the new highlight/focus animation, but it's not focused afterwards. Regular textareas don't have that (nit, obviously).

plugin/src/main/less/workflow-editor.less
```diff
@@ -2,14 +2,226 @@
...
-  width: 100%;
-  height: 0.35rem;
-  cursor: row-resize;
-  opacity: 0.7;
-}
-.workflow-editor-wrapper .ui-resizable-handle:hover {
-  background-color: var(--primary, #f0f0f0);
-  opacity: 1;
+  right: 0;
+  cursor: se-resize;
```
daniel-beck, line 6: This change is inconsistent with regular textarea (e.g., schedule periodically) resize cursors at least on MacOS.

daniel-beck:
Resize handle size is a lot better now, thanks for iterating there. The 1px issue was a nit anyway so no big deal.

plugin/src/main/less/workflow-editor.less
janfaracik: I agree - was for consistency with textareas in Chrome/Safari, at least from the ones I've tried. Makes more sense anyway to have it just vertical.

daniel-beck: Ah, that explains it, I did not consider this to be browser-dependent.

daniel-beck (approved):
(Not a maintainer, just org owner.)

PR #948, opened 2024-10-22
JENKINS-73941 - Option to hide "Use Groovy Sandbox" for users without Administer permission globally in the system

jgarciacloudbees:
[JENKINS-73941](https://issues.jenkins.io/browse/JENKINS-73941)

This is partially implemented in [JENKINS-73470](https://issues.jenkins.io/browse/JENKINS-73470), where we created the new option "Hide Sandbox checkbox in pipeline jobs", but it was created in the scope of  this plugin.

The scope of this ticket is just to create the global option at the "script-security-plugin", to be used by all plugins depending on the sandbox.

Script Security Plugin PR including this change:
https://github.com/jenkinsci/script-security-plugin/pull/584

### Testing done

Created new test method covering the changes + CASC.

### Submitter checklist


plugin/pom.xml
```diff
@@ -54,6 +54,11 @@
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
+                <artifactId>script-security</artifactId>
+                <version>1384.v4cb_f8321c66b_</version>
```
car-roll, line 60:
```suggestion
                <version>1384.v4cb_f8321c66b_</version> <!-- https://github.com/jenkinsci/script-security-plugin/pull/584 -->
```

jgarciacloudbees:
Closing and reopening to run again the checks as there are some flaky failures

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -24,32 +24,58 @@
...
 import hudson.Extension;
 import hudson.ExtensionList;
 import jenkins.model.GlobalConfiguration;
 import jenkins.model.GlobalConfigurationCategory;
+import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.Symbol;
 
+/**
+* @deprecated
+ * This class hs been deprecated and we don't recommend to use it.
```
amuniz, line 40:
```suggestion
 * @deprecated
 * This class has been deprecated and its only configuration value is ignored. Do not rely on it or use it in any way.
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -24,32 +24,58 @@
...
 
     public CPSConfiguration() {
+
         load();
+        if (hideSandbox) {
+            ScriptApproval.get().setForceSandbox(hideSandbox);
+        }
+
+        //Data migration from this configuration to ScriptApproval should be done only once,
+        //so removing the config file after the previous migration
```
amuniz, line 64:
```suggestion
        // Data migration from this configuration to ScriptApproval should be done only once,
        // so removing the config file after the previous migration
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -24,32 +24,58 @@
...
     public boolean isHideSandbox() {
-        return hideSandbox;
+        return ScriptApproval.get().isForceSandbox();
     }
 
     public void setHideSandbox(boolean hideSandbox) {
         this.hideSandbox = hideSandbox;
+        ScriptApproval.get().setForceSandbox(hideSandbox);
         save();
     }
```
amuniz, line 80: I think you can remove this 2 methods and update their usages to call the `ScriptApproval` ones directly.

amuniz: Basically, the class stays just for the migration, nothing else.

plugin/src/test/resources/org/jenkinsci/plugins/workflow/cps/casc_test.yaml
```diff
@@ -0,0 +1,3 @@
+security:
+  cps:
+    hideSandbox: true
```
amuniz, line 3: Are we keeping this as a valid configuration? I think we should just release note this as a breaking change (which will likely affect nobody) and remove all the logic associated to `hideSandbox`  except what's needed for the data migration, and use the new APIs in script-security` directly when needed.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
jgarciacloudbees: Ok, will remove them.

plugin/src/test/resources/org/jenkinsci/plugins/workflow/cps/casc_test.yaml
jgarciacloudbees: Yes, if this casc script is used it will update the ScriptApproval one, but as we are removing the set method it will fail, so OK, I will remove the test related to CASC

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration/config.jelly
```diff
@@ -25,9 +25,4 @@ THE SOFTWARE.
 
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
-    <f:section title="${%Pipeline Sandbox}">
```
jglick, line 28: amending #907

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -24,33 +24,49 @@
...
+import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.Symbol;
 
+/**
+ * @deprecated
+ * This class has been deprecated and its only configuration value is ignored. Do not rely on it or use it in any way.
+ * In order to force using the system sandbox for pipelines, please use the flag
+ * org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.get().isForceSandbox
+ * or
+ * org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.get().isForceSandboxForCurrentUser
```
jglick, line 44: or
```suggestion
 * In order to force using the system sandbox for pipelines, please use {@link ScriptApproval#isForceSandbox} or {@link ScriptApproval#isForceSandboxForCurrentUser}
```

plugin/pom.xml
```diff
@@ -246,6 +251,16 @@
...
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins.configuration-as-code</groupId>
+            <artifactId>test-harness</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jglick, line 263: Unused AFAICT.

jglick:
> Created new test method covering the changes + CASC.

https://github.com/jenkinsci/workflow-cps-plugin/pull/948#discussion_r1815518209 There is no JCasC test here that I can see.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -24,33 +24,49 @@
...
 
-    public boolean isHideSandbox() {
-        return hideSandbox;
-    }
+        load();
+        if (hideSandbox) {
+            ScriptApproval.get().setForceSandbox(hideSandbox);
+        }
 
-    public void setHideSandbox(boolean hideSandbox) {
```
jglick, line 51: Did you try retaining the setter, and having it just delegate to `ScriptApproval.setForceSandbox` after printing a stack trace? I think that would retain compatibility for JCasC users as well.

plugin/pom.xml
jgarciacloudbees: We are going to reenable the CASC compatibility, so it is required for testing

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
jgarciacloudbees: Restoring CASC compatibility

jgarciacloudbees:
Hi @jglick, we have implemented the CASC compatibility, so the breaking change does not apply after the latest commit.
Can you please remove the breaking label?

plugin/pom.xml
```diff
@@ -57,7 +57,7 @@
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>script-security</artifactId>
-                <version>1388.v9907110c2f80</version> <!-- https://github.com/jenkinsci/script-security-plugin/pull/584 -->
+                <version>1392.v64f458436d13</version> <!-- https://github.com/jenkinsci/script-security-plugin/pull/584 -->
```
jglick, line 60:
```suggestion
                <version>1366.vd44b_49a_5c85c</version>
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/config/CPSConfiguration.java
```diff
@@ -24,33 +24,55 @@
...
+        if (hideSandbox) {
+            ScriptApproval.get().setForceSandbox(hideSandbox);
+        }
+
+        // Data migration from this configuration to ScriptApproval should be done only once,
+        // so removing the config file after the previous migration
+        try {
+            this.getConfigFile().delete();
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
amuniz, line 65: IIUC this will always be thrown as long as the CasC configuration keeps using the _old_ format? The first time it will work, the file is removed and everything is ok, but on any subsequent restart this will try to remove the file again?

jgarciacloudbees: This logic should not fail as the get.ConfigFile().delete() has already implemented the file check:
```
    public void delete() throws IOException {
        Files.deleteIfExists(Util.fileToPath(this.file));
    }
```

So this catch /thrown is defined to avoid marking the current method as thrown IOException, as this should never thrown any exception

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/JcascTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    public void cascHideSandBox() throws Exception {
+        assertTrue(ScriptApproval.get().isForceSandbox());
+    }
+
+    @Test
+    public void cascExport() throws Exception {
+        ConfiguratorRegistry registry = ConfiguratorRegistry.get();
+        ConfigurationContext context = new ConfigurationContext(registry);
+        CNode yourAttribute = getSecurityRoot(context).get("cps");
```
jglick, line 37: I think more interesting would be to check what is then exported from the `script-security` version.

jglick (approved):

PR #966, opened 2024-12-30
[JENKINS-75067] Unexport step bodies when completed in synchronous mode

rrjjvv:
Previously, the unexport of a step's body would only happen when scheduling the next execution (async mode).  If the step was not run async (due to having an outcome set prior to the step starting), the body was not removed until the CpsThreadGroup was finishing.  At this point the body's presence was treated as a resource leak and logged a warning.  Doing an explicit unexport for the sync case at the proper time eliminates the warnings from the fall-back cleanup.

Potential fix for https://issues.jenkins.io/browse/JENKINS-75067

I'm not sure if this solution is "correct", but it logically makes sense and all existing tests continue to pass.  Even if the proper fix is different (and/or best handled by you folks), I figured this PR could still prove valuable by providing a test case.

### Testing done

I created a new test representing the scenario that led me to creating the original issue.  There are two relevant assertions; both fail without my change, and pass after the change.  I don't think both are needed, but didn't know which would be preferred.  The assertion for the (lack of) warning log is obviously simpler and asserts what I truly cared about.  The other asserts based on state, which seemed to be the better approach, but feels a bit heavy/hacky (but it's the best I could come up with without touching unrelated code).

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
@@ -359,6 +359,12 @@ public class CpsStepContext extends DefaultStepContext { // TODO add XStream cla
      */
     private void scheduleNextRun() {
         if (syncMode) {
+            // probably rare for a legit sync step to have a body, but it's possible for a (typically) async step
+            // to be *treated* as sync due to having an outcome set prematurely (e.g. from a StepListener)
+            if (threadGroup != null && body != null) {
+                threadGroup.unexport(body);
+                body = null;
+            }
             // if we get the result set before the start method returned, then DSL.invokeMethod() will
             // plan the next action.
             return;
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
@@ -359,8 +359,9 @@ public class CpsStepContext extends DefaultStepContext { // TODO add XStream cla
      */
     private void scheduleNextRun() {
         if (syncMode) {
-            // probably rare for a legit sync step to have a body, but it's possible for a (typically) async step
-            // to be *treated* as sync due to having an outcome set prematurely (e.g. from a StepListener)
+            // probably rare for a legit sync step to have a body (unless short-circuiting execution of the body, as
+            // running a body in sync mode is not allowed), but it's possible for a (typically) async step to be
+            // *treated* as sync due to having an outcome set prematurely (e.g. from a StepListener)
             if (threadGroup != null && body != null) {
                 threadGroup.unexport(body);
                 body = null;
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/StepListenerTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/StepListenerTest.java
@@ -24,28 +24,48 @@
 
 package org.jenkinsci.plugins.workflow;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.AbortException;
+import hudson.ExtensionList;
+import hudson.model.Result;
 import hudson.model.TaskListener;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
+import org.jenkinsci.plugins.workflow.cps.CpsThreadGroup;
+import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.flow.StepListener;
+import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.junit.ClassRule;
+import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.LoggerRule;
 import org.jvnet.hudson.test.TestExtension;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.logging.Level;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.not;
 
 public class StepListenerTest {
-    @ClassRule
-    public static JenkinsRule r = new JenkinsRule();
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
 
     @ClassRule
     public static BuildWatcher buildWatcher = new BuildWatcher();
+
+    @Rule
+    public LoggerRule logger = new LoggerRule();
 
     @Issue("JENKINS-58084")
     @Test
@@ -59,7 +79,7 @@ public class StepListenerTest {
         r.assertLogContains("Step listener saw step echo", build);
     }
 
-    @TestExtension
+    @TestExtension("listener")
     public static class TestStepListener implements StepListener {
         @Override
         public void notifyOfNewStep(@NonNull Step s, @NonNull StepContext context) {
@@ -72,6 +92,50 @@ public class StepListenerTest {
             } catch (Exception e) {
                 // Don't care.
             }
+        }
+    }
+
+    @Issue("JENKINS-75067")
+    @Test
+    public void failingListener() throws Exception {
+        // Even before the fix there's only one warning logged.  Asserting zero records is probably over-stepping,
+        // but asserting just one record with our target message risks a false negative (some other unrelated message
+        // being first, and our being later).
+        logger.record(CpsThreadGroup.class, Level.WARNING).capture(10);
+        WorkflowJob job = r.createProject(WorkflowJob.class, "failingListener");
+        job.setDefinition(new CpsFlowDefinition("node {}\n", true));
+
+        WorkflowRun build = r.buildAndAssertStatus(Result.FAILURE, job);
+        r.assertLogContains("oops", build);
+        assertThat(TestFailingStepListener.get().closureCount, equalTo(0));
+        assertThat(logger.getMessages(), not(hasItem(containsString("Stale closure"))));
+    }
+
+    @TestExtension("failingListener")
+    public static class TestFailingStepListener implements StepListener, GraphListener.Synchronous {
+        int closureCount = -1;
+
+        @Override
+        public void notifyOfNewStep(@NonNull Step s, @NonNull StepContext context) {
+            context.onFailure(new AbortException("oops"));
+        }
+
+        @Override
+        public void onNewHead(FlowNode node) {
+            // this only works using a Synchronous listener, otherwise the fall-back closure cleaning
+            // will have already executed prior to receiving this event
+            if (node instanceof FlowEndNode) {
+              try {
+                closureCount = ((CpsFlowExecution) node.getExecution()).programPromise.get().closures.size();
+              }
+              catch (Exception e) {
+                throw new RuntimeException(e);
+              }
+            }
+        }
+
+        static TestFailingStepListener get() {
+            return ExtensionList.lookupSingleton(TestFailingStepListener.class);
         }
     }
 }
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/StepListenerTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/StepListenerTest.java
@@ -24,48 +24,28 @@
 
 package org.jenkinsci.plugins.workflow;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.AbortException;
-import hudson.ExtensionList;
-import hudson.model.Result;
 import hudson.model.TaskListener;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
-import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
-import org.jenkinsci.plugins.workflow.cps.CpsThreadGroup;
-import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.flow.StepListener;
-import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
-import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.junit.ClassRule;
-import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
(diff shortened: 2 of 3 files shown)


plugin/src/test/java/org/jenkinsci/plugins/workflow/StepListenerTest.java
```diff
@@ -24,29 +24,49 @@
...
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.not;
 
 public class StepListenerTest {
-    @ClassRule
-    public static JenkinsRule r = new JenkinsRule();
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
```
jglick, line 62: Generally prudent; BTW there is then no need to give test projects distinctive names, e.g. just

```java
r.createProject(WorkflowJob.class, "p")
```

is safe.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
```diff
@@ -359,6 +359,12 @@ private static boolean refersTo(Throwable t1, Throwable t2) {
      */
     private void scheduleNextRun() {
         if (syncMode) {
+            // probably rare for a legit sync step to have a body, but it's possible for a (typically) async step
```
jglick, line 362: Might be right, though TBH I am not entirely sure I know how this code works.

Besides the case you identified in the test (normally asynch block-scoped step like `node` terminated in `StepListener`), there are a couple of cases that may or may not currently occur in production code but which seem worth testing:

* A block-scoped step which throws an exception from its `start` method (rather than scheduling its body). For example, bad arguments. This is probably commonplace enough and might already have some test coverage here.
* A block-scoped step which calls `onSuccess` from `start` and then `return true` to complete in synch mode. (E.g., `retry(0) {…}` if that did not just throw an error saying that the count needed to be positive.)

It is not just rare, but illegal, for a step to return in synchronous mode from `start` if it actually _ran_ a body.

plugin/src/test/java/org/jenkinsci/plugins/workflow/StepListenerTest.java
rrjjvv: Yeah... originally it was a `@ClassRule`, so I needed the distinct project name (and followed the pattern set by the existing test).  Then I found that using test-scoped `@TestExtension` doesn't work correctly (which is obvious in hindsight), thus needing to change it to a simple `@Rule`.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
rrjjvv: On the first, I think I've seen such a test; I'll see if I can find it, though it should be trivial enough to have an (effective) duplicate for the sake of clarity.  A test for the second shouldn't be a problem.  I'll also update that comment with stronger language (around illegality of _running_ a block but claiming sync).

rrjjvv: I have the two additional tests for those scenarios, and all is well.  However, they need a home.  My initial test actually used `StepListener` and was validating the function of its documented use-case, so `StepListenerTest` was a logical home.  The two new tests have nothing to do with `StepListener`.  The most logical place would be in a test class for either `CpsThreadGroup` or `CpsStepExecution`, but neither has a dedicated test class.

Thoughts?  I'll check back later tonight, and if you haven't given a recommendation, I'll plan on reverting `StepListenerTest` and move all three to a new `CpsStepExecutionTest` class.

dwnusbaum: I do not think it matters too much where the tests go. I could see arguments for using `CpsFlowExecutionTest`, `CpsBodyExecutionTest`, `DSLTest`, `StepListenerTest`, and probably others depending on what exactly the tests are checking and where the relevant non-test logic lives. There is no real harm in creating a new test class either if you prefer.

Also, looking at code relevant to what is being discussed I am a bit surprised that we use `assert` here instead of throwing an exception or otherwise doing something that will still happen when assertions are disabled: https://github.com/jenkinsci/workflow-cps-plugin/blob/80cad0f47d7feebcfcf69155e459f3dee6d7df41/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java#L337

dwnusbaum: And FWIW, as far as the overall approach, the only thing that seems a bit unusual is the direct usage of the `threadGroup` field, which until now was only checked in methods which would trigger a load if needed. Whether this matters in any practical scenario, I am not sure. One alternative could be to handle this as a special case in `DSL` itself after we apply `StepListener`s and then check to see if the step was aborted before starting: https://github.com/jenkinsci/workflow-cps-plugin/blob/89b2432c3fd2d2aaef19a07bbf3483218cc41682/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java#L320-L327

Not sure what is clearest. The way you currently have things keeps all of the unexport logic in one place, but if the bug can only be observed with `StepListener` maybe it makes sense to do the unexport in the `else` block there.

rrjjvv: > I am a bit surprised that we use assert here

Want that changed while I'm at it, or is that more of note-to-self for you guys?

> Not sure what is clearest.

With those two new tests that were requested, I asserted the things I cared about, but didn't actually look at the behavior without my fix (i.e. if those scenarios result in the undesired warnings).  When I refactor them, I'll do so.  Eyeballing it, I think the first one ("exception thrown in `start`") will trigger the bug as well.  But I don't think your alternative approach will fix that scenario unless we put the unexport in *two* places... in the `else` block, but also in the `catch` block a few lines below.

I'll verify those assumptions, but I'll assume doing one (additional) export, (where I currently have it), is a small tiebreaker over having two unexports in `DSL`?

jglick: > Want that changed while I'm at it, or is that more of note-to-self for you guys?

More of a note-to-self I think. Seems to date at least to https://github.com/jenkinsci/pipeline-plugin/commit/efd3a640c7f47ad45052a4205fd2e1947861afeb. Since this is checking something about external (plugin) code it is incorrect to use an assertion.

rrjjvv: 1. Both test proposed by Jesse did in fact trigger the bug (and are both covered by my original fix)
2. It is possible to cover all three of those cases in `DSL` itself by moving the unexport a little lower, but would require either increasing the visibility of `CpsStepContext#body` or by making the assumption that raw closure will be stored verbatim in `CpsThreadGroup#closures`, and find/remove the `NamedArgsAndClosure#body` (relying on identity).  That seemed like a bad/limiting assumption.  But increasing the visibility, if you're willing, *does* work if put after https://github.com/jenkinsci/workflow-cps-plugin/blob/80cad0f47d7feebcfcf69155e459f3dee6d7df41/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java#L336-L341

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+        public static class Execution extends StepExecution {
+
+            public Execution(StepContext context) {
+                super(context);
+            }
+
+            @Override
+            public boolean start() throws Exception {
+                throw new RuntimeException("oops");
+            }
```
dwnusbaum, line 116: Just FYI there are some helpers that can be used to simplify implementations of `StepExecution` in tests:

```suggestion
        return StepExecutions.synchronous(context, ctx -> {
            throw new AbortException("oops");
        });
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        @TestExtension
```
dwnusbaum, line 119:
```suggestion
        @TestExtension("executionStartExceptionNotLeakClosures")
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+        public static class Execution extends StepExecution {
+            public Execution(StepContext context) {
+                super(context);
+            }
+
+            @Override
+            public boolean start() throws Exception {
+                getContext().onSuccess("hooray");
+                return true;
+            }
```
dwnusbaum, line 158:
```suggestion
            return StepExecutions.synchronous(context, ctx -> {
                return "hooray";
            });
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        @TestExtension
```
dwnusbaum, line 161:
```suggestion
        @TestExtension("executionWithBodyRunningSyncNotLeakClosures")
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+import org.jenkinsci.plugins.workflow.flow.GraphListener;
+import org.jenkinsci.plugins.workflow.flow.StepListener;
+import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
```
dwnusbaum, line 16:
```suggestion
import org.jenkinsci.plugins.workflow.steps.StepExecution;
import org.jenkinsci.plugins.workflow.steps.StepExecutions;
```

dwnusbaum (approved):

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java
dwnusbaum: Given that the new tests all trigger the bug and the current fix handles all of them in one place, I would stick with your current approach.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+            context.onFailure(new AbortException("oops"));
+        }
+    }
+
+    @Issue("JENKINS-75067")
+    @Test
+    public void executionStartExceptionNotLeakClosures() throws Exception {
+        logger.record(CpsThreadGroup.class, Level.WARNING).capture(10);
+        WorkflowJob job = r.createProject(WorkflowJob.class, "p");
+        job.setDefinition(new CpsFlowDefinition("badBlock {}\n", true));
```
jglick, line 76: to simplify
```suggestion
        job.setDefinition(new CpsFlowDefinition("badBlock {}", true));
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Issue("JENKINS-75067")
+    @Test
+    public void failingStepListenerNotLeakClosures() throws Exception {
+        // Even before the fix there's only one warning logged.  Asserting zero records is probably over-stepping,
+        // but asserting just one record with our target message risks a false negative (some other unrelated message
+        // being first, and our being later).
+        logger.record(CpsThreadGroup.class, Level.WARNING).capture(10);
+        WorkflowJob job = r.createProject(WorkflowJob.class, "p");
+        job.setDefinition(new CpsFlowDefinition("node {}\n", true));
```
jglick, line 54:
```suggestion
        job.setDefinition(new CpsFlowDefinition("node {}", true));
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
```diff
@@ -0,0 +1,203 @@
...
+        assertThat(ClosureCounter.get().closureCount, equalTo(0));
+        assertThat(logger.getMessages(), not(hasItem(containsString("Stale closure"))));
+    }
+
+    @Issue("JENKINS-75067")
+    @Test
+    public void executionWithBodyRunningSyncNotLeakClosures() throws Exception {
+        logger.record(CpsThreadGroup.class, Level.WARNING).capture(10);
+        WorkflowJob job = r.createProject(WorkflowJob.class, "p");
+        job.setDefinition(new CpsFlowDefinition("echo passthrough {}\n", true));
```
jglick, line 89: Not clear to me from reading how this even parses. Is it

```groovy
echo(passthrough {})
```

? If so, might be clearer to write e.g.
```suggestion
        job.setDefinition(new CpsFlowDefinition("def r = passthrough {}; echo r", true));
```

dwnusbaum: Yes, that is how it is interpreted. I thought about recommending the same change for clarity.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
jglick: or similarly just
```suggestion
            return StepExecutions.synchronous(context, ctx -> "hooray");
```

jglick (approved):
Looks right.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
dwnusbaum: `throw` is a statement, so the braces are needed here

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
rrjjvv: Yup, only did that because that's common in other tests.  Though it might be for the sake of console output or something.  Definitely happy to remove it, as it is an eyesore!  (I use groovy when I write my own tests, so never have to use them.)

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
rrjjvv: Good call.  I was actually going to remove the `echo` (and the return of an actual value from the step), but originally wanted the confidence of knowing the step actually executed (via build log assertion).  I simply forgot to remove them when I saw it.  Since you didn't question the value of those, I'll keep them and use the more explicit change.

jglick (approved):
(BTW GH lets you apply a batch of suggestions in a single commit.)

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsStepContextTest.java
rrjjvv: Thanks for the reminders!  I was aware of them when I originally set out to create my own plugin, but they were all to simplistic for my actual needs, so I forgot about them (and they didn't jump out at me in the existing tests I looked at).  But they are indeed perfect for these simple tests.

rrjjvv:
Ugh, sorry, I didn't see that 😄.  It's my first PR with these magic GH-specific suggestions (as opposed to textual feedback) and didn't really know what to expect.  Lesson learned.  Also, I assume it's standard practice for you guys to squash these on merge?  I didn't see any docs about your policy on that.

jglick:
> it's standard practice for you guys to squash these on merge

Depends. In this case, some commits undid effects of previous commits, so definitely.

PR #968, opened 2025-01-06
Persist `ReplayFlowFactoryAction#sandbox`

Vlatombe:
In CBCI HA scenarios, the `Queue.Item` (and its actions) can be rescheduled in another Jenkins replica. In such case it is required to persist the `sandbox` field, otherwise the reschedule will run without sandbox.

In a regular OSS Jenkins, you can hit the problem is you disable concurrent build execution and arrange for a build to be queued, then restart the controller, as demonstrated in the provided test.

### Testing done

Tested in CBCI HA context.

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
@@ -52,7 +52,7 @@ class ReplayFlowFactoryAction extends InvisibleAction implements CpsFlowFactoryA
 
     private String replacementMainScript;
     private final Map<String,String> replacementLoadedScripts;
-    private transient final boolean sandbox;
+    private final boolean sandbox;
     
     ReplayFlowFactoryAction(@NonNull String replacementMainScript, @NonNull Map<String,String> replacementLoadedScripts, boolean sandbox) {
         this.replacementMainScript = replacementMainScript;
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/Workflow2Test.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/Workflow2Test.java
@@ -1,0 +1,97 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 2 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
```diff
@@ -52,7 +52,7 @@ class ReplayFlowFactoryAction extends InvisibleAction implements CpsFlowFactoryA
 
     private String replacementMainScript;
     private final Map<String,String> replacementLoadedScripts;
-    private transient final boolean sandbox;
+    private final boolean sandbox;
```
Vlatombe, line 55: Could avoid persisting it in `build.xml` as well, though unlike `replacementMainScript` the size impact is negligible.

plugin/src/test/java/org/jenkinsci/plugins/workflow/Workflow2Test.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.RealJenkinsRule;
+
+/**
+ * Tests of pipelines that involve restarting Jenkins in the middle.
+ */
+public class Workflow2Test {
+    @Rule
+    public RealJenkinsRule story = new RealJenkinsRule();
```
jglick, line 52: Probably `JenkinsSessionRule` would suffice?

jglick (approved):
Looks good, other than questions about test style. Current Windows failures look like flakes.

jglick (approved):

plugin/src/test/java/org/jenkinsci/plugins/workflow/Workflow2Test.java
Vlatombe: Indeed. There was an issue with `RestartableJenkinsRule` but I guess it was due to that testing method. Too used to RJR.

plugin/src/test/java/org/jenkinsci/plugins/workflow/Workflow2Test.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public WorkflowJob p;
+    public WorkflowRun b;
+    public CpsFlowExecution e;
```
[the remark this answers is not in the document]
Vlatombe: Leftovers from copy/pasting indeed. Removed.

PR #972, opened 2025-01-16
Migrate from EE 8 to EE 9

basil:


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -146,8 +148,26 @@ public static class DescriptorImpl extends FlowDefinitionDescriptor {
          * @DataBoundSetters have been invoked (rather than in the @DataBoundConstructor), which is why we use Descriptor.newInstance.
          */
         @Override
+        public FlowDefinition newInstance(@NonNull StaplerRequest2 req, @NonNull JSONObject formData) throws FormException {
+            if (Util.isOverridden(FlowDefinitionDescriptor.class, getClass(), "newInstance", StaplerRequest.class, JSONObject.class)) {
```
jglick, line 152: Are there any actual known subclasses? Probably overkill.

jglick (approved):

PR #986, opened 2025-02-18
`WorkflowTest.env` failure

jglick:
[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-55704)

### DIFF
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
@@ -44,6 +44,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import jenkins.security.QueueItemAuthenticatorConfiguration;
 import org.hamcrest.MatcherAssert;
@@ -76,6 +77,8 @@ import static org.hamcrest.Matchers.containsString;
  * Tests of workflows that involve restarting Jenkins in the middle.
  */
 public class WorkflowTest extends SingleJobTestBase {
+
+    private static final Logger LOGGER = Logger.getLogger(WorkflowTest.class.getName());
 
     /**
      * Restart Jenkins while workflow is executing to make sure it suspends all right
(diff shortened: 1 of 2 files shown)


jglick:
`WorkflowTest.env` still failing here; as in https://github.com/jenkinsci/support-core-plugin/pull/624#issuecomment-2667067356 the logs look like the agent was killed in time, so not sure why.

plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
```diff
@@ -64,50 +65,54 @@
...
-                p.setDefinition(new CpsFlowDefinition("semaphore 'wait'", false));
-                startBuilding();
-                SemaphoreStep.waitForStart("wait/1", b);
-                assertTrue(b.isBuilding());
-                liveness();
-            }
+    @Test public void demo() throws Throwable {
+        rr.then(r -> {
+            var p = r.createProject(WorkflowJob.class, "demo");
+            p.setDefinition(new CpsFlowDefinition("semaphore 'wait'", false));
```
jglick, line 97: (ignore ws)

jglick:
[`eachClosure` failure](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-986/5/testReport/junit/org.jenkinsci.plugins.workflow/SerializationTest/windows_17___Build__windows_17____eachClosure/) from is about `logs\support\all_2025-02-19_02.56.20.log`, like the issues in https://github.com/jenkinsci/support-core-plugin/pull/624, though that looks like a file on the controller? [`persistEphemeralObject` failure](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-986/5/testReport/junit/org.jenkinsci.plugins.workflow/SerializationTest/windows_17___Build__windows_17____persistEphemeralObject/) refers to `agent-work-dirs\slave0\support\all_2025-02-19_02.53.31.log` which makes more sense.

plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowTest.java
```diff
@@ -64,50 +65,54 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import org.junit.Rule;
+import org.jvnet.hudson.test.JenkinsSessionRule;
 
 /**
  * Tests of workflows that involve restarting Jenkins in the middle.
  */
-public class WorkflowTest extends SingleJobTestBase {
+public class WorkflowTest {
```
jglick, line 85: Probably was not required after all, but longstanding tech debt anyway.

dwnusbaum (approved):

MarkEWaite (approved):
Thanks!

dwnusbaum (approved):

PR #988, opened 2025-02-21
More tolerant `FlowHead.readResolve`

jglick:
Allow more of the program to load before we fail


jglick:
Just useful for https://github.com/jenkinsci/workflow-support-plugin/pull/305 I guess.

PR #925, opened 2024-08-24
Clear `CpsBodyExecution.thread`

jglick:
Seems to fix a problem I had originally tried to solve with #923 but misclassified. Similar in spirit to #368 and more directly to https://github.com/jenkinsci/workflow-cps-plugin/pull/367#discussion_r458869636 before that was reverted. Did not try to modify `isLaunched()`; it is not clear under which conditions we would care. Open to other suggestions.

PR #280, opened 2019-04-05
[JENKINS-34973] RejectedAccessException recorded more reliably

jglick:
[JENKINS-34973](https://issues.jenkins-ci.org/browse/JENKINS-34973): `ScriptApproval.accessRejected` not being called reliably. Downstream of https://github.com/jenkinsci/script-security-plugin/pull/243.

PR #994, opened 2025-02-27
Warn user when forgetting `def` on a variable

jglick:
Following up #925: not only avoiding the symptom, but warning about the mistake. (Though not from `@NonCPS` methods.)

Note that this uses a distinct logging system from either #211 or #280, since in this case at the point at which the warning is printed we have easy access to a `TaskListener` (though `PropertyAccessBlock` would have additional access to a `SourceLocation` which may be helpful).

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
@@ -138,12 +138,17 @@ final class LoggingInvoker implements Invoker {
         Class<?> clazz = classOf(lhs);
         maybeRecord(clazz, () -> clazz.getName() + "." + name);
         delegate.setProperty(lhs, name, value);
-        var g = CpsThreadGroup.current();
-        if (g != null) {
-            var e = g.getExecution();
-            if (e != null) {
-                e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(findOwner(lhs).getClass().getName(), name, value.getClass().getName()));
-                assert false : "TODO look for false positives";
+        if (value != null) {
+            var owner = findOwner(lhs);
+            if (owner instanceof CpsScript) {
+                var g = CpsThreadGroup.current();
+                if (g != null) {
+                    var e = g.getExecution();
+                    if (e != null) {
+                        e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(findOwner(lhs).getClass().getName(), name, value.getClass().getName()));
+                        assert false : "TODO look for false positives";
+                    }
+                }
             }
         }
     }
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
@@ -142,11 +142,16 @@ final class LoggingInvoker implements Invoker {
             if (value != null) {
                 var owner = findOwner(lhs);
                 if (owner instanceof CpsScript) {
-                    var g = CpsThreadGroup.current();
-                    if (g != null) {
-                        var e = g.getExecution();
-                        if (e != null) {
-                            e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(owner.getClass().getSimpleName(), name, value.getClass().getSimpleName()));
+                    try {
+                        owner.getClass().getDeclaredField(name);
+                        // OK, @Field, ignore
+                    } catch (NoSuchFieldException x) {
+                        var g = CpsThreadGroup.current();
+                        if (g != null) {
+                            var e = g.getExecution();
+                            if (e != null) {
+                                e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(owner.getClass().getSimpleName(), name, value.getClass().getSimpleName()));
+                            }
                         }
                     }
                 }
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
@@ -138,14 +138,16 @@ final class LoggingInvoker implements Invoker {
         Class<?> clazz = classOf(lhs);
         maybeRecord(clazz, () -> clazz.getName() + "." + name);
         delegate.setProperty(lhs, name, value);
-        if (value != null) {
-            var owner = findOwner(lhs);
-            if (owner instanceof CpsScript) {
-                var g = CpsThreadGroup.current();
-                if (g != null) {
-                    var e = g.getExecution();
-                    if (e != null) {
-                        e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(owner.getClass().getSimpleName(), name, value.getClass().getSimpleName()));
+        if (SystemProperties.getBoolean(LoggingInvoker.class.getName() + ".fieldSetWarning", true)) {
+            if (value != null) {
+                var owner = findOwner(lhs);
+                if (owner instanceof CpsScript) {
+                    var g = CpsThreadGroup.current();
+                    if (g != null) {
+                        var e = g.getExecution();
+                        if (e != null) {
+                            e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(owner.getClass().getSimpleName(), name, value.getClass().getSimpleName()));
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -137,6 +138,18 @@ private void maybeRecord(Class<?> clazz, Supplier<String> message) {
         Class<?> clazz = classOf(lhs);
         maybeRecord(clazz, () -> clazz.getName() + "." + name);
         delegate.setProperty(lhs, name, value);
+        var g = CpsThreadGroup.current();
+        if (g != null) {
+            var e = g.getExecution();
+            if (e != null) {
+                e.getOwner().getListener().getLogger().println(Messages.LoggingInvoker_field_set(findOwner(lhs).getClass().getName(), name, value.getClass().getName()));
```
dwnusbaum, line 145: Are you intentionally warning about _all_ property sets? It seems like for the scenarios being described it would make sense to restrict this to only apply when the receiver is `WorkflowScript` or `DSL` or whatever the right class is.

jglick: Yeah that is what I am trying to figure out how to distinguish.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -137,6 +138,23 @@ private void maybeRecord(Class<?> clazz, Supplier<String> message) {
         Class<?> clazz = classOf(lhs);
         maybeRecord(clazz, () -> clazz.getName() + "." + name);
         delegate.setProperty(lhs, name, value);
+        if (value != null) {
+            var owner = findOwner(lhs);
+            if (owner instanceof CpsScript) {
```
jglick, line 143: Seems to work as desired also in libraries: assigning to an implicit field of a `vars/*.groovy` produces a warning, while assigning to a field of a named class in `src/**/*.groovy` does not.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Messages.properties
```diff
@@ -6,3 +6,8 @@ SnippetizerLink.OnlineDocsLink.displayName=Online Documentation
 SnippetizerLink.ExamplesLink.displayName=Examples Reference
 SnippetizerLink.GDSLLink.displayName=IntelliJ IDEA GDSL
 Pipeline.Syntax=Pipeline Syntax
+# TODO perhaps create a https://jenkins.io/redirect/pipeline-field-set/
+LoggingInvoker.field_set=\
+  Did you forget the `def` keyword? \
+  {0} seems to be setting a field named {1} (to a value of type {2}) \
```
dwnusbaum, line 12: For what it's worth, "WorkflowScript seems to be setting a field..." sounds strange to me. I don't really have any better recommendations though. Maybe "Something is setting WorkflowScript.xyz (to ...)", but then I would expect users to ask "what something?" and there is no easy way for us to answer that question here.

dwnusbaum (approved):

dwnusbaum:
Oh, and it might be worth adding a system property to disable this. Previously, when security fixes broke `@Field` there were lots of complaints, so for people setting lots of script fields intentionally these warning messages might be something they do not care about and will not fix.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Messages.properties
jglick: At least in the cases I know about, it is code in the named script which is setting the field. I could try to make the `SourceLocation` available here (which would give you a line number) though it would be a bit more work since that is only available in `groovy-cps` whereas the `TaskListener` is only available in the plugin.

As to the naming `WorkflowScript`, rather than say `Jenkinsfile`, it would be nice to update that, though I would worry about the implications.

dwnusbaum:
(Or I guess you could also check if the property matches a field defined on the script if `@Field` specifically is a concern)

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Messages.properties
dwnusbaum: Yes, my concern is mainly that users may not realize that `WorkflowScript` means Jenkinsfile. I agree that we should always mention the real class name, but maybe it could make sense to add `(your Jenkinsfile)` or something if the class name is `WorkflowScript`.

dwnusbaum (approved):

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Messages.properties
jonesbusy: Most simple declarative pipeline. Removed all global libraries (but I found zero reference of resolveStrategy or delgate anywhere in the code)

```groovy
pipeline {
    agent none
    stages {
        stage('Hello') {
            steps {
                echo 'Hello World'
            }
        }
    }
}
```
declarative

Looks I'm not the only one https://issues.jenkins.io/browse/JENKINS-75353

jglick: Ah, did not think to check Declarative syntax, will try to fix today.

jglick: Resolved.

vladaurosh:
This is not fully fixed.

https://issues.jenkins.io/browse/JENKINS-75361

PayBas:
@jglick I couldn't get rid of the warnings in some specific cases.

I've boiled it down to:
```groovy
pipeline {
  agent none
  
  stages {
    stage('FOO') {
      matrix {
        axes {
          axis {
            name 'OS'
            values 'Windows', 'Linux'
          }
        }
        stages {
          stage('BAR') {
            agent any
            steps {
              script {
                echo "${OS}"
              }
            }
          }
        }
      }
    }
  }
}
```
Leading to:
```
[Pipeline] Start of Pipeline
Did you forget the `def` keyword? WorkflowScript seems to be setting a field named __model__Closure_0_1201446834__ (to a value of type CpsClosure2) which could lead to memory leaks or other issues.
[Pipeline] stage
[Pipeline] { (FOO)
[Pipeline] parallel
[Pipeline] { (Branch: Matrix - OS = 'Windows')
[Pipeline] { (Branch: Matrix - OS = 'Linux')
[Pipeline] stage
...
```

PayBas:
@dwnusbaum it's not a big deal. We can ignore it. But some of our developers were wondering if they missed something somewhere, which it turns out they didn't.
So I thought I'd at least notify you guys.

Keep up the good work!

hsuram189:
Is there any way to disable these warnings, becuase in our case we need to create variables which are global and accessible in all stages and functions used.

tumaf33:
Hi,
How can we avoid all these warning? it started to appears even before we upgraded our plugin? why it happens?
is there any version that can work without throwing these warnings?

bassg0navy:
@tumaf33 you need to refactor pipelines to explicitly use the def keyword or data type (String, Map, etc) For variables used in multiple stages, set globally outside of pipeline or as environment variables.

tumaf33:
Thanks for replying @bassg0navy
In many cases I define parameters globally without def so other method in the same script (out side of node section) can access them too, and if I use def word they will be accessed only locally.
Also, can you please explain why this behavior started even though I didn't make any upgrade of the plugin? when I check it I can see it's got into version **4046.v90b_1b_9edec67** without anyone upgraded it.

Also the Update Site endpoint is empty (I've disabled it in the past to prevent such a cases).
image

I will be glad if you can provide some clarifications.

plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Messages.properties
tumaf33: @dwnusbaum @jglick @jonesbusy  - can you please explain why this was implemented, it's a nightmare for us to got over all of our scripts (there are hundreds) and lots of places we don't set def in order to set the variable available for the whole script not only locally.
With def we can't expose a global parameters to other method outside of the node {} section.

bassg0navy:
@tumaf33 im not a maintainer, just a fellow engineer who ran into this issue already. We noticed this after upgrading jenkinsci to version 2.5.0. I don't think we upgraded this plugin separately but it was included in that release. I don't have any information on the other issue you're facing.

PR #1003, opened 2025-03-20
[JENKINS-75447] Fix Snippetizer rendering of "properties" step

rrjjvv:
This should address https://issues.jenkins.io/browse/JENKINS-75447

There is no longer a `getItem()` that accepts a plain `StaplerRequest`.  In #972 the signature was changed to accept a `StaplerRequest2`.  Rather than maintain two variants, [it was preferred](https://issues.jenkins.io/browse/JENKINS-75447?focusedId=453247&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-453247) to modify the jelly to use the  new signature.

### Testing done

There are no additional automated tests; there were no existing tests around this and there wasn't much precedence for tests specific to the EE migration.

This was manually validated via running in a debugger (against Jenkins 2.502), but also with the same setup from ticket's reproduction:
```Dockerfile
FROM jenkins/jenkins:2.487
RUN echo 2.0 > /usr/share/jenkins/ref/jenkins.install.UpgradeWizard.state

COPY workflow-cps.hpi /tmp/
RUN jenkins-plugin-cli --latest=true --verbose  -p \
    workflow-cps:custom:file:///tmp/workflow-cps.hpi
RUN jenkins-plugin-cli --latest=false -p \
    workflow-multibranch:795.ve0cb_1f45ca_9a_
```
Following the steps from the ticket resulted in the same (favorable) behavior from `4009.v0089238351a_9`.

### Submitter checklist

### DIFF
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/index.jelly
@@ -40,7 +40,7 @@ THE SOFTWARE.
                 </f:block>
                 <f:section title="${%Steps}"/>
                 <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
-                <j:set var="item" value="${it.getItem(request)}"/>
+                <j:set var="item" value="${it.getItem(request2)}"/>
                 <d:taglib uri="local">
                     <d:tag name="listSteps">
                         <j:forEach var="quasiDescriptor" items="${quasiDescriptors}">


basil (approved):
Thanks for the PR @rrjjvv and sorry for the trouble. I've updated my EE 8 search script to catch cases like this, and I will try to sweep through the ecosystem next week to find other cases like this that I have missed. Since we are already on this topic: if you are interested in opening a PR to make the same change to https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/13f143ac8cd1b90c900dde69a7c2f24d8e1513b2/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly#L46 then that would be much appreciated as well. But if not, then this contribution is already a good step forward. Thanks again!

rrjjvv:
I can certainly do so a little later in the day.  Thanks!

PR #1005, opened 2025-03-25
Ignore `__model__*` variables generated by Declarative in script splitting mode

jglick:
https://issues.jenkins.io/browse/JENKINS-75445 amending #994. Do not know much about this mode but it looks frightening! If you find the need for it, better switch to Scripted.

To reproduce,

```bash
mvnd -pl pipeline-model-definition test -Dtest=BasicModelDefTest#stages300
```

after

```diff
diff --git pom.xml pom.xml
index 2a014654..00f4c26d 100644
--- pom.xml
+++ pom.xml
@@ -75,7 +75,7 @@
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
         <artifactId>bom-${jenkins.baseline}.x</artifactId>
-        <version>4228.v0a_71308d905b_</version>
+        <version>4440.v39a_9eb_b_c6b_4d</version>
         <scope>import</scope>
         <type>pom</type>
       </dependency>
```

Reproduced fix with:

```diff
diff --git pom.xml pom.xml
index 2a014654..a6628e01 100644
--- pom.xml
+++ pom.xml
@@ -75,10 +75,15 @@
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
         <artifactId>bom-${jenkins.baseline}.x</artifactId>
-        <version>4228.v0a_71308d905b_</version>
+        <version>4488.v7fe26526366e</version>
         <scope>import</scope>
         <type>pom</type>
       </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-cps</artifactId>
+        <version>999999-SNAPSHOT</version>
+      </dependency>
       <dependency>
         <groupId>org.jenkins-ci.main</groupId>
         <artifactId>jenkins-test-harness-tools</artifactId>
```

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
@@ -141,7 +141,7 @@ final class LoggingInvoker implements Invoker {
         maybeRecord(clazz, () -> clazz.getName() + "." + name);
         delegate.setProperty(lhs, name, value);
         if (SystemProperties.getBoolean(LoggingInvoker.class.getName() + ".fieldSetWarning", true)) {
-            if (value != null && !CLOSURE_METAPROPS.contains(name)) {
+            if (value != null && !CLOSURE_METAPROPS.contains(name) && /* RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION */ !name.startsWith("__model__")) {
                 var receiver = findReceiver(lhs);
                 if (receiver instanceof CpsScript) {
                     try {


dwnusbaum (approved):
FWIW as mentioned in https://github.com/jenkinsci/workflow-cps-plugin/pull/994#issuecomment-2706783033 I suspect that this applies to all uses of the `matrix` directive as well even if `RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION` is not set.

jglick:
Ran `MatrixTest#matrixPipeline` and did not see any issues. Let me check a broader set of tests…

jglick:
Apparently my new laptop is really fast because it finished already! I only get a failure in `ScriptStepTest.globalLibrarySuccessInScript` which looks legitimate: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/b15bbc93555dc35e9d09d53e312e66d6bf4e0cd7/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java#L352

dwnusbaum:
Just to clarify, I think this PR already fixes all of the cases, both regular `matrix` and `RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION`. I only meant that I think the described issue also applies to pipelines using `matrix` but not otherwise doing anything special due to [this condition](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/87d7955cb934c1e159d7822479d31b38fe3e08f1/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy#L895-L899) (note the `!`). For example, running [`MatrixTest.matrixStageDirectives`](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/b15bbc93555dc35e9d09d53e312e66d6bf4e0cd7/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/MatrixTest.java#L319) which explicitly disables `SCRIPT_SPLITTING_TRANSFORMATION` (it appears to be enabled by default in tests) against `workflow-cps` 4043.va_fb_de6a_a_8b_f5, I see a single "Did you forget the `def` keyword? ..." warning for `__model__Closure_0_2071996837__`, but the test is just running a Declarative Pipeline that uses `matrix`. With this PR that warning should go away.

PR #1008, opened 2025-04-08
If `initializeStorage` fails, option to not `createPlaceholderNodes`

jglick:
Optionally reverting part of 85e99f1519f545c8c96cad183a6be8c53affe727 (#223) which seems to be doing more harm than good. [CloudBees-internal issue](https://cloudbees.atlassian.net/browse/SECO-4412)

```
WARNING	o.j.p.w.cps.CpsFlowExecution#onLoad: Error initializing storage and loading nodes, will try to create placeholders for: CpsFlowExecution[xxx#123]
hudson.AbortException: Cannot resume build because FlowNode 1234 for FlowHead 1 could not be loaded.
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.initializeStorage(CpsFlowExecution.java:774)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.onLoad(CpsFlowExecution.java:812)
	at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowRun.getExecution(WorkflowRun.java:733)
	at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.get(WorkflowRun.java:1003)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$PipelineTimings$1.writeTo(CpsFlowExecution.java:2029)
	at PluginClassLoader for support-core//com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:496)
	at PluginClassLoader for support-core//com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:360)
	at PluginClassLoader for support-core//com.cloudbees.jenkins.support.SupportPlugin$PeriodicWorkImpl.lambda$doRun$0(SupportPlugin.java:1083)
	at java.base/java.lang.Thread.run(Thread.java:1583)
INFO	o.j.p.w.cps.CpsFlowExecution#createPlaceholderNodes: Creating placeholder flownodes for execution: CpsFlowExecution[xxx#123]
```

Unclear why `initializeStorage` failed; AFAICT the build metadata _had_ been valid, and `createPlaceholderNodes` is just messing it up permanently.

Pairs with https://github.com/jenkinsci/workflow-job-plugin/pull/522.

PR #1009, opened 2025-04-09
Deleting `SerializationTest.persistEphemeralObject`

jglick:
Very old test apparently dating to original Workflow prototype, not testing any realistic scenario that I can see. Timed out twice in a row in #1008 on Windows only https://github.com/jenkinsci/workflow-cps-plugin/pull/1008/checks?check_run_id=40279248176 where the thread dump does not show what the test is waiting for (maybe due to limitations in deprecated `RestartableJenkinsRule`, used from deprecated superclass `SingleJobTestBase`). Does not seem worth investigating and fixing.

### DIFF
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
@@ -2,7 +2,6 @@ package org.jenkinsci.plugins.workflow;
 
 import groovy.lang.Closure;
 import hudson.model.Result;
-import hudson.slaves.DumbSlave;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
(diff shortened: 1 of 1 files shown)


plugin/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
```diff
@@ -101,43 +99,6 @@ public boolean start() throws Exception {
...
-        story.addStep(new Statement() {
-            @Override public void evaluate() throws Throwable {
-                jenkins().setNumExecutors(0);
-                DumbSlave s = createSlave(story.j);
-                String nodeName = s.getNodeName();
-
-                p = jenkins().createProject(WorkflowJob.class, "demo");
-                p.setDefinition(new CpsFlowDefinition(
-                    "def s = jenkins.model.Jenkins.instance.getComputer('" + nodeName + "')\n" +
-                    "def r = s.node.rootPath\n" +
```
jglick, line 117: Seems to have been relying on pickles deprecated since https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180.

dwnusbaum (approved):

PR #1052, opened 2025-06-30
Notify `GraphListener` extensions prior to listeners added via `FlowExecution.addListener`

dwnusbaum:
See also https://github.com/jenkinsci/workflow-job-plugin/pull/548, which can only be safely merged and released after this PR.

While investigating whether I could change a synchronous `GraphListener` to an asynchronous one in https://github.com/jenkinsci/opentelemetry-plugin/issues/700, I noticed that `WorklowRun.GraphL` currently always runs _before_ `GraphListener` extensions. This means that asynchronous listeners may observe nodes that occur towards the end of the build _after_ the build has already completed, which seems undesirable.

This PR adjusts the listener notification order so that `GraphListener` extensions are handled first, and then listeners added via `FlowExecution.addListener` are notified in the reverse order that they were added. This PR also starts setting `TimingAction` directly on `FlowStartNode` so that timing is always set on all nodes for all listeners and we can clean up some logic in `WorkflowRun$GraphL`, see https://github.com/jenkinsci/workflow-job-plugin/pull/548.

(Earlier, I also considered https://github.com/jenkinsci/workflow-api-plugin/pull/410 and https://github.com/jenkinsci/workflow-job-plugin/pull/547, but those more complicated changes seem to be unnecessary if we just handle `TimingAction` for `FlowStartNode` here).

Given that the only known issue with `Workflow$GraphL` completing the build before other async listeners run is with a hypothetical change to `opentelemetry` that cannot happen for other unrelated reasons, I am not sure if we actually care about going through with this change, but I wanted to look into it just in case it matters for other listeners in the future.

### Testing done

See new automated tests.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.8</version>
+    <version>1.10</version>
   </extension>
 </extensions>
--- a/lib/pom.xml
+++ b/lib/pom.xml
@@ -50,7 +50,7 @@
         <plugin>
             <groupId>org.codehaus.mojo</groupId>
             <artifactId>exec-maven-plugin</artifactId>
-            <version>3.5.0</version>
+            <version>3.5.1</version>
             <executions>
                 <execution>
                     <phase>generate-sources</phase>
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -47,7 +47,7 @@
     "webpack-merge": "^6.0.1"
   },
   "dependencies": {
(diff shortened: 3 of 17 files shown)


dwnusbaum:
@jglick Do you think there is still value in doing this even though it doesn't actually matter for any known `GraphListener` implementations? I originally looked into this in the context of making a synchronous listener in `opentelemetry` asynchronous (https://github.com/jenkinsci/opentelemetry-plugin/issues/700#issuecomment-3013585494), but there are other unrelated problems which prevent that change.

jglick (approved):
It seems generally desirable, so if you think this is safe to do, we may as well.

dwnusbaum:
Ok, I'll try to run it through the PCT first.

PR #1061, opened 2025-07-15
Prevent `NullPointerException` in `CpsBodyExecution.FailureAdapter`

dwnusbaum:
@gbhat618 observed this error from code added recently in 5406e29e3c90829fd2981a6ed9077f4827220ba6 in a Pipeline which failed to stop after a hard kill. I will try to reproduce this case in a test.

```
java.lang.NullPointerException: Cannot invoke "org.jenkinsci.plugins.workflow.cps.ContextVariableSet.get(java.lang.Class, org.jenkinsci.plugins.workflow.cps.ContextVariableSet$ThrowingSupplier, org.jenkinsci.plugins.workflow.cps.ContextVariableSet$ThrowingSupplier)" because the return value of "org.jenkinsci.plugins.workflow.cps.CpsThread.getContextVariables()" is null
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsBodyExecution$FailureAdapter.handleError(CpsBodyExecution.java:378)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsBodyExecution$FailureAdapter.receive(CpsBodyExecution.java:352)
	at PluginClassLoader for workflow-cps//com.cloudbees.groovy.cps.impl.ThrowBlock$1.receive(ThrowBlock.java:66)
	at PluginClassLoader for workflow-cps//com.cloudbees.groovy.cps.impl.ConstantBlock.eval(ConstantBlock.java:21)
	at PluginClassLoader for workflow-cps//com.cloudbees.groovy.cps.Next.step(Next.java:83)
	at PluginClassLoader for workflow-cps//com.cloudbees.groovy.cps.Continuable.run0(Continuable.java:147)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.SandboxContinuable.access$001(SandboxContinuable.java:17)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.SandboxContinuable.run0(SandboxContinuable.java:49)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsThread.runNextChunk(CpsThread.java:180)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:437)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:345)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsThreadGroup$2.call(CpsThreadGroup.java:298)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService.lambda$wrap$4(CpsVmExecutorService.java:140)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at hudson.remoting.SingleLaneExecutorService$1.run(SingleLaneExecutorService.java:139)
	at jenkins.security.ImpersonatingExecutorService$1.run(ImpersonatingExecutorService.java:68)
	at jenkins.util.ErrorLoggingExecutorService.lambda$wrap$0(ErrorLoggingExecutorService.java:51)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$1.call(CpsVmExecutorService.java:53)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$1.call(CpsVmExecutorService.java:50)
	at org.codehaus.groovy.runtime.GroovyCategorySupport$ThreadCategoryInfo.use(GroovyCategorySupport.java:136)
	at org.codehaus.groovy.runtime.GroovyCategorySupport.use(GroovyCategorySupport.java:275)
	at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService.lambda$categoryThreadFactory$0(CpsVmExecutorService.java:50)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

### Testing done

See new automated test.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -376,13 +376,13 @@ class CpsBodyExecution extends BodyExecution {
             }
             try {
                 var contextVars = localThread.getContextVariables();
-                if (contextVars == null) {
-                    LOGGER.log(Level.FINE, () -> "Context variables already cleaned up for FlowNode" + sc.getNode().getId() + " and thread " + localThread);
-                    return t;
-                }
-                var handler = contextVars.get(FailureHandler.class, localThread::getExecution, sc::getNode);
-                if (handler != null) {
-                    return handler.handle(sc, t);
+                // CpsFlowExecution.start sets the context variables for the initial CpsThread to null, so context
+                // variables for all subsequent threads remain null until a block step adds a variable to its body.
+                if (contextVars != null) {
+                    var handler = contextVars.get(FailureHandler.class, localThread::getExecution, sc::getNode);
+                    if (handler != null) {
+                        return handler.handle(sc, t);
+                    }
                 }
             } catch (Throwable t2) {
                 t.addSuppressed(t2);
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
@@ -375,7 +375,12 @@ class CpsBodyExecution extends BodyExecution {
                 localThread = thread;
             }
             try {
-                var handler = localThread.getContextVariables().get(FailureHandler.class, localThread::getExecution, sc::getNode);
+                var contextVars = localThread.getContextVariables();
+                if (contextVars == null) {
+                    LOGGER.log(Level.FINE, () -> "Context variables already cleaned up for FlowNode" + sc.getNode().getId() + " and thread " + localThread);
+                    return t;
+                }
+                var handler = contextVars.get(FailureHandler.class, localThread::getExecution, sc::getNode);
                 if (handler != null) {
                     return handler.handle(sc, t);
                 }
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
@@ -26,7 +26,11 @@ import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
 import org.jenkinsci.plugins.workflow.steps.BodyExecution;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
+import org.jenkinsci.plugins.workflow.steps.StepExecutions;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 import org.junit.ClassRule;
 import org.junit.Rule;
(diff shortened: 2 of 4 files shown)


dwnusbaum:
I have a test that seems to reproduce the error consistently using `WorkflowRun.doTerm`. I am doing a bit more investigation just to make sure I understand what is happening.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -273,4 +277,35 @@ public static class Execution extends AbstractStepExecutionImpl {
         });
     }
 
+    @Test public void failureWithNoContextVars() throws Throwable {
+        rr.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("noOpBlock { throw new Exception('oops') }", true));
```
dwnusbaum, line 283: I think the Pipeline @gbhat618 was using just had an outermost `parallel` step, and that was enough to cause issues, since it doesn't insert any context variables. `stage` is actually fine because it uses `EnvironmentExpander`. I think various steps would have similar problems to `parallel`, such as `timeout`, but for test stability it seemed clearest to create a synthetic step in case the real steps add context variables in the future.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java
```diff
@@ -375,9 +375,14 @@ private Throwable handleError(CpsBodySubContext sc, Throwable t) {
                 localThread = thread;
             }
             try {
-                var handler = localThread.getContextVariables().get(FailureHandler.class, localThread::getExecution, sc::getNode);
-                if (handler != null) {
-                    return handler.handle(sc, t);
+                var contextVars = localThread.getContextVariables();
+                // CpsFlowExecution.start sets the context variables for the initial CpsThread to null, so context
+                // variables for all subsequent threads remain null until a block step adds a variable to its body.
```
dwnusbaum, line 380: At first, I thought the thread was somehow getting cleaned up prior to `CpsBodyExecution$FailureHandler` running due to some weird case with `WorkflowRun.doTerm` or `WorkflowRun.doKill`. After adding the log message added in `CpsThread.cleanUp` and turning up log levels, I realized that it was actually just that the context variables were always `null` for the thread in question.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
```diff
@@ -273,4 +277,35 @@ public static class Execution extends AbstractStepExecutionImpl {
...
+            var b = r.buildAndAssertStatus(Result.FAILURE, p);
+            r.assertLogNotContains("java.lang.NullPointerException: Cannot invoke \"org.jenkinsci.plugins.workflow.cps.ContextVariableSet.get", b);
```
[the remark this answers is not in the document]
dwnusbaum: Sure, could be simplified, but I kept the method call to try to make it clearer to readers what the test was about.

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
jglick, line 283: So throwing an exception from a block which does not bind any contextual variables suffices to reproduce this error? Are there any known actual steps which would serve as an example? Like `retry` with `conditions` or `timeout`?

dwnusbaum: Yes, you can use `parallel`, `timeout`, etc.

jglick: Ah OK.

dwnusbaum: And if the original exception was an `AbortException`, the NPE will not show up in the build log or Jenkins system log: You would have to look at `ErrorAction` directly to see the suppressed exceptions. It might show up in the Pipeline Steps view, not sure.

jglick (approved):

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecutionTest.java
dwnusbaum: > the requirement is that the full chain of steps up to the root must not bind any context variables. If you put stage, withEnv, etc., at top level, you won't see the issue.

This was not entirely accurate: Because exceptions get rethrown all the way to the top, you will actually see this problem if you have any block-scoped steps at top level which do not supply context variables, even if the exception is thrown in a place where context variables are available. I think you should see one NPE per step that does not supply context variables. So, given a Pipeline like this:

```
timeout(..., activity: false) {
  timeout(..., activity: false) {
    stage('test') {
      timeout(..., activity: false) {
        error 'oops'
      }
    }
  }
}
```

I would expect the build to fail with an `AbortException` from `error`, and for that `AbortException` to have 2 suppressed NPEs from the two top-level `timeout` steps. The `timeout` inside of `stage` would be fine.

PR #1066, opened 2025-07-29
Migrate from Commons Lang 2 to Commons Lang 3

basil:
Migrate from deprecated/EOL Commons Lang 2 to Commons Lang 3. This library wasn't used very much in this plugin (mostly in tests), so in `src/main` I simply rewrote usages to use standard Java Platform functionality.

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -60,7 +60,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import jenkins.model.Jenkins;
-import org.apache.commons.lang.StringUtils;
 import org.codehaus.groovy.reflection.CachedClass;
 import org.codehaus.groovy.reflection.ReflectionCache;
 import org.jenkinsci.Symbol;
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -138,7 +137,7 @@ protected Object readResolve() throws IOException {
 
     private static final String KEEP_STEP_ARGUMENTS_PROPERTYNAME = (DSL.class.getName()+".keepStepArguments");
 
-    private static boolean isKeepStepArguments = StringUtils.isEmpty(System.getProperty(KEEP_STEP_ARGUMENTS_PROPERTYNAME))
+    private static boolean isKeepStepArguments = System.getProperty(KEEP_STEP_ARGUMENTS_PROPERTYNAME) == null || System.getProperty(KEEP_STEP_ARGUMENTS_PROPERTYNAME).isEmpty()
             || Boolean.parseBoolean(System.getProperty(KEEP_STEP_ARGUMENTS_PROPERTYNAME));
```
jglick, line 141: Actually this should be replaced with `SystemProperties.getBoolean` (or `Boolean.getBoolean`).

jglick (approved):

PR #1085, opened 2025-09-15
Add ext-searchbox in workflow-editor

panicking:
Add a way to search inside workflow editor

image

There are some problem to solve, like that now as soon the select is enabled, there is no way to hide it so I can arrive to this problem

image

The problem that I'm not an expert on it. The test is a bit tricky. I usually deploy using:

mvn install -DskipTests

This generate the hpi that I upload using another session
that execute
mvn hpi:run

To test it I need to install a bit of plugin to have pipeline support with groovy

### DIFF
--- a/plugin/src/main/js/workflow-editor.js
+++ b/plugin/src/main/js/workflow-editor.js
@@ -8,6 +8,7 @@ import ace from "ace-builds/src-noconflict/ace";
 import "ace-builds/src-noconflict/ext-language_tools";
 import "ace-builds/src-noconflict/mode-groovy";
 import "ace-builds/src-noconflict/snippets/javascript";
+import "ace-builds/src-noconflict/ext-searchbox";
 
 // Import custom snippets
 import "./snippets/workflow";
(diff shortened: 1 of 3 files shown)


timja:
I ran 
```
mvn clean install -P quick-build
mvn hpi:run
```

and still get the same issue =/

image

```
java.lang.IllegalArgumentException: No such adjunct found: org/jenkinsci/plugins/workflow/cps/ext-searchbox.js
	at org.kohsuke.stapler.framework.adjunct.AdjunctManager.doDynamic(AdjunctManager.java:184)

panicking:
> I ran
> 
> ```
> mvn clean install -P quick-build
> mvn hpi:run
> ```
> 
> and still get the same issue =/
> 
> image
> ```
> java.lang.IllegalArgumentException: No such adjunct found: org/jenkinsci/plugins/workflow/cps/ext-searchbox.js
> 	at org.kohsuke.stapler.framework.adjunct.AdjunctManager.doDynamic(AdjunctManager.java:184)
> ```

@timja You neeed to manually update it after you build. You need to build generate the hpi and upload later after you run hpi:run. The only way to see the changes are really upload as package from the UI interface the hpi:run will restart again after you upload from the gui. I'm not able to see the changes in one run

timja:
😢 annoying, I thought it might be something like that, a bug in maven-hpi-plugin or stapler I guess.

panicking:
> 😢 annoying, I thought it might be something like that, a bug in maven-hpi-plugin or stapler I guess.

I spent 3 night, but I found when I have integrated monaca, that continue to give an old error :). So then I can back to my test enviroment.

PR #1086, opened 2025-09-19
Ensure Java classes loaded by a Pipeline's `GroovyClassLoader` are cleaned up when the Pipeline completes

dwnusbaum:
When using `@Grab`, the `GroovyClassLoader` for the Pipeline can load not only Groovy classes, but also Java classes. `CpsFlowExecution.cleanUpHeap` has some logic that uses `GroovyClassLoader.getLoadedClasses`, which only returns Groovy classes, not Java classes, and so that cleanup logic is incomplete.

I am not aware of any equivalent to `GroovyClassLoader.getLoadedClasses` that would provide access to loaded Java classes, so this PR switches to a reverse lookup on `ClassHelper$ClassHelperCache.classCache`, which is similar to what we already do in `cleanUpGlobalClassValue`. 

### Testing done

TODO: Downstream PR against `pipeline-groovy-lib`

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -82,6 +82,7 @@ import org.kohsuke.accmod.restrictions.NoExternalUse;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1490,30 +1494,50 @@ private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader) throws
...
+        removeClassesFromOtherClassLoaders(it, loader);
+        LOGGER.fine(() -> "cleaning up " + toRemove + " associated with " + loader);
+        Method removeM = classCache.getClass().getMethod("remove", Object.class);
+        for (Class<?> klazz : toRemove) {
+            loadedClasses.add(klazz);
+            removeM.invoke(classCache, klazz);
+        }
+    }
+
+    private static void removeClassesFromOtherClassLoaders(Iterator<Class<?>> it, ClassLoader loader) {
```
jglick, line 1530: Perhaps clearer to express via https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Collection.html#removeIf(java.util.function.Predicate)

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1490,30 +1494,50 @@ private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader) throws
...
+        while (it.hasNext()) {
+            Class<?> klazz = it.next();
+            ClassLoader encounteredLoader = klazz.getClassLoader();
+            if (encounteredLoader != loader) {
+                it.remove();
+                if (LOGGER.isLoggable(Level.FINEST)) {
+                  LOGGER.finest(() -> "ignoring " + klazz + " with loader " + encounteredLoader);
+                }
```
jglick, line 1538:
```suggestion
                LOGGER.finest(() -> "ignoring " + klazz + " with loader " + encounteredLoader);
```

I guess (already using the `Supplier` overload)

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
dwnusbaum: [ca2595e](https://github.com/jenkinsci/workflow-cps-plugin/pull/1086/commits/ca2595e8d1ae750c53858bde54e33944cf8c334d)

PR #1104, opened 2025-10-29
Fix tests using removed Groovy file

janfaracik:
https://github.com/jenkinsci/jenkins/pull/11208 replaces `main.groovy` with `main.jelly` - inadvertently breaking the tests in this plugin. As a fix I've used another Groovy file in core.

This feels rather brittle - theres a trend of moving away from Groovy to Jelly in core, and it'd be good to avoid this breakage happening again in the future. Not very knowledgeable in this area, but would it be possible to instead point to a Groovy file inside of this plugin?

### Submitter checklist


jglick:
This fails tests on _current_ core versions.

Let me take a look and see if I can find something more maintainable.

PR #1762, opened 2026-04-07
Memory leak when using `new SimpleTemplateEngine()` / `new GStringTemplateEngine()`

RanjitCS:
<!-- Please describe your pull request here. -->
SimpleTemplateEngine() and GStringTemplateEngine() default constructors use GroovyShell.class.getClassLoader() as parent that resolves to Jenkins core classloader (WebAppClassLoader), not the pipeline classloader. cleanUpHeap() starts from the pipeline script class, walks up through CleanGroovyClassLoader instances and never sees the template engine branch. Each createTemplate() call leaks a classloader. 

Existing tests (CpsFlowExecutionMemoryTest.loaderReleased, LibraryMemoryTest.loaderReleased) only track classloaders from this.getClass().getClassLoader() which gives the pipeline CleanGroovyClassLoader. Template engine classloaders are on a separate branch so  not reachable from this or its parent chain and we need reflection into the template object internals , which is what the test does                                                                                                                                                                                  
                                                                                                                                                                                                           
  So when createTemplate() compiles a script, it goes into a new GroovyClassLoader parented to WebAppClassLoader
### Testing done

<!-- Comment:
Provide a clear description of how this change was tested.
At minimum this should include proof that a computer has executed the changed lines.
Ideally this should include an automated test or an explanation as to why this change has no tests.
Note that automated test coverage is less than complete, so a successful PR build does not necessarily imply that a computer has executed the changed lines.
If automated test coverage does not exist for the lines you are changing, you must describe the scenario(s) in which you manually tested the change.
For frontend changes, include screenshots of the relevant page(s) before and after the change.
For refactoring and code cleanup changes, exercise the code before and after the change and verify the behavior remains the same.
-->

### Submitter checklist

<!--
Put an `x` into the [ ] to show you have filled the information.
The template comes from https://github.com/jenkinsci/.github/blob/master/.github/pull_request_template.md 
You can override it by creating .github/pull_request_template.md in your own repository 
-->

[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-71045)

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -166,11 +166,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.jenkins.plugins</groupId>
-      <artifactId>pipeline-groovy-lib</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <version>4.3.0</version>
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -182,6 +182,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>pipeline-groovy-lib</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>git</artifactId>
       <scope>test</scope>
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -166,6 +166,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>pipeline-groovy-lib</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <version>4.3.0</version>
@@ -179,11 +184,6 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials-binding</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>io.jenkins.plugins</groupId>
-      <artifactId>pipeline-groovy-lib</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1485,20 +1485,45 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         closeShells();
         if (scriptClass != null) {
             try {
-                cleanUpLoader(scriptClass.getClassLoader(), new HashSet<>(), new HashSet<>());
+                Set<ClassLoader> encounteredLoaders = new HashSet<>();
+                Set<Class<?>> encounteredClasses = new HashSet<>();
+                cleanUpLoader(scriptClass.getClassLoader(), encounteredLoaders, encounteredClasses);
+                cleanUpChildLoaders(encounteredLoaders, encounteredClasses);
             } catch (Exception x) {
                 LOGGER.log(Level.WARNING, "failed to clean up memory from " + owner, x);
-            }
-            try {
-                cleanUpTemplateEngineClasses();
-            } catch (Exception x) {
-                LOGGER.log(Level.WARNING, "failed to clean up template engine classes from " + owner, x);
             }
             scriptClass = null;
         } else {
             LOGGER.fine("no scriptClass");
         }
         // perhaps also set programPromise to null or a precompleted failure?
+    }
+
+    // cleans up child GroovyClassLoaders (eg template engine GCLs) that are parented to
+    // the pipeline classloader but not reachable by the main cleanUpLoader parent-chain walk
+    private static void cleanUpChildLoaders(Set<ClassLoader> encounteredLoaders, Set<Class<?>> encounteredClasses)
+            throws Exception {
+        Set<ClassLoader> childLoaders = new HashSet<>();
+        for (ClassInfo ci : ClassInfo.getAllClassInfo()) {
+            Class<?> clazz = ci.getTheClass();
+            if (clazz == null) {
+                continue;
+            }
+            ClassLoader loader = clazz.getClassLoader();
+            if (loader == null || encounteredLoaders.contains(loader)) {
+                continue;
+            }
+            for (ClassLoader parent = loader.getParent(); parent != null; parent = parent.getParent()) {
+                if (encounteredLoaders.contains(parent)) {
+                    childLoaders.add(loader);
+                    break;
+                }
+            }
+        }
+        for (ClassLoader loader : childLoaders) {
+            LOGGER.fine(() -> "cleaning up child classloader " + loader);
+            cleanUpLoader(loader, encounteredLoaders, encounteredClasses);
+        }
     }
 
     private static void cleanUpLoader(
@@ -1533,35 +1558,6 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             }
         }
         gcl.clearCache();
-    }
-
-    private static void cleanUpTemplateEngineClasses() throws Exception {
-        // collect all template engine classloaders from ClassInfo.
-        Set<ClassLoader> templateEngineLoaders = new HashSet<>();
-        for (ClassInfo ci : ClassInfo.getAllClassInfo()) {
-            Class<?> clazz = ci.getTheClass();
-            if (clazz == null) {
-                continue;
-            }
-            ClassLoader loader = clazz.getClassLoader();
-            ClassLoader parent = loader;
-            while (parent instanceof GroovyClassLoader.InnerLoader) {
-                parent = parent.getParent();
-            }
-            if (parent instanceof GroovyClassLoader && !(parent.getParent() instanceof GroovyClassLoader)) {
-                templateEngineLoaders.add(loader);
-            }
-        }
-        // now clean up each classloaders entries
-        for (ClassLoader loader : templateEngineLoaders) {
-            LOGGER.fine(() -> "cleaning up template engine classes from " + loader);
-            Set<Class<?>> loadedClasses = new HashSet<>();
-            cleanUpGlobalClassValue(loader, loadedClasses);
-            cleanUpClassHelperCache(loader, loadedClasses);
-            for (Class<?> clazz : loadedClasses) {
-                Introspector.flushFromCaches(clazz);
-            }
-        }
     }
 
     private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader, Set<Class<?>> loadedClasses)
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -124,6 +124,7 @@ import jenkins.model.Jenkins;
 import jenkins.util.SystemProperties;
 import net.jcip.annotations.GuardedBy;
 import org.codehaus.groovy.GroovyBugError;
+import org.codehaus.groovy.reflection.ClassInfo;
 import org.jboss.marshalling.Unmarshaller;
 import org.jboss.marshalling.reflect.SerializableClassRegistry;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
@@ -1488,6 +1489,11 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             } catch (Exception x) {
                 LOGGER.log(Level.WARNING, "failed to clean up memory from " + owner, x);
             }
+            try {
+                cleanUpTemplateEngineClassLoaders();
+            } catch (Exception x) {
+                LOGGER.log(Level.WARNING, "failed to clean up template engine classloaders from " + owner, x);
+            }
             scriptClass = null;
         } else {
             LOGGER.fine("no scriptClass");
@@ -1527,6 +1533,27 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
             }
         }
         gcl.clearCache();
+    }
+
+    private static void cleanUpTemplateEngineClassLoaders() throws Exception {
+        Set<GroovyClassLoader> templateEngineLoaders = new HashSet<>();
+        for (ClassInfo ci : ClassInfo.getAllClassInfo()) {
+            Class<?> clazz = ci.getTheClass();
+            if (clazz == null) {
+                continue;
+            }
+            ClassLoader loader = clazz.getClassLoader();
+            while (loader instanceof GroovyClassLoader.InnerLoader) {
+                loader = loader.getParent();
+            }
+            if (loader instanceof GroovyClassLoader && !(loader.getParent() instanceof GroovyClassLoader)) {
+                templateEngineLoaders.add((GroovyClassLoader) loader);
+            }
+        }
+        for (GroovyClassLoader gcl : templateEngineLoaders) {
+            LOGGER.fine(() -> "cleaning up template engine GroovyClassLoader " + gcl);
+            cleanUpLoader(gcl, new HashSet<>(), new HashSet<>());
+        }
     }
 
     private static void cleanUpGlobalClassValue(@NonNull ClassLoader loader, Set<Class<?>> loadedClasses)
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1490,9 +1490,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 LOGGER.log(Level.WARNING, "failed to clean up memory from " + owner, x);
             }
             try {
-                cleanUpTemplateEngineClassLoaders();
+                cleanUpTemplateEngineClasses();
             } catch (Exception x) {
-                LOGGER.log(Level.WARNING, "failed to clean up template engine classloaders from " + owner, x);
+                LOGGER.log(Level.WARNING, "failed to clean up template engine classes from " + owner, x);
             }
             scriptClass = null;
         } else {
@@ -1535,24 +1535,32 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
         gcl.clearCache();
     }
 
-    private static void cleanUpTemplateEngineClassLoaders() throws Exception {
-        Set<GroovyClassLoader> templateEngineLoaders = new HashSet<>();
+    private static void cleanUpTemplateEngineClasses() throws Exception {
+        // collect all template engine classloaders from ClassInfo.
+        Set<ClassLoader> templateEngineLoaders = new HashSet<>();
         for (ClassInfo ci : ClassInfo.getAllClassInfo()) {
             Class<?> clazz = ci.getTheClass();
             if (clazz == null) {
                 continue;
             }
             ClassLoader loader = clazz.getClassLoader();
-            while (loader instanceof GroovyClassLoader.InnerLoader) {
-                loader = loader.getParent();
+            ClassLoader parent = loader;
+            while (parent instanceof GroovyClassLoader.InnerLoader) {
+                parent = parent.getParent();
             }
-            if (loader instanceof GroovyClassLoader && !(loader.getParent() instanceof GroovyClassLoader)) {
-                templateEngineLoaders.add((GroovyClassLoader) loader);
+            if (parent instanceof GroovyClassLoader && !(parent.getParent() instanceof GroovyClassLoader)) {
+                templateEngineLoaders.add(loader);
             }
         }
-        for (GroovyClassLoader gcl : templateEngineLoaders) {
-            LOGGER.fine(() -> "cleaning up template engine GroovyClassLoader " + gcl);
-            cleanUpLoader(gcl, new HashSet<>(), new HashSet<>());
+        // now clean up each classloaders entries
+        for (ClassLoader loader : templateEngineLoaders) {
+            LOGGER.fine(() -> "cleaning up template engine classes from " + loader);
+            Set<Class<?>> loadedClasses = new HashSet<>();
+            cleanUpGlobalClassValue(loader, loadedClasses);
+            cleanUpClassHelperCache(loader, loadedClasses);
+            for (Class<?> clazz : loadedClasses) {
+                Introspector.flushFromCaches(clazz);
+            }
         }
     }
 
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -1925,8 +1925,9 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                     w,
                     context,
                     "timings",
-                    e.liveTimings.entrySet().stream().collect(Collectors.toMap(kv -> kv.getKey(), kv -> kv.getValue()
-                            .longValue())),
+                    e.liveTimings.entrySet().stream()
+                            .collect(Collectors.toMap(
+                                    kv -> kv.getKey(), kv -> kv.getValue().longValue())),
                     Map.class);
             writeChild(w, context, "internalCalls", new TreeSet<>(e.internalCalls), Set.class);
             writeChild(w, context, "sandbox", e.sandbox, Boolean.class);
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
@@ -32,9 +32,13 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import groovy.lang.Closure;
 import groovy.lang.GroovyClassLoader;
+import groovy.text.GStringTemplateEngine;
+import groovy.text.SimpleTemplateEngine;
 import hudson.ExtensionList;
 import java.util.Set;
 import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import jenkins.util.SystemProperties;
 import org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper;
 
@@ -123,7 +127,32 @@ final class LoggingInvoker implements Invoker {
     @Override
     public Object constructorCall(Class lhs, Object[] args) throws Throwable {
         maybeRecord(lhs, () -> lhs.getName() + ".<init>");
+        args = fixTemplateEngineClassLoader(lhs, args);
         return delegate.constructorCall(lhs, args);
+    }
+
+    private static final Logger LOGGER = Logger.getLogger(LoggingInvoker.class.getName());
+
+    private static final Set<Class<?>> TEMPLATE_ENGINE_CLASSES =
+            Set.of(SimpleTemplateEngine.class, GStringTemplateEngine.class);
+
+    // template engine default constructors parent their GCL to Jenkins core classloader
+    // instead of the pipeline classloader, inject the pipeline classloader for zero-arg calls
+    private static Object[] fixTemplateEngineClassLoader(Class<?> lhs, Object[] args) {
+        if (args.length != 0 || !TEMPLATE_ENGINE_CLASSES.contains(lhs)) {
+            return args;
+        }
+        CpsThreadGroup g = CpsThreadGroup.current();
+        if (g == null) {
+            return args;
+        }
+        CpsFlowExecution execution = g.getExecution();
+        if (execution == null) {
+            return args;
+        }
+        ClassLoader classLoader = execution.getShell().getClassLoader();
+        LOGGER.log(Level.FINE, "injecting pipeline classloader into {0} constructor", lhs.getName());
+        return new Object[] {classLoader};
     }
 
     @Override
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
@@ -7,6 +7,8 @@ import org.junit.Test;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.MemoryAssert;
 
+import java.lang.ref.WeakReference;
+
 public class MemoryUsageTest {
     @Rule
     public JenkinsRule j = new JenkinsRule();
@@ -16,21 +18,11 @@ public class MemoryUsageTest {
         WorkflowJob project = j.createProject(WorkflowJob.class, "test");
         project.setDefinition(new CpsFlowDefinition("""
             def text = 'Hello $name'
-            for(int i=0;i<5000;i++) { // TODO change loop upper limit to trigger the memory leak
-                def engine = new groovy.text.SimpleTemplateEngine()
-                def template = engine.createTemplate(text).make(['name': 'foo ' + i])
-                println template.toString()
-            }
+            def engine = new groovy.text.SimpleTemplateEngine()
+            def template = engine.createTemplate(text).make(['name': 'foo'])
+            println template.toString()
         """, false));
-
-        int additionalMemory = MemoryAssert.increasedMemory(
-                () -> {
-                    j.assertBuildStatusSuccess(project.scheduleBuild2(0));
-                    return null;
-                },
-                (obj, referredFrom, reference) -> !obj.getClass().getName().startsWith("SimpleTemplateEngine")
-        ).stream().mapToInt(value -> value.byteSize).sum();
-
-        assert additionalMemory == 0;
+        j.assertBuildStatusSuccess(project.scheduleBuild2(0));
+        MemoryAssert.assertGC(new WeakReference<>(this.getClass().getClassLoader()), true);
     }
 }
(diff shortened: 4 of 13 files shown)


plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.junit.After;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.LoggerRule;
+import org.jvnet.hudson.test.MemoryAssert;
+
+public class MemoryUsageTest {
```
jglick: Better give it a more specific name, since this seems to be about `TemplateEngine` specifically.

plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public static void register(Object o) {
+        ClassLoader loader = o.getClass().getClassLoader();
+        System.err.println("registering " + o + " from " + loader);
+        LOADERS.add(new WeakReference<>(loader));
+    }
+
+    @Test
+    public void simpleTemplateEngineLeaksClassLoader() throws Exception {
```
jglick: All four tests currently pass for me (and, apparently, CI). Am I missing something?

plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
```diff
@@ -0,0 +1,180 @@
...
+            %s.registerSimpleTemplate(template)
+            echo template.make(['name': 'foo']).toString()
+        """.formatted(MemoryUsageTest.class.getName()), false));
+        j.assertBuildStatusSuccess(project.scheduleBuild2(0));
+
+        assertFalse("expected at least one registered classloader", LOADERS.isEmpty());
+        clearInvocationCaches();
+        for (WeakReference<ClassLoader> loader : LOADERS) {
+            assertNotNull(
+                    "template GroovyClassLoader parented to WebAppClassLoader not reachable from pipeline classloader chain, not cleaned up by cleanUpHeap",
```
jglick: Phrase assertions in terms of what you _expect_ to happen, not what _did_ (supposedly) happen prior to a fix.

RanjitCS: [16032aa](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/16032aad9374bce3c2d5bf4deb5a529a7addfc97)

plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
RanjitCS: [16032aa](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/16032aad9374bce3c2d5bf4deb5a529a7addfc97) i changed assertion to show that that after cleanup classes are still present

plugin/src/test/java/org/jenkinsci/plugins/workflow/MemoryUsageTest.java
RanjitCS: [16032aa](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/16032aad9374bce3c2d5bf4deb5a529a7addfc97)

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -42,27 +45,34 @@ public void clearLoaders() {
     }
 
     private static final List<WeakReference<ClassLoader>> LOADERS = new ArrayList<>();
+    // strong references to classloaders to prevent GC during the test
+    private static final List<ClassLoader> STRONG_REFS = new ArrayList<>();
```
jglick: Huh? That would make the test always fail even with the purported bug fixed. Right?

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -42,27 +45,34 @@ public void clearLoaders() {
...
         Object closure = templateField.get(template);
-        Class<?> ownerClass = closure.getClass();
-        ClassLoader loader = ownerClass.getClassLoader();
-        System.err.println(
-                "registering " + ownerClass.getName() + " from " + loader + " (parent: " + loader.getParent() + ")");
+        ClassLoader loader = closure.getClass().getClassLoader();
+        while (loader instanceof GroovyClassLoader.InnerLoader) {
+            loader = loader.getParent();
+        }
+        System.err.println("registering " + closure.getClass().getName() + " from " + loader);
```
jglick: May as well add all the intermediate loaders to `LOADERS`?

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -82,20 +93,22 @@ public void simpleTemplateEngineLeaksClassLoader() throws Exception {
...
-            assertNotNull(
-                    "template GroovyClassLoader parented to WebAppClassLoader not reachable from pipeline classloader chain, not cleaned up by cleanUpHeap",
-                    loader.get());
-        }
+        GroovyClassLoader gcl = (GroovyClassLoader) STRONG_REFS.get(0);
+        List<String> leftover = java.util.Arrays.stream(gcl.getLoadedClasses())
+                .map(Class::getName)
+                .collect(java.util.stream.Collectors.toList());
+        // TODO cleanUpHeap should clear this but currently misses it; change to empty() once fixed
+        assertThat(leftover, hasItems("SimpleTemplateScript1"));
```
jglick: This is the bug you want to fix, right? So you should change it to `empty()` now. (If you merely want to file the steps to reproduce but do not wish to fix the bug now, that is fine, but in that case add `@Ignore`.)

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -104,19 +117,25 @@ public void gstringTemplateEngineLeaksClassLoader() throws Exception {
...
+        GroovyClassLoader gcl = (GroovyClassLoader) STRONG_REFS.get(0);
+        List<String> leftover = java.util.Arrays.stream(gcl.getLoadedClasses())
+                .map(Class::getName)
+                .collect(java.util.stream.Collectors.toList());
+        // TODO cleanUpHeap should clear this but currently misses it; change to empty() once fixed
+        assertThat(
+                leftover,
+                hasItems(
+                        "groovy.tmp.templates.GStringTemplateScript1",
+                        "groovy.tmp.templates.GStringTemplateScript1$_getTemplate_closure1"));
```
jglick: ditto

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -82,20 +93,22 @@ public void simpleTemplateEngineLeaksClassLoader() throws Exception {
...
-                    "template GroovyClassLoader parented to WebAppClassLoader not reachable from pipeline classloader chain, not cleaned up by cleanUpHeap",
-                    loader.get());
-        }
+        GroovyClassLoader gcl = (GroovyClassLoader) STRONG_REFS.get(0);
+        List<String> leftover = java.util.Arrays.stream(gcl.getLoadedClasses())
+                .map(Class::getName)
+                .collect(java.util.stream.Collectors.toList());
+        // TODO cleanUpHeap should clear this but currently misses it; change to empty() once fixed
+        assertThat(leftover, hasItems("SimpleTemplateScript1"));
+        STRONG_REFS.clear();
```
jglick: `@After`?

jglick (commented):
All four tests are still passing, so I am unclear on what this PR is doing.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: IIUC the assertion checks gcl.getLoadedClasses() not whether the classloader itself is gc'd . clearCache() empties the loaded classes list even when a strong reference to the classloader is held.
Without the strong ref, GC could collect the classloader between assertBuildStatusSuccess and the assertion, making it unreachable. The strong ref just keeps it alive long enough to inspect. I updated to comment to make more sense

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: [30aa7c3](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/30aa7c39467e195a6f88d1c2789dbe3bb4f65a95)

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: should have been more clear, test for now is just reproducing the problem, will work on the fix soon

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: [30aa7c3](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/30aa7c39467e195a6f88d1c2789dbe3bb4f65a95)

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
jglick: I mean the test should _fail_ when it reproduces the problem.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -0,0 +1,199 @@
...
+        """.formatted(TemplateEngineMemoryTest.class.getName()), false));
+        j.assertBuildStatusSuccess(project.scheduleBuild2(0));
+
+        assertFalse(STRONG_REFS.isEmpty());
+        clearInvocationCaches();
+        GroovyClassLoader gcl = (GroovyClassLoader) STRONG_REFS.get(0);
+        List<String> leftover = java.util.Arrays.stream(gcl.getLoadedClasses())
+                .map(Class::getName)
+                .collect(java.util.stream.Collectors.toList());
+        // TODO cleanUpHeap should clear this but currently misses it; change to empty() once fixed
```
jglick: Sorry, still not following https://github.com/jenkinsci/workflow-cps-plugin/pull/1762#discussion_r3046690970 at all.

For this kind of test what you want to do is:
* set up a `WeakReference` placeholder
* run some build which stashes something that _should_ be collected into that reference
* then assert that the reference can be GC’d

When the bug is present, the test will fail with an assertion error showing the reference chain.

`GroovyClassLoader.getLoadedClasses` is some implementation detail. What matters is that something that should have been collected is not.

RanjitCS: IIUC,  `assertGC` allocates memory in a loop until `OutOfMemoryError`. The JVM must clear all soft references before throwing OOM,  After the build, nothing holds a strong reference to the template engines `GroovyClassLoader`. The only things keeping it alive are soft references in `Groovys ClassInfo`. So assertGC clears those soft refs, GC collects it, and the test always passes.

I am trying to prove it using alternative, We check `GroovyClassLoader.getLoadedClasses()` after the build completes. `cleanUpHeap `walks from `scriptClass.getClassLoader() `up through the pipeline classloader chain, calling `clearCache()` on each GroovyClassLoader. But the template engines GCL is parented to the Jenkins core classloader which is a completely separate branch. `cleanUpHeap `never reaches it, never calls clearCache(), and the compiled `SimpleTemplateScript` classes remain in the cache

plugin/pom.xml
```diff
@@ -181,6 +181,11 @@
       <artifactId>credentials-binding</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>pipeline-groovy-lib</artifactId>
+      <scope>test</scope>
```
jglick: Do not do this please. If the bug is actually only reproducible when using a library, then add it to `pipeline-groovy-lib-plugin`, even if the fix is ultimately made in this repo.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -36,126 +36,146 @@ public class TemplateEngineMemoryTest {
...
-            def engine = new groovy.text.GStringTemplateEngine()
-            def template = engine.createTemplate(text)
-            %s.registerGStringTemplate(template)
-            echo template.make(['name': 'foo']).toString()
-        """.formatted(TemplateEngineMemoryTest.class.getName()), false));
+        project.setDefinition(new CpsFlowDefinition(
+                """
+                @Library('testLib@master') _
+                def defaultTemplateStr = 'Hello ${name}!'
+                for (int i = 0; i < 500; i++) {
```
jglick: Again—why the loop? If there is a leak, generally it should be reproducible with one iteration.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
jglick: There is an option to `assertGC` to fail when there are inappropriate soft references: i.e., when there is a soft leak that does not ultimately result in an OOME in production but does cause poor performance.

RanjitCS: Dug a little bit and afaict ClassInfo global map keys classes via `WeakReference`. After the build, nothing holds the template engine classes strongly. The GCL and its classes have only a weak ref key connecting them to any GC root. First `System.gc()` clears the weak key and it is collected instantly, regardless of heap size, soft ref policy, or the allowSoft parameter. The soft references inside ClassInfo exist but cannot be reached, so `assertGC` scanner never sees them.

I do not see any other option to verify the bug with `assertGC(ref, false)` other than `gcl.getLoadedClasses()`

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: [b0095e4](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/b0095e4d5aacbd319ea95f10ec25bc2132e00000), not needed was just trying to see what would happen

plugin/pom.xml
RanjitCS: [b0095e4](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/b0095e4d5aacbd319ea95f10ec25bc2132e00000)

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
jglick: > the soft references inside ClassInfo exist but cannot be reached

Sorry, why? What is the root reference chain?

jglick: I seem to be missing something fundamental here. One of the following should be true after a build completes regarding objects instantiated by Groovy code that ran in the build:

* There is some object which is held by at least one strong reference chain from a GC root. `assertGC` should fail and print that chain.
* There is some object which is held by at least one soft reference chain from a GC root (but no strong reference chains). `assertGC` should fail if passed the option to complain even on soft refs, and print that reference chain.
* There are no such objects: any root reference chains that might exist are weak. `assertGC` should pass in either mode, and there is no bug.

RanjitCS: We are hitting your second case. After the build, the template engine's GroovyClassLoader is held by a soft reference chain from `com.sun.beans.introspect.ClassInfo.CACHE` 
```
Class Name                                                                                                    | Shallow Heap | Retained Heap
---------------------------------------------------------------------------------------------------------------------------------------------
groovy.lang.GroovyClassLoader @ 0x7f7614fe0                                                                   |          112 |         1,368
'- delegate, parent groovy.lang.GroovyClassLoader$InnerLoader @ 0x7f7614f60                                   |          120 |         3,856
   '- <classloader> class SimpleTemplateScript15613 @ 0x7f7615550                                             |           16 |         4,136
      |- referent com.sun.beans.util.Cache$Kind$Soft @ 0x7f76480f8                                            |           40 |            40
      |  '- key com.sun.beans.util.Cache$CacheEntry @ 0x7f76480d8                                             |           32 |           264
      |     '- next com.sun.beans.util.Cache$CacheEntry @ 0x7f7647f20                                         |           32 |           528
      |        '- [3339] com.sun.beans.util.Cache$CacheEntry[4096] @ 0x7f3ee57d0                              |       16,400 |      1,60,808
      |           '- table com.sun.beans.introspect.ClassInfo$1 @ 0x7f45d1318                                 |           40 |      1,60,952
      |              '- CACHE class com.sun.beans.introspect.ClassInfo @ 0x7f45e1ee8 System Class             |            8 |      1,61,008
      |- clazz java.lang.reflect.Method @ 0x7f76481e0  public java.lang.Object SimpleTemplateScript15613.run()|           88 |           176
      '- theClass groovy.lang.MetaClassImpl @ 0x7f7679be0                                                     |          112 |        24,464
---------------------------------------------------------------------------------------------------------------------------------------------

```
`assertGC(ref, false)`can detect and prints the chain: `"Apparent soft references to groovy.lang.GroovyClassLoader: ClassInfo.globalClassValue→..."`. But without the fix the test does not fail,the implementation only calls `fail(softErr)` inside `catch (OutOfMemoryError)` and for a pure soft leak the JVM clears the soft reference just before OOM, allowing the object to be collected. With the fix the test passes in ~5 seconds,without it, it takes ~70-135 seconds before the soft ref is cleared. i am not sure if there is a way to reproduce exactly with `assertGc`, but imo there is a leak for sure

jglick: Perhaps there is a mistake in `assertGC`?

RanjitCS: yes, imo there is an error here https://github.com/jenkinsci/jenkins-test-harness/blob/bdf9405557dcf28c317224812ba87592de2edacf/src/main/java/org/jvnet/hudson/test/MemoryAssert.java#L182. What happens is if the JVM ran out of memory but the object is still alive and we already know there are soft refs it should fail. But the JVM clears weak/soft references before throwing OOM. So the weak key in the ClassInfo map gets collected first, cascading cleanup frees the GCL, reference.get() returns null, the while loop exits normally. We could change this to  `if softErr != null && !allowSoft` then fail, as object was collected only due to memory pressure, which is not the same as production.

I can make this test fail with getLoadedClasses, but there is a leak clearly, is this okay to proceed?

jglick: > imo there is an error

So, file a fix for it?

RanjitCS: fix is already present in this pr, i will move it to ready status

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -0,0 +1,99 @@
...
+        WorkflowJob p = j.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("""
+                def engine = new groovy.text.SimpleTemplateEngine()
+                def template = engine.createTemplate('Hello $name')
+                def scriptField = template.getClass().getDeclaredField('script')
+                scriptField.setAccessible(true)
+                %s.register(scriptField.get(template))
+                echo template.make([name: 'world']).toString()
+                """.formatted(TemplateEngineMemoryTest.class.getName()), false));
+        j.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick:
```suggestion
        j.buildAndAssertSuccess(p);
```

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -0,0 +1,99 @@
...
+        WorkflowJob p = j.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("""
+                def engine = new groovy.text.GStringTemplateEngine()
+                def template = engine.createTemplate('Hello $name')
+                def templateField = template.getClass().getDeclaredField('template')
+                templateField.setAccessible(true)
+                %s.register(templateField.get(template))
+                echo template.make([name: 'world']).toString()
+                """.formatted(TemplateEngineMemoryTest.class.getName()), false));
+        j.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick:
```suggestion
        j.buildAndAssertSuccess(p);
```

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1529,6 +1535,27 @@ private static void cleanUpLoader(
         gcl.clearCache();
     }
 
+    private static void cleanUpTemplateEngineClassLoaders() throws Exception {
```
jglick: This is not being passed any parameter indicating which build just completed. So it looks like it could break builds that are still running.

jglick (commented):
I am not sure I understood the upshot of https://github.com/jenkinsci/workflow-cps-plugin/pull/1762#discussion_r3071269363 but I can see that with the `CpsFlowExecution` change reverted, the test still passes, so it is not helping guard against regression.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: [7390c78](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/7390c7816a2968692d159bcd720c3f30caae419e)

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: [7390c78](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/7390c7816a2968692d159bcd720c3f30caae419e)

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
RanjitCS: [7390c78](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/7390c7816a2968692d159bcd720c3f30caae419e) and added a test to check this scenario

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -61,8 +68,12 @@ public void simpleTemplateEngineLoaderReleased() throws Exception {
                 %s.register(scriptField.get(template))
                 echo template.make([name: 'world']).toString()
                 """.formatted(TemplateEngineMemoryTest.class.getName()), false));
-        j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        j.buildAndAssertSuccess(p);
         assertFalse(LOADERS.isEmpty());
+        assertThat(
+                "template engine classes should be removed from groovy globalClassValue by cleanUpHeap",
+                findLeakedGlobalClassValueEntries(),
+                empty());
```
jglick: Would still prefer for all of these specific assertions to be deleted, and whatever change to `MemoryAssert` is needed (https://github.com/jenkinsci/workflow-cps-plugin/pull/1762#discussion_r3074358466) to ensure that the test fails if the purported fix is reverted.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
```diff
@@ -81,14 +92,97 @@ public void gstringTemplateEngineLoaderReleased() throws Exception {
...
+        WorkflowJob a = j.createProject(WorkflowJob.class, "a");
+        a.setDefinition(new CpsFlowDefinition("""
+                def engine = new groovy.text.SimpleTemplateEngine()
+                def template = engine.createTemplate('Hello $name')
+                echo template.make([name: 'world']).toString()
+                """, false));
+        j.buildAndAssertSuccess(a);
+
+        // release build b and verify it completes successfully
+        // if the cleanup broke build b template engine, this would fail
```
jglick: Why? The test is pausing in between uses of `SimpleTemplateEngine`, so it is not in fact simulating cleanups to live template engine loaders happening in different native threads. (All operations in a given build occur in a single native thread: the CPS VM uses a “green thread” model. But operations in different builds can certainly occur simultaneously.)

In fact this test does not fail when the `src/main/` part of 7390c7816a2968692d159bcd720c3f30caae419e is reverted.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1535,24 +1535,32 @@ private static void cleanUpLoader(
         gcl.clearCache();
     }
 
-    private static void cleanUpTemplateEngineClassLoaders() throws Exception {
-        Set<GroovyClassLoader> templateEngineLoaders = new HashSet<>();
+    private static void cleanUpTemplateEngineClasses() throws Exception {
```
jglick: This is still not capturing any information about the specific recently completed build, so I think https://github.com/jenkinsci/workflow-cps-plugin/pull/1762#discussion_r3074576934 still applies and advise finding some other way to determine which `templateEngineLoaders` are actually in scope. Unless 7390c7816a2968692d159bcd720c3f30caae419e claims to be performing solely harmless operations (clearing caches) which could (by code inspection—you are not going to be able to test this) not possibly harm any ongoing builds?

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -1529,6 +1535,35 @@ private static void cleanUpLoader(
...
+            if (clazz == null) {
+                continue;
+            }
+            ClassLoader loader = clazz.getClassLoader();
+            ClassLoader parent = loader;
+            while (parent instanceof GroovyClassLoader.InnerLoader) {
+                parent = parent.getParent();
+            }
+            if (parent instanceof GroovyClassLoader && !(parent.getParent() instanceof GroovyClassLoader)) {
+                templateEngineLoaders.add(loader);
```
jglick: Am I missing something? I do not see anything in this logic which would indicate that it is collecting only loaders relating to `TemplateEngine`. It seems to be just processing everything used by Groovy, which sounds quite dangerous.

jglick:
Going back to the start:

> `SimpleTemplateEngine()` and `GStringTemplateEngine()` default constructors use `GroovyShell.class.getClassLoader()` as parent that resolves to Jenkins core classloader (`WebAppClassLoader`), not the pipeline classloader.

Perhaps it is that we should be fixing. Does the leak still occur if user pipelines wrote

```groovy
def te = new SimpleTemplateEngine(getClass().classLoader)
```

? If not, perhaps we could special-case these classes in https://github.com/jenkinsci/workflow-cps-plugin/blob/8df38f05c3c56b0772851985ddd333aada8c8be2/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java#L97-L98 or similar.

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: i removed this test compeletely in [51cc390](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/51cc39055f71e1126183d9daa3bf1c5126addf80) as the changes there is that so cleanup is now build-scoped via encounteredLoaders from the completed builds cleanUpLoader walk, so it cannot affect other builds by design

plugin/src/test/java/org/jenkinsci/plugins/workflow/TemplateEngineMemoryTest.java
RanjitCS: [51cc390](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/51cc39055f71e1126183d9daa3bf1c5126addf80), after i check the impact of https://github.com/jenkinsci/jenkins-test-harness/pull/1029#issuecomment-4251893596 this will fail without the fix, tested locally

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
RanjitCS: [51cc390](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/51cc39055f71e1126183d9daa3bf1c5126addf80) cleanUpChildLoaders takes encounteredLoaders from the completed builds cleanup walk, only processes child GCLs of that builds classloaders

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
RanjitCS: [51cc390](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/51cc39055f71e1126183d9daa3bf1c5126addf80) only collects classloaders whose parent chain includes a loader already encountered during this builds cleanup, not all groovy classloaders

RanjitCS:
> Going back to the start:
> 
> > `SimpleTemplateEngine()` and `GStringTemplateEngine()` default constructors use `GroovyShell.class.getClassLoader()` as parent that resolves to Jenkins core classloader (`WebAppClassLoader`), not the pipeline classloader.
> 
> Perhaps it is that we should be fixing. Does the leak still occur if user pipelines wrote
> 
> ```groovy
> def te = new SimpleTemplateEngine(getClass().classLoader)
> ```
> 
> ? If not, perhaps we could special-case these classes in
> 
> https://github.com/jenkinsci/workflow-cps-plugin/blob/8df38f05c3c56b0772851985ddd333aada8c8be2/lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java#L97-L98
> 
> or similar.

[51cc390](https://github.com/jenkinsci/workflow-cps-plugin/pull/1762/commits/51cc39055f71e1126183d9daa3bf1c5126addf80) LoggingInvoker.constructorCall intercepts zero-arg template engine constructors and injects the pipeline classloader, cleanUpChildLoaders is still needed alongside to break the ClassInfo referencechain for the child GCLs

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/LoggingInvoker.java
```diff
@@ -123,9 +127,34 @@
...
     }
 
+    private static final Logger LOGGER = Logger.getLogger(LoggingInvoker.class.getName());
+
+    private static final Set<Class<?>> TEMPLATE_ENGINE_CLASSES =
+            Set.of(SimpleTemplateEngine.class, GStringTemplateEngine.class);
+
+    // template engine default constructors parent their GCL to Jenkins core classloader
+    // instead of the pipeline classloader, inject the pipeline classloader for zero-arg calls
+    private static Object[] fixTemplateEngineClassLoader(Class<?> lhs, Object[] args) {
```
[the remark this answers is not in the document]
RanjitCS: this is a false positive imo, we only call getShell().getClassLoader() to get the pipeline classloader reference, the shell is not stored or used for code execution, same pattern as existing calls to getShell() throughout CpsFlowExecution

jglick (approved):
Looks right. Just need to test & ship the JTH bump (override via `jenkins-test-harness.version` pending `plugin-pom` bump) so that the test here is active.

jglick:
(updated PR summary to what IIUC the bug consists of)

RanjitCS:
Tested with https://github.com/jenkinsci/jenkins-test-harness/releases/tag/2565.vd1eb_7c961d1b_, the test fails without the fix and this test-harness version and passes with the fix as well.

jglick:
Great, so just temporarily override the JTH version here with a TODO comment reminding maintainers to delete the override when a newer parent includes it.

timja:
> Great, so just temporarily override the JTH version here with a TODO comment reminding maintainers to delete the override when a newer parent includes it.

2 clicks here to trigger the release 😆, triggered for you: https://github.com/jenkinsci/plugin-pom/actions/workflows/cd.yaml

timja:
https://github.com/jenkinsci/plugin-pom/releases/tag/6.2153.vcf31911d10c4

jglick:
Oh thanks, I am still used to releases being a laborious process that happens every few weeks! Thanks @timja, and @RanjitCS you should be able to just delete the override and bump the parent (just recheck that the test fails with `src/main/` reverted).

RanjitCS:
> Oh thanks, I am still used to releases being a laborious process that happens every few weeks! Thanks @timja, and @RanjitCS you should be able to just delete the override and bump the parent (just recheck that the test fails with `src/main/` reverted).

Done, confirmed that with the parent bump. Without fix test fail and with fix test passes

PR #1765, opened 2026-04-13
Block `pipeline {…}` when `pipeline-model-definition` not enabled

jglick:
Addresses the most complaint about #1740 without handling the general case.

[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-31174)

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
@@ -145,6 +145,11 @@ public class DSL extends GroovyObjectSupport implements Serializable {
         return isKeepStepArguments;
     }
 
+    /** Pending full #1740 impl, particular symbols we know should not be run as functions. */
+    private static final Set<String> BLACKLISTED_SYMBOLS = Set.of(
+            // ignore WorkflowJob.DescriptorImpl; only interpret ModelStepLoader
+            "pipeline");
+
     /**
      * Executes the {@link Step} implementation specified by the name argument.
      *
@@ -193,7 +198,7 @@ public class DSL extends GroovyObjectSupport implements Serializable {
             }
             return invokeStep(sd, name, args);
         }
-        if (SymbolLookup.get().findDescriptor(Describable.class, name) != null) {
+        if (!BLACKLISTED_SYMBOLS.contains(name) && SymbolLookup.get().findDescriptor(Describable.class, name) != null) {
             return invokeDescribable(name, args);
         }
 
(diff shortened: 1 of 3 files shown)


jglick:
[`ReplayActionTest.rebuildNeedScriptApproval` failure](https://github.com/jenkinsci/workflow-cps-plugin/pull/1765/checks?check_run_id=71149966659) :eyes:

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -473,23 +473,15 @@ public void evaluate() throws Throwable {
                         .to("dev1"));
 
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "SECURITY-3362");
-                String script = "pipeline {\n" + "  agent any\n"
-                        + "  stages {\n"
-                        + "    stage('List Jobs') {\n"
-                        + "      steps {\n"
-                        + "        script {\n"
-                        + "           println \"Jobs: ${jenkins.model.Jenkins.instance.getItemByFullName(env.JOB_NAME)?.parent?.items*.fullName.join(', ')}!\""
```
jglick, line 481: 3e20a37282f82dff2470a8d5af12151a29a4e4fa. This line was never actually run! Yet the test passed anyway.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -145,6 +145,11 @@ public static boolean isKeepStepArguments() {
         return isKeepStepArguments;
     }
 
+    /** Pending full #1740 impl, particular symbols we know should not be run as functions. */
```
gbhat618, line 148: quick link https://github.com/jenkinsci/workflow-cps-plugin/issues/1740

gbhat618 (approved):
lgtm. I don't have much experience with this plugin, but the change understandable to me,
* don't dispatch `pipeline` (or any blocked symbol found) if the plugin is not there.
* don't show the blocked symbol in the suggestions -- but the `pipeline` step will appear from the actual plugin if it is installed https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/a98ed49400ef757b6d5dc4cbed3acd602f4c8e9f/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java#L44

jglick:
> the `pipeline` step

Technically implemented as a global variable, but yes otherwise that is accurate.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java
```diff
@@ -145,6 +145,11 @@ public static boolean isKeepStepArguments() {
...
+    private static final Set<String> BLACKLISTED_SYMBOLS = Set.of(
```
daniel-beck, line 149: Please note https://jenkins.io/jep/16 for new code.

PR #1766, opened 2026-04-21
Remove the confusing second Rebuild button when Rebuilder plugin can do the job (better)

jimklimov:
https://issues.jenkins.io/browse/JENKINS-76292

> Non-admin users see two "Rebuild" buttons on Pipeline builds when the Rebuilder plugin is installed.
> This does not happen for admin users, who instead see Replay + Rebuild, which is expected.

The PR extends `ReplayAction::isRebuildEnabled()` with a check for `rebuild` plugin availability, and a reflection call to see if its `com.sonyericsson.rebuild.AbstractRebuildAction.isRebuildAvailable()` would render the link. 

In this case, the workflow-cps-plugin should not show a duplicate and potentially inferior (no Parameterized Rebuild concept here) fallback link when Replay is not permitted.

### Testing done

Not yet, checking.

Would be visual/interactive on a Jenkins controller with rebuildable pipelines.

Further PRs are welcome to add an automated test case with installed `rebuild` (Rebuilder) plugin, but only for some test cases.
* https://github.com/jenkinsci/rebuild-plugin
* https://plugins.jenkins.io/rebuild/

### Submitter checklist

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
@@ -141,12 +141,34 @@ public class ReplayAction implements Action {
         return null;
     }
 
+    /** Offer a fallback Rebuild link if Replay is not permitted? */
     /* accessible to Jelly */ public boolean isRebuildEnabled() {
         if (!run.hasPermission(Item.BUILD)) {
             return false;
         }
         if (!run.getParent().isBuildable()) {
             return false;
+        }
+        if (Jenkins.get().getPlugin("rebuild") != null) {
+            // Go via reflection to avoid a pom.xml (and subsequent
+            // Jenkins controller) dependency on the "competing" plugin:
+            try {
+                Class<?> abstractRebuildAction = Jenkins.get()
+                        .getPluginManager()
+                        .uberClassLoader
+                        .loadClass("com.sonyericsson.rebuild.AbstractRebuildAction");
+                if ((Boolean)
+                        abstractRebuildAction.getMethod("isRebuildAvailable").invoke(null)) {
+                    // THAT plugin will handle the Rebuild link better
+                    // (parameterized if need be), so we here should
+                    // show nothing at all
+                    return false;
+                }
+            } catch (ClassNotFoundException e) {
+                // ignore
+            } catch (Exception e) {
+                LOGGER.log(Level.WARNING, "failed to check for rebuild-plugin", e);
+            }
         }
 
         return true;
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction/index.jelly
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction/index.jelly
@@ -29,15 +29,17 @@
                         </f:form>
                     </j:when>
                     <j:otherwise>
-                        <h1>${%Rebuild(it.owner.displayName)}</h1>
-                        <p>
-                            <j:out value="${%blurb.rebuild}"/>
-                        </p>
-                        <f:form action="rebuild" method="POST" name="rebuild">
-                            <f:bottomButtonBar>
-                                <f:submit value="${%Run}"/>
-                            </f:bottomButtonBar>
-                        </f:form>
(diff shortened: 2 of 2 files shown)


jtnord:
> Non-admin users see two "Rebuild" buttons on Pipeline builds when the Rebuilder plugin is installed.

But are these even the same thing?  the pipeline "Rebuild" uses the exact same pipeline definition and libraries.
the rebuild plugin appears to just use the same parameters to kick off a new build that would use the version of the pipeline as currently defined (either from the branch in SCM that this points to or the version currently inline in the configuration).

jglick (changes requested):
https://github.com/jenkinsci/workflow-cps-plugin/blob/8df38f05c3c56b0772851985ddd333aada8c8be2/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java#L258 is also specialized behavior intended to allow `checkout scm` to produce the same result when replaying a build of a branch project even if the remote head has since moved.

Somehow distinguishing the actions visually would make sense but as written I would not accept this.

mawinter69:
Wouldn't it be ok to just always show `Replay` as the actions displayname? It will anyway open another page where you get an explanation what you can do and a corresponding button

jimklimov:
> > Non-admin users see two "Rebuild" buttons on Pipeline builds when the Rebuilder plugin is installed.
> 
> But are these even the same thing? the pipeline "Rebuild" uses the exact same pipeline definition and libraries. the rebuild plugin appears to just use the same parameters to kick off a new build that would use the version of the pipeline as currently defined (either from the branch in SCM that this points to or the version currently inline in the configuration).

From what I gather, Jelly files in both plugins append an URI to `$BUILD_URL`, and the fallback in workflow-cps-plugin adds `rebuild` so leads to same activity as the Rebuilder plugin for jobs without parameters. It is actually counter-productive I think for jobs with parameters, where users expect to tweak in the next page but get the build with same parameters (like in Replay) queued/running immediately.

> Wouldn't it be ok to just always show Replay as the actions displayname? It will anyway open another page where you get an explanation what you can do and a corresponding button

I think not: for one, without the privileges to see/edit the pipeline code, the fallback Rebuild is not really a Replay (as the logic people have associated with the name). Also there is no other page with `rebuild` (not `rebuild/parameterized`).

jimklimov:
> `ReplayAction` is also specialized behavior intended to allow checkout scm to produce the same result when replaying a build of a branch project even if the remote head has since moved.

Are you sure `ReplayAction` class plays any role when the fallback `$BUILD_URL/rebuild` is called?

This gets me wondering now, actually: can two plugins serve the same URI -- so on controllers without the Rebuilder plugin the `$BUILD_URL/rebuild` does something very different from controllers that have both plugins (so in review we are unconsciously talking about behaviors of unrelated logic hidden behind same name and somehow overridden by presence of the other plugin)?

And/or if the Rebuilder plugin taps into Jenkins-provided `.../rebuild` URL whichever way that is defined by core/workflow, and only adds the `.../parameterized` URI handling?

mawinter69:
ReplayAction has a fixed url `replay`. The text in the sidepanel depends whether you have the replay permission (or configure) or not. In both cases when you click there you get forwarded to the index.jelly of the ReplayAction. There depending on the permissions you can either edit the script or not., With edit permissinos it calls the `doRun` method in the action, without it calls the `doRebuild` method.
So there is no conflict with the rebuild plugin

mawinter69:
And the rebuild behaviour here is closer to a Replay than a Rebuild. The Rebuild plugin behaviour is when parameterized `Trigger the job with same parameters prefilled` basically, no guarantee that you get the same pipeline script or libraries.
Even without the replay permission, the behaviour here is closer to a replay I think.

jimklimov:
If that is so, then probably renaming of the link would suffice, e.g. `Replay` (for non-admins) and `Replay (editable)`. Not that it would be easy to explain the difference from Rebuild (from the other plugin) without an additional article on the nuances, but at least they would differ visually and cause fewer mistakes and questions.

These lines in Jelly led me to believe it calls the `rebuild` URI: https://github.com/jenkinsci/workflow-cps-plugin/blob/8df38f05c3c56b0772851985ddd333aada8c8be2/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction/index.jelly#L31-L41

PR #1769, opened 2026-04-28
Default `initializeStorageFromOnLoad` to false

jglick:
Switching the default for the option introduced in #1008. CloudBees CI has been running in this mode in HA controllers for most of a year now and I am unaware of any resulting problems. In the meantime, the deadlock that can result from `createPlaceholderNodes` from `onLoad` has been observed in non-HA controllers (https://github.com/jenkinsci/workflow-job-plugin/pull/747) so I think it is best to just deprecate this trick for everyone.

Also bumping the Jenkins baseline from 2.504.x to 2.516.x, to match that of `workflow-job`.

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -60,6 +60,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <!-- TODO until in BOM -->
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-job</artifactId>
+        <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/758 -->
+        <version>1571.1579.v703855f75573</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
   <dependencies>
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -60,12 +60,10 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-      <!-- TODO until in BOM -->
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-job</artifactId>
-        <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/758 -->
-        <version>1571.1579.v703855f75573</version>
+        <version>1571.1580.v18e46842c125</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
@@ -863,7 +863,7 @@ public class CpsFlowExecution extends FlowExecution implements BlockableResume {
                 initializeStorage(); // Throws exception and bombs out if we can't load FlowNodes
             } catch (Exception ex) {
                 var propName = CpsFlowExecution.class.getName() + ".initializeStorageFromOnLoad";
-                if (SystemProperties.getBoolean(propName, true)) {
+                if (SystemProperties.getBoolean(propName)) {
                     LOGGER.log(
                             Level.WARNING,
                             ex,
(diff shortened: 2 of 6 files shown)


jglick:
Need to adapt some tests, restoring changes made in #1008 that I had reverted when making no behavioral change by default.

plugin/pom.xml
```diff
@@ -60,6 +60,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <!-- TODO until in BOM -->
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-job</artifactId>
+        <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/758 -->
```
jglick: https://github.com/jenkinsci/workflow-job-plugin/pull/758

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/PersistenceProblemsTest.java
```diff
@@ -168,46 +171,37 @@ private static InputStepExecution getInputStepExecution(WorkflowRun run, String
...
-            executionAndBuildResult[1] = run.getResult();
         });
         story.then(j -> {
             WorkflowJob r = j.jenkins.getItemByFullName(DEFAULT_JOBNAME, WorkflowJob.class);
             WorkflowRun run = r.getBuildByNumber(build[0]);
-            assertCompletedCleanly(run);
-            //            assertNulledExecution(run);
-            Assert.assertEquals(Result.SUCCESS, run.getResult());
-            assertResultMatchExecutionAndRun(run, executionAndBuildResult);
+            assertCompletedCleanly(run, true);
```
RanjitCS: running this locally and i think the `true` here is dead code. After restart,`getExecution()` returns `null` (the `initializeStorage` failure hits the catch block in `WorkflowRun.getExecution` which returns null), so the instanceof `CpsFlowExecution` check is false and the `verifyFlowEndNode` assertions are never reached. Same for `completedNoNodesPersisted` below

RanjitCS (commented):
minor change on test, else LGTM

RanjitCS (approved):

plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/FlowDurabilityTest.java
```diff
@@ -358,23 +361,23 @@ static void verifyFailedCleanly(Jenkins j, WorkflowRun run) throws Exception {
...
         // Assert that we have the appropriate flow graph entries
         FlowExecution exec = run.getExecution();
-        List<FlowNode> heads = exec.getCurrentHeads();
-        Assert.assertEquals(1, heads.size());
+        if (verifyFlowEndNode) {
+            List<FlowNode> heads = exec.getCurrentHeads();
+            Assert.assertEquals(1, heads.size());
+        }
         verifyNoTasksRunning(j);
         Assert.assertEquals(0, exec.getCurrentExecutions(false).get().size());
```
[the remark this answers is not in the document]
jglick: If and when such a situation occurs, the test will fail and be adapted accordingly.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
gbhat618, line 875:
```suggestion
                    LOGGER.log(
                        Level.WARNING,
                        ex,
                        () -> "Error initializing storage and loading nodes for " + this
                            + "; marking build as failed (use -D" + propName + "=true to create placeholder nodes instead)");
                    throw ex;
```

because the default is flipped now

gbhat618 (approved):
Went through the discussions in https://github.com/jenkinsci/workflow-cps-plugin/pull/1008 but as it is proven to be working in production - I think the change is good.

plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: -0 I guess. We do not want to advertise the system property to users; this is there for emergency use in case this change turns out to cause some unforeseen issue with some plugin or some unusual situation. And logging the stack trace here should be redundant (it is up to the ultimate catcher to log it).

plugin/pom.xml
```diff
@@ -60,6 +60,11 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
```
jglick, line 63: Note: not bothering to add TODO comments any more, since https://github.com/jenkinsci/plugin-pom/pull/1381 should be used to cover that going forward.

Issue #1772, opened 2026-05-04
Move Declarative-specific samples to `pipeline-model-definition-plugin`


jglick:
Noted in https://github.com/jenkinsci/workflow-cps-plugin/pull/350. At least as of https://github.com/jenkinsci/workflow-cps-plugin/pull/1765 the result is a bit clearer.

PR #1775, opened 2026-05-11
Playwright test of `workflow-editor.js`

jglick:
https://github.com/jenkinsci/workflow-cps-plugin/pull/1738#pullrequestreview-4221129000

[CloudBees-internal link](https://cloudbees.atlassian.net/browse/BEE-71824)

----

My first concern is how to run Playwright itself in CI. I do not really want it [downloading Chrome from Microsoft’s CDN](https://playwright.dev/java/docs/browsers#install-behind-a-firewall-or-a-proxy) on every build. I thought about [running in Docker instead](https://playwright.dev/java/docs/ci#jenkins), though this introduces its own problems: using Testcontainers does not seem practical since `RealJenkinsRule` does not support running in a container yet, and `JenkinsRule` does not let you listen on anything other than 127.0.0.1, which would not allow the Docker host to be exposed to the Playwright container; I could run this test directly using Maven inside that image, but `pipeline-library` does not support that, so I would need to run this test in its own CI branch and collect test results manually.

PR #1776, opened 2026-05-11
Stop proposing to use Maven tool installer in samples

jglick:
We do not generally want to encourage use of tool installers these days, especially something downloading from a public (non-Jenkins) server. Made samples not run out of the box. Predates #350. Not addressing #1772 at this time.

### DIFF
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
@@ -1,11 +1,6 @@
 pipeline {
     agent any
 
-    tools {
-        // Install the Maven version configured as "M3" and add it to the path.
-        maven "M3"
-    }
-
     stages {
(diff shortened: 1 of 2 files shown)


plugin/src/main/resources/org/jenkinsci/plugins/workflow/cps/samples/github-maven.groovy
```diff
@@ -1,22 +1,17 @@
...
-
     stages {
         stage('Build') {
             steps {
                 // Get some code from a GitHub repository
                 git 'https://github.com/jglick/simple-maven-project-with-tests.git'
 
                 // Run Maven on a Unix agent.
-                sh "mvn -Dmaven.test.failure.ignore=true clean package"
+                sh 'mvn -Dmaven.test.failure.ignore=true clean package'
```
jglick, line 11: Should not have been using `GString`!

timja (approved):

jglick:
(context: #1775)

PR #1780, opened 2026-05-14
Upgrade to jQuery 4.0 and jQuery UI 1.14.2

jglick:
Subsumes https://github.com/jenkinsci/workflow-cps-plugin/pull/1738 and amends https://github.com/jenkinsci/workflow-cps-plugin/pull/1775.

### JavaScript dependency upgrades
- Upgrades jQuery from 3.7.1 to 4.0.0
- Upgrades jQuery UI from 1.14.1 to 1.14.2
- Upgrades ace-builds from 1.43.5 to 1.44.0

### Yarn lockfile deduplication
Works around https://github.com/dependabot/dependabot-core/issues/5830 where Dependabot does not deduplicate dependencies. The original Dependabot PR had jQuery 3.7.1 installed to satisfy jquery-ui's transitive dependency `>=1.12.0 <5.0.0`, while the direct dependency requested `^4.0.0`. Running `yarn install` deduplicates and resolves both ranges to jQuery 4.0.0.

### Test improvement
Enhances `WorkflowEditorTest` to capture and assert on browser console errors/warnings using Playwright. Uses JUnit 5's `assertAll()` to report both functional test failures and console issues together, making it easier to diagnose root causes when UI elements fail to load due to JavaScript errors.

CloudBees-internal link: https://cloudbees.atlassian.net/browse/BEE-71825

### DIFF
--- a/plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowEditorTest.java
+++ b/plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowEditorTest.java
@@ -25,15 +25,19 @@
 package org.jenkinsci.plugins.workflow;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.startsWith;
+import static org.junit.jupiter.api.Assertions.assertAll;
 
+import com.microsoft.playwright.ConsoleMessage;
 import com.microsoft.playwright.Page;
 import com.microsoft.playwright.junit.Options;
 import com.microsoft.playwright.junit.OptionsFactory;
 import com.microsoft.playwright.junit.UsePlaywright;
 import com.microsoft.playwright.options.AriaRole;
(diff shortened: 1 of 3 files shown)


plugin/src/test/java/org/jenkinsci/plugins/workflow/WorkflowEditorTest.java
```diff
@@ -55,41 +59,57 @@ void smokes(JenkinsRule r, Page page) throws Exception {
...
-        // Cut the sample to clipboard
-        textbox.press("ControlOrMeta+a");
-        textbox.press("ControlOrMeta+x");
-        assertThat(
-                "looks like scripted.groovy",
-                (String) page.evaluate("navigator.clipboard.readText()"),
-                startsWith("node {"));
+        assertAll(
+                () -> {
+                    page.locator("#workflow-editor-samples")
```
jglick, line 72: (ignore WS)

plugin/yarn.lock
jglick, line 1: (CC @timja; this is all Greek to me)

jtnord (approved):

PR #1011, opened 2025-04-10
Dependabot groups take 2

jglick:
Trying to fix #1010.

PR #1782, opened 2026-05-14
Label `javascript-security` updates as `bug`

jglick:
Amends #1011 to apply the `bug` label to security updates in the `javascript-security` group (e.g. #1773), rather than the default `dependencies` label applied to all other Dependabot PRs.

[CloudBees-internal reference](https://cloudbees.atlassian.net/browse/BEE-71826)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/label-dependabot-security.yml
+++ b/.github/workflows/label-dependabot-security.yml
@@ -1,0 +1,21 @@
+name: Label Dependabot security updates
+
+on:
+  pull_request_target:
+    types: [opened]
+    branches:
+      - 'dependabot/npm_and_yarn/plugin/javascript-security-*'
+
+permissions:
+  pull-requests: write
+
+jobs:
+  label-security-updates:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]' && contains(github.head_ref, 'javascript-security')
+    steps:
+      - name: Replace dependencies label with bug
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr edit ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --remove-label dependencies --add-label bug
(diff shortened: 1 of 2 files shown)


.github/workflows/label-dependabot-security.yml
```diff
@@ -0,0 +1,21 @@
+name: Label Dependabot security updates
```
timja, line 1: can't you just use configuration for this?
https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/customizing-dependabot-security-prs#example-1-configuration-for-security-updates-only

.github/workflows/label-dependabot-security.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on:
+  pull_request_target:
+    types: [opened]
+    branches:
+      - 'dependabot/npm_and_yarn/plugin/javascript-security-*'
```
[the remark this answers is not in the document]
jglick: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request_target-workflow-based-on-the-head-or-base-branch-of-a-pull-request indeed. Maybe I could use https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs instead?

.github/workflows/label-dependabot-security.yml
jglick: That was what I tried first, but the syntax was invalid: `labels` is only permitted at the level of an `updates` element, not a group.

timja (approved):

jglick:
Will have to wait to see if it works. Given past experience, we should not be waiting long.

PR #1790, opened 2026-06-05
Add Replay option to experimental Run UI + support for other plugins

janfaracik:
Up for discussion.

The intention behind this PR is to add the _Replay_ (and _Edit and replay_) action to the experimental Run UI app bar. This is very much inspired by the Pipeline Graph View plugin, which features a similar UI. This PR also introduces support for plugins to contribute actions to this menu.

The current proposed hierarchy is:

* Replay
  * Edit and replay
  * Rebuild _(via Rebuilder plugin)_
  * Restart from stage _(via Pipeline: Declarative plugin)_

### Screenshot

image

### Questions around this

* Is this the right spot for this extension? Should this live in core instead?
* Is Replay the most commonly used action, and thus deserves top spot over other potential actions?

@timja 

### Submitter checklist


timja:
Replay makes the most sense, although terminology wise I think it should be Rebuild or something, replay is just so weird - checking pipeline graph view we've used Rerun there.

------

I wonder if restart from stage should have a different treatment?

From what I remember in the classic view there's an action that takes you to a page to restart it.
From BlueOcean there's a link above the logs for the selected stage? which makes more sense doesn't it?

janfaracik:
> From BlueOcean there's a link above the logs for the selected stage? which makes more sense doesn't it?

~Could you ping a screenshot of this?~

Ah I see - I could see having both? E.g. the menu opens a dialog with a select input _and_ you could do it from the Pipeline Graph View page? I like the idea of grouping all related _new build_ controls.

image

PR #1802, opened 2026-07-03
Adding in OAUTH login

NickTheDevOpsGuy:
Adds Git OAuth/PAT support for Pipeline SCM credential lookups by adapting Jenkins secret-text credentials into Git-compatible username/password credentials for HTTP/HTTPS remotes.

The Git credential flow asks Jenkins for `StandardUsernameCredentials`, while OAuth/PAT values are commonly stored as `StringCredentials` from the plain credentials plugin. This change adds a credentials provider that exposes those secret-text tokens as username/password credentials using the token as the password and `oauth2` as the username, scoped only to HTTP/HTTPS Git credential lookups.

### Testing done

Ran the focused automated test suite:

```bash
mvn -pl plugin -am -Dtest=GitOAuthTokenCredentialsProviderTest test
Result: BUILD SUCCESS
Covered scenarios:
Secret-text OAuth token is exposed as Git-compatible HTTPS username/password credentials.
Token can be selected by credential id.
Token is not exposed for SSH Git credential lookups.
Adapter is not exposed during broad credential lookups.
Submitter checklist

Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch!

Ensure that the pull request title represents the desired changelog entry

Please describe what you did

Link to relevant issues in GitHub or Jira

Link to relevant pull requests, esp. upstream and downstream changes

Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GitOAuthTokenCredentialsProvider.java
+++ b/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GitOAuthTokenCredentialsProvider.java
@@ -1,0 +1,143 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2026.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/GitOAuthTokenCredentialsProvider.java [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Exposes secret-text OAuth tokens as Git-compatible username/password credentials.
+ *
+ * <p>The git plugin asks the credentials subsystem for {@link StandardUsernameCredentials} when authenticating
+ * HTTP(S) remotes. OAuth/PAT credentials are commonly stored in Jenkins as {@link StringCredentials}, so without this
+ * small adapter those token credentials are invisible to Pipeline jobs using {@link CpsScmFlowDefinition}.
+ */
+@Extension
+public class GitOAuthTokenCredentialsProvider extends CredentialsProvider {
```
timja, line 51: I don't see what this has to do with the workflow cps plugin.

Just change the git plugin to be more flexible? or move this somewhere more appropriate
cc @MarkEWaite

NickTheDevOpsGuy: Sorry I a guess I am confused where is the git plugin located that is where this change needs to go the help section on that plugin links here.

NickTheDevOpsGuy: CleanShot 2026-07-03 at 07 14 22@2x

If I need to add OAUTH to this git plugin where do I make that change then?

jtnord: bitbucket already supports OAUth  https://github.com/jenkinsci/bitbucket-branch-source-plugin/blob/c0f3b6ce58c723b565142c2fcd1d2ab0c0872b63/src/main/java/com/cloudbees/jenkins/plugins/bitbucket/impl/credentials/BitbucketOAuthAuthenticatorSource.java#L33 ?

(so use the bitbucket source not the git source when your code is in bitbucket)

jtnord:
code does not belong here.  Please take any follow up questions to a [community chat/forum](https://www.jenkins.io/participate/connect/#join-our-chats-and-forums) (I believe the functionality you require already exists)

PR #1811, opened 2026-07-24
[JENKINS-62410] Do not render the rich script editor for read-only viewers

Racknaraock:
Fixes https://github.com/jenkinsci/workflow-cps-plugin/issues/1810

# [JENKINS-62410] Do not render the rich script editor for read-only viewers

## Problem

`workflow-editor.jelly` rendered the rich ACE-based script editor for any user who could reach a job's `configure` page, regardless of whether they actually held `Item/Configure`. A user granted only `Item/Read` + `Item/ExtendedRead` (i.e. allowed to view, but not edit, the pipeline definition) still got a fully interactive editor widget, even though any edit they made could never be saved.

Reported in [JENKINS-62410](https://issues.jenkins.io/browse/JENKINS-62410), and confirmed by a maintainer in [jenkinsci/workflow-aggregator-plugin#1173](https://github.com/jenkinsci/workflow-aggregator-plugin/issues/1173):

> when showing the page to a user that only has read privileges, the script should be visible but not presented in an editor

Jenkins core already has a convention for this case: pages that set the ambient `readOnlyMode` Jelly variable (e.g. `Job/configure.jelly` sets it based on `Item.CONFIGURE`) expect form controls to render read-only via `lib/form/possibleReadOnlyField.jelly`. `workflow-editor.jelly` never checked this variable, so it always rendered the editor whenever `enableWorkflowEditor` was on.

## Changes

- `workflow-editor.jelly`: extend the existing `j:when` condition that decides whether to render the ACE editor to also require `!readOnlyMode`. When `readOnlyMode` is true, the `j:otherwise` branch is used instead, which already renders a plain `f:textarea` — respected by core's read-only rendering machinery, so the script shows as plain text instead of the ACE editor. One line changed, no other files touched.

## Testing

- **New unit test**: `CpsFlowDefinitionTest.readOnlyViewerSeesPlainScriptNotEditor`, which uses `MockAuthorizationStrategy` + `JenkinsRule.WebClient` to verify that:
  - a viewer with `Item.READ`/`Item.EXTENDED_READ` but not `Item.CONFIGURE` never gets the `.workflow-editor-wrapper`, and instead sees the script as plain read-only text (`pre.jenkins-readonly`) containing the original script content.
  - a user with `Item.CONFIGURE` still gets the rich editor as before (no regression for the normal case).
- Targeted test run (`mvn test -Dtest=CpsFlowDefinitionTest`): 11/11 passing, no regressions in the rest of the class.
- **Manual verification** on a local `mvn hpi:run` Jenkins instance: compared an admin user (Item/Configure granted) against a read-only viewer (Item/Read + Item/ExtendedRead only, `-Dhudson.security.ExtendedReadPermission=true`) opening the same job's configure page. Confirmed the admin still sees the ACE editor, while the viewer sees the script as inert read-only text with no editor widget.

## Checklist

- [x] Referenced the Jira issue (JENKINS-62410) in the commit messages
- [x] Added a test that verifies the change
- [x] Unit tests pass locally
- [x] Interactively tested the change on a local Jenkins instance

### DIFF
--- a/plugin/src/main/resources/org/jenkinsci/plugins/workflow/editor/workflow-editor.jelly
+++ b/plugin/src/main/resources/org/jenkinsci/plugins/workflow/editor/workflow-editor.jelly
@@ -7,7 +7,7 @@
         </st:attribute>
     </st:documentation>
     <j:choose>
-        <j:when test="${app.getDescriptor('org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition').enableWorkflowEditor}">
+        <j:when test="${app.getDescriptor('org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition').enableWorkflowEditor and !readOnlyMode}">
             <div class="workflow-editor-wrapper" style="display: none; position: relative">
                 <f:textarea value="${script}" checkMethod="post" checkUrl="${attrs.checkUrl}" checkDependsOn="${attrs.checkDependsOn}"/>
                 <div class="editor" style="height: 350px;" samplesUrl="${rootURL}/workflow-cps-samples/" />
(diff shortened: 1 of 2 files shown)


timja:
Please provide screenshots before and after

Racknaraock:
> Please provide screenshots before and after

@timja this works? 

Before my change :

JENKINS-62410-antes-full

After my change: 

JENKINS-62410-despues-full

timja (approved):
thanks!

PR #1834, opened 2026-09-16
Address tests failure after 2026-09-16 security advisory

Kevin-CB:
https://www.jenkins.io/security/advisory/2026-09-16/ address some vulnerabilities on script-security that make a test message different. (Fix in question: https://github.com/jenkinsci/script-security-plugin/commit/06869826dd9b3cc608d6bae6ca8db72241c45154 & https://github.com/jenkinsci/groovy-sandbox/commit/7fd9e9530eada33ae5cfcc080f10d0f441fa2b60)

This is needed to avoid BOM PCT failures, once the new version is released, this need to be bumped as part of https://github.com/jenkinsci/bom/pull/7435.

### Testing done
I confirmed locally that the two tests now passes.

### Submitter checklist
- N/A Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- N/A Ensure you have provided tests that demonstrate the feature works or the issue is fixed


plugin/pom.xml
```diff
@@ -102,6 +102,8 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>script-security</artifactId>
+      <!-- TODO: remove once the BOM picks up the SECURITY-3923 fix. -->
+      <version>1422.v06869826dd9b_</version>
```
jglick: Please put this in `dependencyManagement` where https://github.com/jenkinsci/workflow-cps-plugin/blob/bb753c50f0772c2b2e1f6f755bc25b29eda80c0e/plugin/pom.xml#L53 expects it.

jglick (changes requested):

plugin/pom.xml
jglick: (no need for a comment, just a version override suffices)

jglick:
(`plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java` changes look fine to the extent I understand them)

plugin/pom.xml
Kevin-CB: https://github.com/jenkinsci/workflow-cps-plugin/pull/1834/commits/34f231946af5ab4dde1d7cf28f462aeea982a897

jglick (approved):

